PR #2, opened 2022-10-11
Add annotations

thephw:
### Changelog
- Test annotation changes in k8s job
- Update README.md


sj26 (approved):
Neat! I left a comment on the plugin — I think `annotations` should be a top-level property. But otherwise good to go 👍

PR #7, opened 2022-11-14
v2: Update README

benmoss:
cc @jmcshane

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,10 +8,15 @@ A Kubernetes controller that runs Buildkite jobs as workloads on Kubernetes.
 
 The controller uses the [Buildkite GraphQL API](https://buildkite.com/docs/apis/graphql-api) to watch for scheduled work that uses the `kubernetes` plugin.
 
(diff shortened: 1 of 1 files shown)


jmcshane (commented):
Aren’t we using PodTemplate here instead of PodSpec? We will need to be able to set metadata such as labels and annotations

jmcshane:
> Aren’t we using PodTemplate here instead of PodSpec? We will need to be able to set metadata such as labels and annotations

Sorry, this was meant to be inline with the Readme code, but GitHub made it a full review comment

README.md
```diff
@@ -1,94 +1,125 @@
...
+kind: Pod
+metadata:
+  name: agent-abc
+spec:
+  initContainers:
+  - image: buildkite/agent:latest
+    args:
+    - start
+    - --acquire-job=abc
+    volumeMounts:
```
jmcshane, line 82: I know this is a future thing, but we might want to say something early on in the readme about “workspaces” and checking out the code into these volumes

benmoss: Added more description of the workspace in the "Explain it like I'm 5" section

PR #10, opened 2022-11-17
Support for user-specified containers

benmoss:
This PR addresses splitting up the agent bootstrap so that it can happen in a separate container from the container that does the actual job execution, as specified in the kubernetes plugin section.

As of this PR we're not cloning the source repo into the workspace, so the test example still just runs "echo hello world" from inside an alpine container, with no references to the repo source.

Tracking changes to the agent codebase in a submodule, the code is currently in a fork in my personal account: https://github.com/benmoss/agent/tree/kubernetes-job-runner

Longer term we can get these changes merged upstream and released into the normal images

### DIFF
--- a/scheduler/scheduler.go
+++ b/scheduler/scheduler.go
@@ -190,7 +190,6 @@ func podFromJob(
 	for k, v := range envMap {
 		switch k {
 		case "BUILDKITE_PLUGINS": //noop
-			env = append(env, corev1.EnvVar{Name: k, Value: "[]"})
 		case "BUILDKITE_COMMAND": //noop
 		default:
 			env = append(env, corev1.EnvVar{Name: k, Value: v})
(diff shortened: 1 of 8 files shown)


scheduler/scheduler.go
```diff
@@ -151,25 +188,57 @@ func podFromJob(
 		Value: job.Uuid,
 	})
 	for k, v := range envMap {
-		env = append(env, corev1.EnvVar{Name: k, Value: v})
+		switch k {
+		case "BUILDKITE_PLUGINS": //noop
```
jmcshane, line 192: You have a // noop comment here, but you are appending the EnvVar in this case?

benmoss: Good call, I think I was wrestling with how this would work and thought at one point that overriding plugins like this was necessary. It's not, we can drop the env var completely.

PR #21, opened 2022-12-08
Generate an agent token at startup

benmoss:
- No longer need both an API token and an agent token
- Slightly more secure than reusing the same agent token every time


benmoss:
We're going to keep requiring this, @sj26 points out that token generation won't work with users who have the Clusters feature enabled

PR #22, opened 2022-12-08
Support uncapped job count

jmcshane:


benmoss:
I think I'm getting an error [from here](https://github.com/buildkite/agent-stack-k8s/blob/f8260034015ddf74202598d08668d323c80520a6/monitor/monitor.go#L41) when I try to use `0` as the max-in-flight. I think we can drop the cache entirely in the case of unlimited, or maybe just set it to a large size. WDYT?

benmoss (approved):

PR #23, opened 2022-12-08
Watch builds on all pipelines, use agent tags

benmoss:
- Instead of requiring the user to specify specific pipelines to watch, watch all jobs in an org
- Support agent tag filtering as a way of restricting which jobs we schedule for

### DIFF
--- a/api/genqlient.graphql
+++ b/api/genqlient.graphql
@@ -85,8 +85,8 @@ query GetBuild($uuid: ID!) {
   }
 }
 
-query GetScheduledBuilds($slug: ID!) {
-  pipeline(slug: $slug) {
+query GetScheduledBuilds($slug: ID!, $agentQueryRules: [String!]) {
+  organization(slug: $slug) {
     # @genqlient(pointer: true)
     id
     jobs(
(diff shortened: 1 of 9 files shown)


api/genqlient.graphql
```diff
@@ -85,15 +85,16 @@ query GetBuild($uuid: ID!) {
...
-  pipeline(slug: $slug) {
+query GetScheduledBuilds($slug: ID!, $agentQueryRules: [String!]) {
+  organization(slug: $slug) {
     # @genqlient(pointer: true)
     id
     jobs(
       state: [SCHEDULED]
       type: [COMMAND]
       first: 100
       order: RECENTLY_ASSIGNED
```
sj26, line 96: Jobs won't have an "agent assigned at" timestamp which is the "recently assigned" ordering while they are "scheduled", so I reckon this might be better as the default of "RECENTLY_CREATED". Ideally this should use the same ordering as agent dispatch, which is based on the job's "scheduled at", but I don't think we have a graphql cursor order option for that currently.

sj26: Oh.. but RECENTLY_CREATED will be the opposite of what we want, I think. Frustrating, I don't think there is a better option right now. So maybe ignore the above for the moment.

sj26 (approved):
Nice

api/genqlient.graphql
benmoss: Yeah, I tested both orderings and `RECENTLY_ASSIGNED` seemed to be the one that gave me them in order created, confusingly enough

PR #34, opened 2022-12-15
Monitor watches for completions

benmoss:
- Simplifies the relationship between the monitor and the scheduler
- Fixes a bug with the previous "scanKnownJobs" implementation where it found jobs that didn't match the configured tags
- Adds the tag that we acquired the job through as a label so that we can correlate it after a restart

### DIFF
--- a/monitor/monitor.go
+++ b/monitor/monitor.go
@@ -84,7 +84,7 @@ func (m *Monitor) Scheduled() <-chan Job {
 }
 
 func (m *Monitor) start() {
-	m.logger.Debug("started", zap.Strings("tags", m.cfg.Tags))
+	m.logger.Info("started", zap.Strings("tags", m.cfg.Tags), zap.String("org", m.cfg.Org), zap.String("namespace", m.cfg.Namespace), zap.Int("max-in-flight", m.cfg.MaxInFlight))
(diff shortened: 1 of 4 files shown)


monitor/monitor.go
```diff
@@ -124,23 +140,100 @@ func (m *Monitor) start() {
...
+		ListFunc: func(options metav1.ListOptions) (runtime.Object, error) {
+			options.LabelSelector = selector
+			return m.k8s.BatchV1().Jobs(m.cfg.Namespace).List(ctx, options)
+		},
+		WatchFunc: func(options metav1.ListOptions) (watch.Interface, error) {
+			options.LabelSelector = selector
+			return m.k8s.BatchV1().Jobs(m.cfg.Namespace).Watch(ctx, options)
+		},
+	}
+	removeIfCompleted := func(obj interface{}) {
```
jmcshane, line 163: We talked last week about how the buildkite agent has a tendency towards outputting a ton of information for user diagnosis. I wonder if this is a good place to start adding some non-debug logs, especially around job lifecycle.

benmoss: Yeah, I have been pretty much logging everything at debug, but it probably is a good idea to have some non-debug stuff

PR #45, opened 2023-01-05
Plugins support

titilambert:


benmoss:
We're planning to open up the issues on this repo, so we can move that there once @dabarrell does so 😄

dabarrell:
@benmoss I was just given admin perms, so have now enabled issues :)

PR #48, opened 2023-01-05
Add golangci-lint, fix warnings

benmoss:
- Simplify the monitor test, it had an unused variable warning from `functionReturned`
- The job lister didn't need to add a namespace indexer to the job informer, it comes with one by default. When I added error handling to the `AddIndexers` call, we were missing an "indexer conflict" error.


benmoss:
Looks like it's extremely slow in CI, maybe due to lack of caching. Gonna investigate further.

benmoss:
Gonna ignore the caching/slowness problem, it's about as slow as our integration tests and they run in parallel so it doesn't add any time to the total CI run

PR #52, opened 2023-01-09
Lint pipelines with json schema

benmoss:
Adds a subcommand to the agent-stack-k8s binary to lint a pipeline. Thinking this will be a useful tool for users to run before a dynamic pipeline upload step in order to find errors in their configuration.


benmoss:
Have a PR open against the go-buildkite sdk to support plugins being an array instead of a map, for now using a go mod `replace` directive to use my fork: https://github.com/buildkite/go-buildkite/pull/106

PR #72, opened 2023-01-17
Fix agent not failing when plugin errors

### LINKED ISSUE #70: Step green if a plugins crashed
I just got an example where my step was green even if the plugin crashed because of a missing permission

The step should be red if there is this kind of issue

benmoss:
Fixes #70 

The fix in the agent was embarrassingly simple, since they already had named returns [on the method](https://github.com/buildkite/agent/blob/kubernetes-job-runner/bootstrap/bootstrap.go#L68), we just have to send the same `exitCode` to our rpc server.

### DIFF
--- a/.buildkite/steps/agent.sh
+++ b/.buildkite/steps/agent.sh
@@ -4,7 +4,7 @@ set -euxo pipefail
 apk add just go jq
 
 docker buildx create --driver kubernetes --use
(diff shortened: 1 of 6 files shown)


benmoss:
Just merged in https://github.com/buildkite/agent-stack-k8s/pull/69 so that this PR can pass (it requires the latest changes to the agent).

PR #88, opened 2023-01-23
Max in flight check should not pass default value

jmcshane:


monitor/monitor.go
```diff
@@ -152,7 +152,7 @@ func (m *Monitor) reachedMaxInFlight() bool {
 		return true
 	}
 	for _, job := range jobList {
-		if job.Status.Active != 0 {
+		if job.Status.CompletionTime != nil {
```
jmcshane, line 155:
```suggestion
		if job.Status.CompletionTime == nil {
```

benmoss:
Replacing with https://github.com/buildkite/agent-stack-k8s/pull/90

PR #98, opened 2023-01-30
Enhance linter to act as a preprocessor

benmoss:
Allow users to move step templates to a separate file. Maybe we want to rename it from being a linter to being a .... parser? templater?


benmoss:
Gonna close this, I'm not sure it's really adding that much value or is a direction we want to go in right now

PR #104, opened 2023-01-31
Add experimental sidecar support

benmoss:
Adds the ability to specify `sidecars` as a new top-level field, aside `podSpec` and `gitEnvFrom`. 

It's a list of `v1.Container` objects, like in the pod spec, but which will be started immediately and will run in parallel with the regular job containers. The controller will now terminate the whole pod when the agent container exits, because otherwise these containers would cause the pod/job to never complete.

An unfortunate consequence is that now the Kubernetes Jobs will always be considered failures, since the pod will be terminated instead of exiting on its own. This doesn't really matter from the perspective of the Buildkite job.

We might want to revisit whether we need Kubernetes Jobs at all anymore, or if we just want to run regular pods.

### DIFF
--- a/scheduler/scheduler.go
+++ b/scheduler/scheduler.go
@@ -34,6 +34,7 @@ func New(logger *zap.Logger, client kubernetes.Interface, cfg api.Config) *worke
 type KubernetesPlugin struct {
 	PodSpec    *corev1.PodSpec
 	GitEnvFrom []corev1.EnvFromSource
+	Sidecars   []corev1.Container `json:"sidecars,omitempty"`
 }
 
 func (w *worker) k8sify(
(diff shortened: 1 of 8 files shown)


scheduler/scheduler.go
```diff
@@ -171,6 +194,15 @@ func (w *worker) k8sify(
 	}
 
 	containerCount := len(podSpec.Containers) + systemContainers
+
+	for i, c := range k8sPlugin.Sidecars {
+		if c.Name == "" {
```
jmcshane, line 199: have we tested what happens here if there are conflicts in the names? I guess that would be a failure to schedule, but can we work around that somehow?

benmoss: Created a separate issue to think more about this kind of thing: https://github.com/buildkite/agent-stack-k8s/issues/111

PR #105, opened 2023-02-01
Add overview diagram

spkane:
This adds an overview diagram to the main README and also cleans up the README just a bit.


spkane:
@benmoss As, I was going over the workflow and trying to understand it, I pulled together a diagram of the k8s components and how things are generally wired together, and thought that it might be generally useful. Let me know if this needs any tweaks.

benmoss:
The color scheme of the svg is hard to read, maybe a different background color instead of blue

benmoss:
I'm not sure the what the svg is trying to communicate, it's kind of confusing in general

jmcshane:
@spkane I had an opportunity to review this diagram and I have a few notes here

* This "one big diagram" is a little challenging to parse out and glean important information quickly. I think there are two key things that we want to get answered in a diagram:
    1. how does a buildkite job land on an agent -> a pod get scheduled that uses --acquire-job that picks up the job ID and then is cleaned up at completion
    2. How is a job managed within the pod execution -> `buildkite-agent` binary copied by the init container, bootstrap called by the agent container, then work executed by the additional ones

* A more advanced diagram showing the `buildkite-agent` communication in the pod across the shared mounted workspace might be a nice thing to add when talking about how the job execution is handled. This could include things like the responsibility for uploading the logs, reporting status, and uploading artifacts.

Bottom line, let's make a simple map from buildkite job scheduled in the queue to pod execution at the top level, then a more detailed pod execution diagram that shows how the communication happens further down in the document.

jmcshane:
For example, in a diagram, I don't need to be concerned about all the "kubernetes"-ifying of the stack, yes it has a cluster role, service account, deployment, configmap, but to an end user this is a single component. They can look at the helm chart for the stack if they need to make customizations to the agent itself.

PR #135, opened 2023-02-26
Use `/bin/sh` as the default shell for buildkite agent start

triarius:
This will allow jobs to run containers based on the alpine images. These are small and should start up quicker, so are an attractive image to use for our users.


triarius:
Superseeded by https://github.com/buildkite/agent-stack-k8s/pull/140

PR #136, opened 2023-03-07
Fix integration tests were failing

triarius:
The reason was that merging the v2 branch into main automatically deleted the v2 branch, and the tests explicitly used this branch.

I've changed the `branch` used in the tests to be an empty string valued variable. The script that runs the tests will override it with BUILDKITE_BRANCH using ldflags.

I've also cleaned up some of the code involved in testing.


sj26 (approved):
Lovely 💚

PR #137, opened 2023-03-07
Add tags to agents launched by the controller

triarius:
This adds the agent-stack-k8s version, as well as the node name, namespace, and service account name of the pod.

The version is obtained through a variable set at compile time. It is passed directly as a tag to the `buildkite-agent start` command. The others are set as environment variables using the [downward api](https://kubernetes.io/docs/tasks/inject-data-application/environment-variable-expose-pod-information/). The agent then has the responsibility to read these and added the appropriate tags. Support for this is added in the companion PR in the agent repo: https://github.com/buildkite/agent/pull/2003

## Screenshot
2023-03-09T10:47:40,722595643+11:00

## Also

- Refactored the test code to break out the `testcase` objet
- Added an integration test that the agent token is interpolated
- Refactored the pipeline
- Changed the pipeline to push agent/controller/helm chart on branches, so we can test them out
	- we will re-visit this decision later, but I need it for now to do things like make the screenshot above
- made scheduler tests run in parallel

### DIFF
--- a/internal/integration/integration_test.go
+++ b/internal/integration/integration_test.go
@@ -272,3 +272,21 @@ func TestCleanupOrphanedPipelines(t *testing.T) {
 		})
 	}
 }
+
+func TestEnvVariables(t *testing.T) {
+	tc := testcase{
+		T:       t,
+		Fixture: "env.yaml",
+		Repo:    repoHTTP,
+		GraphQL: api.NewClient(cfg.BuildkiteToken),
+	}.Init()
+	ctx := context.Background()
+	pipelineID := tc.CreatePipeline(ctx)
+	tc.StartController(ctx, cfg)
+	build := tc.TriggerBuild(ctx, pipelineID)
+	tc.AssertSuccess(ctx, build)
+	tc.AssertLogsContain(
+		build,
+		"Testing some env variables: set",
+	)
+}
--- a/internal/integration/integration_test.go
+++ b/internal/integration/integration_test.go
@@ -1,35 +1,22 @@
 package integration_test
 
 import (
-	"bytes"
 	"context"
 	"embed"
 	"fmt"
 	"log"
-	"net/http"
 	"os"
 	"reflect"
-	"strconv"
 	"strings"
 	"sync"
 	"testing"
-	"text/template"
 	"time"
 
-	"github.com/Khan/genqlient/graphql"
 	"github.com/buildkite/agent-stack-k8s/v2/api"
 	"github.com/buildkite/agent-stack-k8s/v2/cmd/controller"
-	"github.com/buildkite/go-buildkite/v3/buildkite"
-	"github.com/buildkite/roko"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
-	"go.uber.org/zap"
-	"go.uber.org/zap/zaptest"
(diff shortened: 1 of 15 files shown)


internal/integration/integration_test.go
```diff
@@ -286,223 +273,20 @@ func TestCleanupOrphanedPipelines(t *testing.T) {
 	}
 }
 
-type testcase struct {
```
triarius, line 289: moved to `testcase_test.go`

sj26 (approved):
This is really nice. Love the tests and refactors. And this:

> Changed the pipeline to push agent/controller/helm chart on branches, so we can test them out

👌🏻

sj26:
How would I spin this up to give it a go, with that branch support? I probably should do that to properly approve 😅

internal/scheduler/scheduler.go
```diff
@@ -317,10 +326,39 @@ func (w *jobWrapper) Build() (*batchv1.Job, error) {
...
 				Value: "kubernetes-exec",
-			}, {
+			},
+			{
 				Name:  "BUILDKITE_CONTAINER_COUNT",
 				Value: strconv.Itoa(containerCount),
 			},
+			{
+				Name:  "BUILDKITE_AGENT_TAGS",
+				Value: createAgentTagString(agentTags),
```
sj26, line 336: ... wait, if we're using this env, why not pass the others in here as well?

triarius: So the env vars are passed as a `valueFrom` so the scheduler does not know their values. These are filled in by some part of Kubernetes by the time the containers in the pod are running.

Here's an excerpt of what the scheduler will output in yaml form:
```yaml
...
    spec:
      containers:
...
	  - name: agent
	    command: buildkite-agent
        args: start
	    env:
        - name: BUILDKITE_K8S_NODE
          valueFrom:
            fieldRef:
              fieldPath: spec.nodeName
...
```

Not only does the scheduler not what know that the value of `BUILDKITE_K8S_NODE` will be at this time, but for this particular example of the node name, it **can't** know because the ultimate decision about which node the pod is run on is made by [kube-scheduler](https://kubernetes.io/docs/reference/command-line-tools-reference/kube-scheduler/), not this scheduler. (For this reason I prefer to call this code path the "controller" and not the "scheduler", but as the filename is `scheduler.go`, that generally leads to more confusion.)

This controller just calls the Kubernetes API with the definition of the Kubernetes job. The YAML form of that job definition will contain the yaml snippet above.

For `BUILDKITE_K8S_NAMESPACE` and `BUILDKITE_K8S_SERVICE_ACCOUNT`, it is possible to inject them into the `--tags` argument here, but that would require making sure the source of truth for these is respected in this method, which will be fragile. There's less coupling between different parts of this very long method if we just set these once in this method and pick them up at runtime in the same way as the node name.

sj26: Ahhh, yeah I see. Sorry, I wasn't quite tracing that through in my head. But this makes sense. :+1:

sj26:
Nice! I still haven't spun it up, but the code looks good to me and the screenshot proves it works, so my approval stands, and don't wait for me :+1:

triarius:
Thanks, @sj26. If you could review https://github.com/buildkite/agent/pull/2003 too, I can merge that and update the submodule to resolve the merge conflict.

PR #139, opened 2023-03-09
Add nodeSelector to values.yaml schema

triarius:
This will allow setting the nodeSelector on the pod for the controller. This is a pretty standard
option in helm charts as it's part of the values.yaml created by the `helm create` command.

### DIFF
--- a/charts/agent-stack-k8s/templates/deployment.yaml.tpl
+++ b/charts/agent-stack-k8s/templates/deployment.yaml.tpl
@@ -16,6 +16,8 @@ spec:
         checksum/secrets: {{ include (print $.Template.BasePath "/secrets.yaml.tpl") . | sha256sum }}
     spec:
       serviceAccountName: {{ .Release.Name }}
+      nodeSelector:
+{{ toYaml $.Values.nodeSelector | indent 8 }}
       containers:
       - name: controller
         terminationMessagePolicy: FallbackToLogsOnError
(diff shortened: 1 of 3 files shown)


sj26 (approved):
So this allows constraining the nodes on which agent pods run on the k8s cluster? Neat.

I presume this won't flow to pods created for jobs?

Does this need an update in the README, too?

triarius:
> So this allows constraining the nodes on which agent pods run on the k8s cluster? Neat.
> 
> I presume this won't flow to (job) pods created for jobs?
> 
> Does this need an update in the README, too?

Not so much the "agent" pods. This just covers the one conatiner pod in agent-stack-k8s (the controller/scheduler). You can use nodeSelectors for all the containers involved in a job too. I've done it here for a non-trivial purpose here (building on top of the matrix interpolation for plugins you added): https://github.com/triarius/agent/blob/cd3f4ca508ac0403047538746e3f93ddc4516354/.buildkite/pipeline.yml#L34-L52

It's probably best to put the nodeSelector for the helm chart in a README for the helm chart itself. Usually it's a good practice to document all the options in a values.yaml there. A good example is: https://github.com/aws/karpenter/tree/main/charts/karpenter#readme.

The nodeSelectors aren't on the critical path to either creating a stack or running a job, so I don't think exposing them in the main README is necessary.

PR #140, opened 2023-03-09
Use sh as default shell

triarius:
Now that https://github.com/buildkite/agent/pull/1995 has merged, this should allow us to run
buildkite jobs where all the hooks set the hashbang to /bin/sh to run in a container without bash installed.


sj26 (approved):
Nice

PR #141, opened 2023-03-12
Fix deploy script

triarius:
A recent refactor broke the deploy script


.buildkite/steps/deploy.sh
```diff
@@ -3,13 +3,14 @@
 set -euo pipefail
 
 echo --- :hammer: Installing tools
-apk add helm yq skopeo git --quiet --no-progress
```
triarius, line 6: We don't need these anymore as we are not building the container in this script.

sj26 (approved):

PR #142, opened 2023-03-14
Fix repos and add linting and log grouping

triarius:
This is mainly a miscellaneous set of changes to the CI process.
However, a common source of truth for the repo names is needed.

### DIFF
--- a/.buildkite/steps/build_and_push.sh
+++ b/.buildkite/steps/build_and_push.sh
@@ -5,11 +5,7 @@ set -euo pipefail
 echo --- :hammer: Installing tools
 apk add helm yq skopeo git --quiet --no-progress
 
-tag=$(git describe)
-version=$(echo "$tag" | sed 's/v//')
-temp_agent_image=$(buildkite-agent meta-data get "agent-image")
-agent_image="${KO_DOCKER_REPO}/agent-k8s:${tag}"
-controller_image=$(buildkite-agent meta-data get "controller-image")
+. .buildkite/steps/repo_info.sh
 
 echo --- :docker: Logging into ghcr.io
 skopeo login ghcr.io -u "$REGISTRY_USERNAME" --password "$REGISTRY_PASSWORD" --authfile ~/.docker/config.json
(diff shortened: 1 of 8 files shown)


.buildkite/steps/build_and_push.sh
```diff
@@ -5,24 +5,18 @@ set -euo pipefail
 echo --- :hammer: Installing tools
 apk add helm yq skopeo git --quiet --no-progress
 
-tag=$(git describe)
-version=$(echo "$tag" | sed 's/v//')
-temp_agent_image=$(buildkite-agent meta-data get "agent-image")
-agent_image="${KO_DOCKER_REPO}/agent-k8s:${tag}"
-controller_image=$(buildkite-agent meta-data get "controller-image")
+. .buildkite/steps/repo_info.sh
```
moskyb, line 8: what does the preceding `.` do here? is it a fancy way of sourcing?

.buildkite/steps/controller.sh
```diff
@@ -2,19 +2,24 @@
...
+KO_VERSION=0.13.0
 OS=$(go env GOOS)
-ARCH=$(go env GOARCH)
-if [ "$ARCH" == "amd64" ]; then
-    ARCH="x86_64"
-fi
-latest=$(curl -L -s https://api.github.com/repos/ko-build/ko/releases/latest | jq -r '.tag_name')
-curl -sSfL "https://github.com/ko-build/ko/releases/download/${latest}/ko_${latest:1}_${OS^}_${ARCH}.tar.gz" | tar -xzv -C /bin ko
+ARCH=$(uname -m)
+curl -sSfL "https://github.com/ko-build/ko/releases/download/v${KO_VERSION}/ko_${KO_VERSION}_${OS^}_${ARCH}.tar.gz" | tar -xzv -C /bin ko
```
moskyb, line 13: i hadn't heard of [ko](https://github.com/ko-build/ko) before, cool stuff

moskyb (approved):
:shipit:

.buildkite/steps/build_and_push.sh
triarius: I think it's the original way of sourcing, `source` being a synonym for `.` is a later development of some shells (probably even prior to BASH) and not part of POSIX.

See https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html where is says the result of the command `source` is unspecified. Whereas `.` is a [builtin](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#dot)

.buildkite/steps/controller.sh
triarius: Could be nice to have in the agent pipeline... But we make a few different docker images and this does not have that flexibility.

PR #143, opened 2023-03-14
Add an informer to the controller that watches for pods in ImagePullBackOff and cancels the Buildkite job

triarius:
Container that are in ImagePullBackOff may eventually run when the image becomes avaliable.
The Buildkite job will run indefinately and the Kubernetes pod will be retriying the image pull.

I suspect most image pull backoff errors are due to errors in the pipeline definition, either a error with the image name or the credentials to pull it from a registry.
I can think of two situations where ImagePullBackOffs may occur that are not errors.

1. If the image is built in one stem of the pipeline and used in a container in another, then a explict dependency should be created between the step that pushes the image and the step that uses it. I don't think this is a feature of a good CI system to eagarly retry running jobs one of their dependencies has not been met.

2. Rate limiting of the ImagePull. In this case I think a more reliable CI system needs to be designed. Occasionally allowing the conatiners to be pulled when the rate limiter is not exhausted will lead to large variance in the duration of builds, which I don't think are desirable either. Failing fast when the pipeline in being rate limited will expose this quicker.

For this reason it is apropriate to fail job that have conatiners in ImagePullBackOff.


api/generated.go
```diff
@@ -361,6 +361,31 @@ const (
 	BuildStatesNotRun BuildStates = "NOT_RUN"
 )
 
+// CancelCommandJobJobTypeCommandCancelJobTypeCommandCancelPayload includes the requested fields of the GraphQL type JobTypeCommandCancelPayload.
+// The GraphQL type's documentation follows.
+//
+// Autogenerated return type of JobTypeCommandCancel
+type CancelCommandJobJobTypeCommandCancelJobTypeCommandCancelPayload struct {
```
moskyb, line 368: all-time generated type name right here

moskyb (approved):
LGTM! awesome work

PR #146, opened 2023-03-17
Downgrade client-go for version incompatibility

jmcshane:
Moving down to `client-go` 0.24 and `controller-runtime` 0.12 to support kubernetes 1.23 runtimes.

Tested against k8s versions 1.23-1.26 successfully.


triarius (approved):
Thanks for finding this and making a PR!

PR #147, opened 2023-03-24
Fix deploy script

triarius:
There was an extra `oci://` somewhere.

I also refactored some log groups and log output and made apk (the package manager for alpine linux) update its cache of the repos before installing new packages. Finally, I made sure that only the current version of the chart is pushed. While it's unlikely that there will be multiple versions packaged in the same container. I think this is a good defence in case someone retrofits that script to run elsewhere.

### DIFF
--- a/.buildkite/steps/build_and_push.sh
+++ b/.buildkite/steps/build_and_push.sh
@@ -1,4 +1,4 @@
-#!/bin/ash
+#!/usr/bin/env ash
 
 set -eufo pipefail
 
@@ -16,9 +16,9 @@ skopeo copy --multi-arch=all "docker://${temp_agent_image}" "docker://${agent_im
 echo --- :helm: Packaging helm chart
 yq -i ".image = \"$controller_image\"" charts/agent-stack-k8s/values.yaml
 yq -i ".config.image = \"$agent_image\"" charts/agent-stack-k8s/values.yaml
-helm package ./charts/agent-stack-k8s --app-version "$version" -d dist --version "$version"
+helm package charts/agent-stack-k8s --app-version "$version" -d dist --version "$version"
 
 echo --- :helm: Pushing helm chart to ghcr.io
-helm push ./dist/agent-stack-k8s-*.tgz "$helm_repo"
+helm push "dist/agent-stack-k8s-${version}.tgz" "$helm_repo"
 
 buildkite-agent annotate "Version: $version of the helm chart has been pushed to $helm_repo/agent-stack-k8s:$version"
--- a/.buildkite/steps/build_and_push.sh
+++ b/.buildkite/steps/build_and_push.sh
@@ -5,7 +5,7 @@ set -eufo pipefail
 echo --- :hammer: Installing tools
 apk add --update-cache --no-progress helm yq skopeo git
 
-. .buildkite/steps/repo_info.sh
+source .buildkite/steps/repo_info.sh
 
 echo --- :docker: Logging into ghcr.io
 skopeo login ghcr.io -u "$REGISTRY_USERNAME" --password "$REGISTRY_PASSWORD" --authfile ~/.docker/config.json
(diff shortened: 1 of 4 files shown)


.buildkite/steps/deploy.sh
```diff
@@ -3,12 +3,12 @@
 set -eufo pipefail
 
 echo --- :hammer: Installing tools
-apk add helm git --quiet --no-progress
+apk add --update-cache --no-progress helm git
 
 . .buildkite/steps/repo_info.sh
 
 echo --- :helm: Helm upgrade
-helm upgrade agent-stack-k8s "oci://${helm_repo}/agent-stack-k8s" \
```
triarius, line 11: `helm_repo` already has `oci://` in it.

.buildkite/steps/build_and_push.sh
```diff
@@ -1,9 +1,9 @@
-#!/bin/ash
+#!/usr/bin/env ash
 
-set -euo pipefail
+set -eufo pipefail
```
pda, line 3: I've never used `set -f` to disable pathname expansion.
My usual default is `set -e -o pipefail -u`.

Is the `-f` here for a specific reason? I see this further down the diff, eliminating the only(?) instance of pathname expansion in here:

```diff
-helm push ./dist/agent-stack-k8s-*.tgz "$helm_repo"
+helm push "dist/agent-stack-k8s-${version}.tgz" "$helm_repo"
```

But I'd be confused if I tried to add a wildcard somewhere later on and didn't realise `-f` was preventing it 🤔

Also I guess the `. .buildkite/steps/repo_info.sh` is subject to the same `set` rules… that also currently doesn't seem to do any pathname expansion, but it'd be confusing if some was added and it didn't work.

(Is it also possible to `source .buildkite/steps/repo_info.sh` rather than `. .buildkite/steps/repo_info.sh`… I seem to remember there's subtle differences between the two perhaps re. `$PATH` etc, and I don't know if `ash` and `bash` differ on that, but I find `source` much easier to read than `.`)

.buildkite/steps/tests.sh
```diff
@@ -2,7 +2,7 @@
 
 set -eufo pipefail
 
-echo "+++ Installing gotestsum :golang::test_tube:"
+echo "--- Installing gotestsum :golang::test_tube:"
```
pda, line 5: Nice UX improvement 👌🏼

pda (approved):

.buildkite/steps/build_and_push.sh
triarius: I've come round to the frame of mind that it's safer to be confused why the paths are not expanding than why files are missing or copied into places they should not be.

Usually, it's the case that not capturing all the files you want will result in an error, which may be hard to debug, but you will be forced to debug it.

Deleting/Uploading files you should not because you have a `*` somewhere is usually not an error. And you won't notice it if you don't inspect the logs, if it is in the logs at all.

triarius: You make a good point about the `set`s affecting the source. I'll move the set to after the `.` for this reasons.

Apparently `.` and `source` are synonyms in bash: https://askubuntu.com/a/25491

`source` isn't POSIX though. It might also be in `ash`, but as this is some weird busy box implementation of it, it could not be bothered checking.

It might be fine if it's compiled with the right flags though: https://github.com/brgl/busybox/blob/abbf17abccbf832365d9acf1c280369ba7d5f8b2/shell/ash.c#L9464-L9466

triarius: Looks like it is compiled with `BASH_SOURCE` in practice
```
🕙 16:14:55 ❯ docker run --rm -it --entrypoint ash alpine:3.17.2
/ # echo 'echo a' > a.sh
/ # source a.sh
a
/ #
```

triarius: So, past @triarius reminded me that you need to install `git` before sourcing the file. So I'm just going to put the `set`s in all the files to reduce confusion.

pda: 👍🏼

PR #148, opened 2023-03-26
Remove `v` from docker tags

triarius:
It's confusing that the helm chart image does not have a `v` in its tag, but the docker images do. The tags for the agent image on DockerHub do not have a `v` either. So consistency lies in removing the `v`. I also suspect it's not possible to add a `v` to the helm image tag.


pda:
I lean towards including the `v` prefix in version tags — strongly so in git, where I'd say it's de facto standard, but I'm less clear on the de facto standard / convention in container image tags 🤷‍♂️

triarius:
@pda The git tags will still have the `v`, the docker tags will not. That's how it is for the [agent repo](https://hub.docker.com/layers/buildkite/agent/3.45.0/images/sha256-49c7945c678130958630d287f5a3d18fb0fbb93e551591dc03c9044a412c5da7?context=explore), so I think the same pattern here is what we want. The practice in docker seems mixed: [alpine](https://hub.docker.com/layers/library/alpine/3.17.2/images/sha256-c41ab5c992deb4fe7e5da09f67a8804a46bd0592bfdf0b1847dde0e0889d2bff?context=explore) 
[ruby](https://hub.docker.com/layers/library/ruby/3.2.1/images/sha256-96ab211eea5e82fcc040c8cc66a388a2eaa976dd81be17dacd63a1ea55ce5d58?context=explore), [golang](https://hub.docker.com/layers/library/golang/1.20.2/images/sha256-2101aa981e68ab1e06e3d4ac35ae75ed122f0380e5331e3ae4ba7e811bf9d256?context=explore) do not have `v`, [traefik](https://hub.docker.com/layers/library/traefik/v2.10/images/sha256-f4b7fa08f26b46232a1750a80bf5882e912ee79660cc2995310082ec0eeb90b7?context=explore) does.

PR #151, opened 2023-03-27
Fix broken link in SSH configuration README section. 

clbx:
https://github.com/buildkite/agent-stack-k8s#cloning-repos-via-ssh

Looks like the markdown formatting for the ssh-env-config script is swapped.


triarius (approved):
Thanks for finding this 💖

PR #153, opened 2023-03-28
Use CI to ensure `go mod tidy` is a no-op

triarius:
This runs `go mod tidy` and then uses git to detect if anything has changed, only passing the job if it hasn't.
There is a complication caused by the fact that the checkout container will `chmod 777` the workspace, causing git to think
that the go.mod file has been changed to be executable. We are ignoring this for now, and we may remove the `chmod 777` behaviour later.


justfile
```diff
@@ -20,14 +20,15 @@ lint *FLAGS: gomod
...
 
 gomod:
-  #!/usr/bin/env bash
-  set -euo pipefail
+  #!/usr/bin/env sh
+  set -euf
+
   go mod tidy
-  git diff --no-ext-diff --exit-code go.mod go.sum
+  git diff -G. --no-ext-diff --exit-code go.mod go.sum
```
pda, line 31:
```
       -G<regex>
           Look for differences whose patch text contains added/removed lines that match <regex>.
```

I assume `-G.` means “only show diffs that match a regexp of at least one character”, making it ignore metadata-only changes like file permissions?

pda (approved):

justfile
triarius: Yep, I should have documented it in a comment, but I want to remove this ASAP

PR #154, opened 2023-03-29
Fix jobs are not acquired in when in a queue in a cluster

triarius:
While the the controller was only pulling jobs from the `kubernetes` queue, it was starting the agents in the `default` queue because that's the default, and it did not specify one. With a cluster token, jobs can no longer be acquired across queues, so the API would give a 404 when the agent tried to acquire the job.

So in this PR, we add the queue to the `BUILDKITE_AGENT_TAGS` env variable that will be passed into the `agent start` container, which will then acquire the job from the correct queue.

### DIFF
--- a/internal/scheduler/scheduler.go
+++ b/internal/scheduler/scheduler.go
@@ -320,6 +320,12 @@ func (w *jobWrapper) Build() (*batchv1.Job, error) {
 		},
 	}
 
+	if tag, err := agentTagFromJob(w.job); err != nil {
+		w.logger.Warn("error parsing job tags", zap.String("job", w.job.Uuid))
+	} else {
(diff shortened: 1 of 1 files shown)


tessereth (approved):
Seems sensible to me. I assume you've tested it and it works?

triarius:
@tessereth [This agent](https://buildkite.com/organizations/nepas-test-org/agents/01872c15-14cf-4c55-b585-584f94a17c59) acquired and ran a job in a cluster on Kubernetes and the controller was built from the [commit](https://github.com/buildkite/agent-stack-k8s/pull/154/commits/41106e96a9c46644f59df0ef1e5dd24ddfa0de4c) that this branch points to.

PR #156, opened 2023-04-06
Fix nullifying annotations when there are no labels

### LINKED ISSUE #155: Job does not get annotations from configuration
Somewhere in the Build function in the scheduler, annotations that are passed in via the plugin configuration are being lost. 

For example I have a configuration of:

```yaml
name: "test pipeline"
agents:
   queue: kubernetes
plugins:
- kubernetes:
     metadata:
        annotations:
              example.com/annotation-test: "value"
     podSpec:
       containers:
       - name: app
          image: debian
          command: ["echo", "hello world"]
```
The resultant job will not have the annotation provided, but will have the buildkite build-url and job-url annotations

I think they're being overwritten somewhere in here:
https://github.com/buildkite/agent-stack-k8s/blob/b5811c8fbb21b4ab4c7869a775cc08454696cc2a/internal/scheduler/scheduler.go#L177-L186

clbx:
Fixes #155 

Currently if there are no labels described in the configuration, annotations also get set to an empty map, nuking any annotations set in the configuration. 

https://github.com/buildkite/agent-stack-k8s/blob/b5811c8fbb21b4ab4c7869a775cc08454696cc2a/internal/scheduler/scheduler.go#L173-L176

This splits them up so that annotations are only set to an empty map if no annotations are described.

### DIFF
--- a/internal/scheduler/scheduler.go
+++ b/internal/scheduler/scheduler.go
@@ -170,10 +170,15 @@ func (w *jobWrapper) Build() (*batchv1.Job, error) {
 			},
 		}
 	}
+
 	if w.k8sPlugin.Metadata.Labels == nil {
 		w.k8sPlugin.Metadata.Labels = map[string]string{}
+	}
+
+	if w.k8sPlugin.Metadata.Annotations == nil {
 		w.k8sPlugin.Metadata.Annotations = map[string]string{}
 	}
+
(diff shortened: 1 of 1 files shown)


DrJosh9000 (approved):
LGTM ✅

Thanks for the contribution!

PR #160, opened 2023-04-23
Support Restricting Monitoring to a Single Cluster

triarius:
Previously, the controller would poll the GraphQL API for all jobs in an organization that matched certain tags. However, if the controller were only equipped with a Cluster token, the buildkite agent in the resulting pods would not be able to acquire the Buildkite Job from a different (or no) cluster.

In this PR, we change the GraphQL query to either restrict jobs to a configured Cluster UUID, or if no Cluster UUID is in the config, to jobs NOT belonging to a cluster.

There is also some improvements to how the GraphQL client is generated. In particular, there is a recipe for downloading the schema file that may be used by those with access to the buildkite repositories.


api/api.go
```diff
@@ -1,7 +1,5 @@
 package api
 
-//go:generate go run github.com/Khan/genqlient
-
```
triarius, line 4: Moved to `api/client.go` as it is what uses the generated code.

api/schema.graphql
triarius, line 1: Copied from the bk/bk source code. I have no idea why it is different.

triarius, line 1: Copied from the bk/bk source code. I have no idea why it is so different. The diff only needs to be the new arguments to the job field, allowing filtering by cluster.

internal/monitor/monitor.go
```diff
@@ -46,42 +59,89 @@ func New(logger *zap.Logger, k8s kubernetes.Interface, cfg api.Config) (*Monitor
...
+
+		if jobsResp.Organization.Id == nil {
+			return nil, true, fmt.Errorf("invalid organization: %s", m.cfg.Org)
+		}
+
+		jobs = make([]*api.JobJobTypeCommand, 0, len(jobsResp.Organization.Jobs.Edges))
+		for _, job := range jobsResp.Organization.Jobs.Edges {
+			jobs = append(jobs, job.Node.(*api.JobJobTypeCommand))
+		}
+	} else {
```
DrJosh9000, line 88: [Optional] Probably worth considering dividing the the method into two sub-methods, one for each of the branches.

DrJosh9000 (commented):
Looking good!

internal/monitor/monitor.go
```diff
@@ -46,42 +59,89 @@ func New(logger *zap.Logger, k8s kubernetes.Interface, cfg api.Config) (*Monitor
 	}, nil
 }
 
+func (m *Monitor) getScheduledCommandJobs(
+	ctx context.Context,
+	tag string,
+) (
+	jobs []*api.JobJobTypeCommand,
+	fatal bool,
```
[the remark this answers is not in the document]
triarius: This is nice, but I've changed things to not need it.

internal/monitor/monitor.go
triarius: I think this is more of a symptom of how the generated methods return 2 struct types, `GetScheduledJobsReponse` and `GetScheduledJobsClusteredResponse` that have the same structure and need to be handled identically, but the type system won't let us write only one block of code to handle both.

I decided to rectify this using an interface to abstract over the parts of the structs we need. There's still a bit of duplication, but it is limited to identical implementations of the interface methods only do "one thing". If go had a more expressive syntax, both of these methods could be one-liners. In any case, this should be earlier to maintain that larger duplicated blocks that do multiple things.

DrJosh9000 (approved):
Nice! 🍕

PR #161, opened 2023-04-23
Refactor packages to make controller more explicit

triarius:
Consolidate package for the `Config` object.

It was confusing me that the Config for the controller lived in the `api` package. And there was no `controller` package at all. So I created one.


DrJosh9000 (approved):
🙌

PR #162, opened 2023-04-26
Fix controller may steal jobs when configured with multiple agent tags

triarius:
This fixes an observed issue where `agents-stack-k8s` could run jobs that it should not have according to the [targeting rules](https://buildkite.com/docs/agent/v3/cli-start#agent-targeting) in the docs. This is demonstrated and explained further below. There is also a change to how the tags become k8s labels that is discussed at the end.

### Outside `agent-stack-k8s`

Tags are key-value pairs that may be applied to an agent and also to a job. These are used to target which jobs an agent may run.

For an agent to pick up a job, [ALL the tags on the job must be present on the agent](https://buildkite.com/docs/agent/v3/cli-start#agent-targeting), and the values have to match (up to wildcards being allowed on the tags for the job).

#### Example 1
For example, if there is an agent with the tags
```yaml
cats: "true"
queue: default
``` 
and there is a scheduled job with the tags
```yaml
queue: default
```
Then the agent can run the job.

#### Example 2
However, if it were the other way around, and the job's tags were
```yaml
cats: "true"
queue: default
```
and the agent's tag were
```yaml
queue: default
```
Then the agent could not run the job as not all the job's tags matched the agent's.

#### Example 3
If job's tags were
```yaml
cats: "true"
queue: default
```
and the agent's tag were
```yaml
cats: "true"
queue: not-default
```
Then the agent should not run the job either.

### In `agent-stack-k8s`
The agents created by agent-stack-k8s did not follow this. In `internal/controller/monitor/monitor.go`, there would be a separate jobs query for each tag the controller configured with. While these query do return jobs that have ALL these tags, these jobs could have other tags not specified in the query. So, in the situation of example 3 above, if the query were for jobs with `cats: "true"`, then the job in that example would be run.

This is fixed by [425708d](https://github.com/buildkite/agent-stack-k8s/pull/162/commits/425708d9c0a4b154ebb449fdda9a9aa212394fbf). In that commit, the queries are changed to specify ALL tags that the stack was launched with, so the job in example 3 would not be returned by the jobs query.

However, there is a further problem that arise in the situation in example 2. If the stack was started with just the tag `queue: default`, then the jobs query would return the job in example 2, and the stack would run it. For this reason, in [99be029](https://github.com/buildkite/agent-stack-k8s/pull/162/commits/99be0296e7b679d31d7b7bea1105bc8dd3d4f3fa), we reject jobs returned by the query whose tags do not match all the tags for a job return by the query.

## Tags to Labels
Previously, each K8s Job would be labelled with a single Buildkite tag (even if the stack was listening for multiple tags) as each job was acquired in response to a query for a single tag. So if that tag were `queue: kubernetes`, then the label would be
```yaml
metadata:
  labels:
    buildkite.com/job-tag: queue_kubernetes

```
Now, when the controller receives the job definition, it will apply a label for each tag on the job, so that if the job was from a step with
```yaml
agents:
  queue: kubernetes
  cats: "true"
```
then the new k8s job will have:
```yaml
metadata:
  labels:
    tag.buildkite.com/queue: kubernetes
    tag.buildkite.com/cats: "true"
```


go.mod
```diff
@@ -33,6 +33,7 @@ require (
 	github.com/google/shlex v0.0.0-20191202100458-e7afc7fbc510 // indirect
 	github.com/mattn/go-colorable v0.1.12 // indirect
 	github.com/mattn/go-isatty v0.0.14 // indirect
+	golang.org/x/exp v0.0.0-20230425010034-47ecfdc1ba53 // indirect
```
DrJosh9000, line 36: [No action required in this PR] It's not great that this is depending on exp.

internal/controller/agenttags/tags.go
```diff
@@ -0,0 +1,72 @@
...
+
+	}
+	return m, errs
+}
+
+func mapToLabels(m map[string]string) (map[string]string, []error) {
+	labels := map[string]string{}
+	errs := []error{}
+	for k, v := range m {
+		namespacedKey := "buildkite.com/" + k
```
DrJosh9000, line 30: Is it worth sub-namespacing inside "buildkite.com/"? Should namespacedKey look more like "buildkite.com/tag/key"?

internal/controller/monitor/monitor.go
```diff
@@ -103,40 +99,58 @@ func (m *Monitor) Start(ctx context.Context, handler JobHandler) <-chan error {
...
-					}); err != nil {
-						logger.Error("failed to create job", zap.Error(err))
+			for _, job := range jobs {
+				var jobTags map[string]string
+				{
+					var tagErrs []error
+					if jobTags, tagErrs = agenttags.ToMap(job.AgentQueryRules); len(tagErrs) != 0 {
+						logger.Warn("making a map of agent tag in job response", zap.Errors("err", tagErrs))
 					}
 				}
```
DrJosh9000, line 141:
```suggestion
				jobTags, tagErrs := agenttags.ToMap(job.AgentQueryRules)
				if len(tagErrs) != 0 {
					logger.Warn("making a map of agent tag in job response", zap.Errors("err", tagErrs))
				}
```

internal/controller/agenttags/tags.go
```diff
@@ -0,0 +1,72 @@
...
+
+import (
+	"errors"
+	"fmt"
+	"strings"
+
+	"k8s.io/apimachinery/pkg/util/validation"
+)
+
+func ToMap(tags []string) (map[string]string, []error) {
```
DrJosh9000, line 11: [Opinion] Exported identifiers should have doc comments ... though, admittedly, this one is pretty straightforward.

[Opinion dressed as a leading question] Is it more idomatic for a function to expose errors through `[]error` or `error`?

internal/controller/monitor/monitor.go
```diff
@@ -103,40 +99,58 @@ func (m *Monitor) Start(ctx context.Context, handler JobHandler) <-chan error {
 	logger := m.logger.With(zap.String("org", m.cfg.Org))
 	errs := make(chan error, 1)
 
+	var agentTags map[string]string
+	{
+		var tagErrs []error
+		if agentTags, tagErrs = agenttags.ToMap(m.cfg.Tags); len(tagErrs) != 0 {
+			logger.Warn("making a map of agent tags", zap.Errors("err", tagErrs))
+		}
+	}
```
DrJosh9000, line 108: What about:

```suggestion
	agentTags, tagErrs := agenttags.ToMap(m.cfg.Tags)
	if len(tagErrs) != 0 {
		logger.Warn("making a map of agent tags", zap.Errors("err", tagErrs))
	}
```

DrJosh9000 (commented):
🧐📋

internal/controller/agenttags/tags.go
triarius: In this case, I don't want to return on the first error, so some form of error aggregation is needed. I think it's the caller's responsibility to decide what to do with the individual errors, for example the caller could decide to log each error on in a new log message, or combine them into one. Thus, I don't think accumulating them and returning the output of `errors.Join` is right here. I could make a custom error type to house the errors, and then the caller can coerce that back to `[]errors` if it wants. But then the caller needs to delve into the implementation of the `ToMap` method to know to expect the custom error type to get the individual errors.

So I think the exposing `[]errors` in the function interface is the most idiomatic way to capture this situation.

internal/controller/agenttags/tags.go
triarius: The bit after the first `/` can't contain a `/`. It has to be alphanumeric or `-` or `_`, starting and ending in an alphanumeric. So it perhaps would have to be `buildkite.com/tag-key` or `buildkite.com/tag_key`. Not a fan of either of these TBH. How about `tags.buildkite.com/key`?

internal/controller/monitor/monitor.go
triarius: I don't want the `tagErrs` value to leak outside this scope. I personally don't think using curly braces to isolate variables' scopes is unusual, but I suppose it's pretty cheap to put it into a function to avoid raising any eyebrows, though, so I will do that instead.

triarius: It's gone now

internal/controller/monitor/monitor.go
triarius: It's gone now

internal/controller/agenttags/tags.go
DrJosh9000: Hmm. I guess `tags.buildkite.com/key` is the best option.

DrJosh9000 (approved):
Looking pretty good!

PR #163, opened 2023-04-27
This example doesn't work

jmcshane:
Alpine doesn't have `/bin/bash` so this sample pipeline would fail. Changing to ubuntu would make this sample work.

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,7 +77,7 @@ steps:
       - kubernetes:
           podSpec:
             containers:
-              - image: alpine:latest
+              - image: ubuntu:latest
                 command: [echo]
                 args:
                 - "Hello, world!"


triarius:
Hi @jmcshane 

It should work on the current versions of the stack (v0.5.1).

The stack [has been setting](https://github.com/buildkite/agent-stack-k8s/pull/140) `BUILDKITE_SHELL` to be `/bin/sh`, which should be good enough to run commands when bash is not available. For hooks, since https://github.com/buildkite/agent/pull/1995, the agent will fall back to `/bin/sh` if bash is not found.

You can see it in action here: https://buildkite.com/buildkite-kubernetes-stack/kubernetes-agent-stack/builds/690#0187c4f5-4c22-44a9-b4d8-1f4f4bfa6e3f/44

jmcshane:
thanks, sorry, was running with an old version.

PR #164, opened 2023-04-29
Change the ClusterRole to a Role

triarius:
Previously, the controller was given permission to crud jobs in any namespace in the cluster. This is too broad, so this PR isolates this into one namespace.

I also moved the creation of the informer factory to the `controller` package as it is called from there.


DrJosh9000 (approved):
LGTM 🪴

PR #165, opened 2023-04-29
Make line length shorter

triarius:
Some of the lines are too long. I was trying to fix some issue and I had a hard time deciphering what some of the code was doing because the lines were wrapped at arbitrary locations.
I ran [golines](https://github.com/segmentio/golines) and [gofumpt](https://github.com/segmentio/golines) over all the files.

### DIFF
--- a/internal/controller/config/config.go
+++ b/internal/controller/config/config.go
@@ -38,15 +38,15 @@ func JobName(uuid string) string {
 }
 
 type Config struct {
-	AgentTokenSecret string `mapstructure:"agent-token-secret" validate:"required"`
-	BuildkiteToken   string `mapstructure:"buildkite-token"    validate:"required"`
-	Debug            bool
-	Image            string        `                                  validate:"required"`
+	Debug            bool          `mapstructure:"debug"`
+	AgentTokenSecret string        `mapstructure:"agent-token-secret" validate:"required"`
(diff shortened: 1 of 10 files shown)


internal/controller/config/config.go
```diff
@@ -39,16 +39,16 @@ func JobName(uuid string) string {
 
 type Config struct {
 	AgentTokenSecret string `mapstructure:"agent-token-secret" validate:"required"`
-	BuildkiteToken   string `mapstructure:"buildkite-token" validate:"required"`
+	BuildkiteToken   string `mapstructure:"buildkite-token"    validate:"required"`
 	Debug            bool
-	Image            string        `validate:"required"`
+	Image            string        `                                  validate:"required"`
```
DrJosh9000, line 44: Is `mapstructure` needed before `validate`? Perhaps they could be swapped, to avoid the huge blanks.

triarius: There's a case where there's no `validate` tag, but there is a `mapstructure` tag. Rather bizarrely, mapstructure is case-insensitive: https://pkg.go.dev/github.com/mitchellh/mapstructure?utm_source=godoc#hdr-Field_Tags.

So I think it's best to always explicitly specify lowercase kebab case field names. While this could break something, we only use these for integration with viper and cobra in cmd/controller/controller.go, and there, all the field names are lower kebab case.

DrJosh9000 (approved):

Issue #167, opened 2023-05-17
Support externally created secrets


abhishekmukherg:
Currently, the agent-stack-k8s requires that the secret material for both `agentToken` and `graphqlToken` are provided to helm, for helm to templatize the data into a `Secret` object. This is incompatible with other secret management solutions like [External Secrets Operator](https://external-secrets.io/v0.8.1/). In these solutions, you would want to tell helm where to look for the secret and such, and allow the operator to create the actual Secret object. And very specifically for us, doing so would imply committing those rendered files to git, as we leverage a workflow as outlined in [Robbie's talk at Unblock a few year ago](https://www.youtube.com/watch?v=wPkHS4srJtY)

## Example

As an example for this workflow, you might have in your chart, that leverages `agent-stack-k8s` as a subchart, a template like

```yaml
apiVersion: external-secrets.io/v1beta1
kind: ExternalSecret
metadata:
  name: template
spec:
  # ...
  target:
    name: my-secret
    creationPolicy: Owner
  data:
  - secretKey: agent-token
    remoteRef:
      key: /agent-token
```

The ESO will come in and fetch the secret from an external store (vault, AWS Secrets Manager) and create `my-secret` Secret from this `template` .

You could then reference it in a helm chart with

```yaml
secret:
  create: false
  existingSecret: my-secret
  agentTokenKey: agent-token
```

## Prior Art

This is effectively an ask to reproduce the work in https://github.com/buildkite/charts/pull/96 into the new chart

## Acceptance Criteria i.e. TLDR

Secrets for `agentToken` and `graphqlToken` can be specified via a combination of an externally created secret name, and a key within that Secret

abhishekmukherg:
Note that this is an easy enough change to implement that I can do it, but wanted to get a ticket up as it's not obvious what the DSL should look like given the current values DSL. As in, what does this look like in the future:

```yaml
agentToken: foo
graphqlToken: bar
```

triarius:
Thanks for bringing this up @abhishekmukherg, this is a much-needed improvement. We're happy for you to have a go at implementing this.

Perhaps the template can be flexible enough to check if `agentToken` and `graphqlToken` are strings or objects that reference ExternalSecrets. I think you should be able to put a giant if condition around `charts/agent-stack-k8s/templates/secrets.yaml.tpl` so that, in the later case, it is an empty YAML document.

abhishekmukherg:
Just to get it on paper, I think you're suggesting

```yaml
[...]

PR #171, opened 2023-06-07
Enable secrets to be referenced in the install namespace so this can be installed with tools like argo

jmcshane:
This came up in a conversation with a team using Buildkite right now and felt like a good pattern to support.

If this is installed via argo, you would likely have another mechanism that would install the secret in the namespace, such as external secrets operator. We should enable this secret to be pre-provisioned and referenced in the `envFrom` block by name.


triarius (approved):
Thanks, @jmcshane, and apologies for taking so long to review this, everyone.

I suggested a different approach in https://github.com/buildkite/agent-stack-k8s/issues/167, but this is delivering value from today (or 3 weeks ago 😅) so I think we can merge it now and tweak things later.

PR #173, opened 2023-06-23
Revert "add optional labels"

triarius:
Reverts buildkite/agent-stack-k8s#172

This failed to upgrade an existing helm release of agent-stack-k8s with the following error:
```
Error: UPGRADE FAILED: cannot patch "agent-stack-k8s" with kind Deployment: Deployment.apps "agent-stack-k8s" is invalid: spec.template.metadata.labels: Invalid value: map[string]string{"app":""}: `selector` does not match template `labels`
```

It seems we need the deployment's `spec.selector.matchLabels` object to be a sub-object of `template.metadata.labels`. The chart used to guarantee this (it hard coded that they were the same), but that's not the case with this PR, so I'm reverting it.

Also, I think we should preserve the label `app: {{ .Release.Name }}`. So, for resubmission, I think we should merge the labels specified in `.Values.labels` with the object `app: {{ .Release.Name }}`.

@jiaquan1 we'll be happy to accept another PR that does this.

### DIFF
--- a/charts/agent-stack-k8s/templates/deployment.yaml.tpl
+++ b/charts/agent-stack-k8s/templates/deployment.yaml.tpl
@@ -10,7 +10,7 @@ spec:
   template:
     metadata:
       labels:
-        {{- toYaml $.Values.labels | nindent 8 }}
+        app: {{ .Release.Name }}
       annotations:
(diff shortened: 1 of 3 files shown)


sj26 (approved):
Sounds reasonable :+1:

PR #176, opened 2023-07-07
Fix helm template with default value for the label

triarius:
The default value for `.Values.label` is `{}`, but is always another key, `app`, so the template renders as
```yaml
labels:
   {}
   app: foo
```
which is not valid yaml.

In this PR we merge the object provided by the user as `.Values.label` with an object of mandatory label, which the user cannot modify.

These mandatory labels are used as selectors for the deployment, so the user should not have too much control over them.


DrJosh9000 (approved):
Makes sense, also I like the `{{-` tidyups

PR #179, opened 2023-07-10
remove runasnonroot

jiaquan1:


triarius:
@jiaquan1 As we discussed on the call, the controller is able to run as non-root, so I'm going to close this.

PR #184, opened 2023-07-18
Automate releases

triarius:
The new step will run on buildkite tag builds and create a draft release on GitHub.
It uses goreleaser, ghch, and github cli. A human will then be able to edit the draft and publish the release.

There was a flaw in the previous release process that meant the controller and container images reported their versions to be different from the release. This was because when the images were built, the tag had not yet been created. The new processes fixes this by creating the release from the tag build rather than creating the tag from the release.


DrJosh9000 (approved):
Nice!

PR #185, opened 2023-07-19
spinnaker ver false

jiaquan1:
Spinnaker added a version as the suffix in the secret name, confused the container to get the right secrets.


moskyb:
g'day @jiaquan1! unfortunately we won't be able to accept this PR, as spinnaker isn't something that all users use.

if you made a PR allowing people to put in arbitrary annotations (similar to #174 and #176), i think that would allow you to work around this?

PR #187, opened 2023-07-23
Fix cleaning stray pipelines from tests

triarius:
There is a special integration test that is used to delete pipeline from integration tests that have not been cleaned up. I'm not sure that it needs to be a test, but for now, let's keep it that way.

There needed to be some changes to get it to work again, so I made them. I also separated it from the other tests.

It may be invoked from a command line with a k8s context into the cluster's namespace with
```
just cleanup-orphans --org=<org slug of k8s stack> --buildkite-token=<graphql token>
```


sj26:
"just cleanup-orphans" is very funny to me 😅

Just .. solve poverty!

Slightly dark 😅

sj26 (approved):
> There needed to be some changes to get it to work again ...

Is this only used for local dev/test? I don't see it exercised by the regular pipeline.

This seems to be test suite only, and looks good from my limited Go depth. I'll give you a +1, but defer for a Go-er brain if you want a detailed code review.

triarius:
Yep, it is skipped unless the tests are run with a certain flag (`--delete-orphaned-pipelines`). I'm not sure that we want to keep this approach. It's not a test at all, just a script to clean up after tests are run. The main reason (I presume) to write it in go and not bash is that we can re-use the machinery to call the graphql api that's in this code base.

But it also adds unnecessary code to the graphql client (the mutations to delete pipelines, which the controller will never need). Another oddity is that it fails if not run in a Kubernetes context, but reading the code, it only seems to call the graphql api, so should be independent of what's happening in Kubernetes. However, arguably, it should also be extended to delete the resources in Kubernetes created by the integration tests. Currently, these are quite intertwined with the other jobs run by this pipeline. I'm working (in my spare time) on isolating the integration tests into their own Kubernetes namespace, so they should be easier to clean up after that is done.

Either way, I think it should not share a code base with the reset of the stack. It should be a standalone script, possibly, but not necessarily, written in go.

PR #188, opened 2023-07-24
Fix deploy step did not have a key

triarius:
Also, the release step did not have a key, but that don't matter as
noting depended on it. The deploy step no longer needs an explicit
dependence on the integration tests as it depends on them transitively
through `push-main-or-tag`.


triarius:
Merging this to get the release process working. I will need to iterate on the main branch to do so, so there's likely more such force merges incoming. Feel free to post-merge review.

PR #196, opened 2023-08-14
adding tolerations parameters to the helm chart

kingatlas:
same as `nodeSelector` parameter, we need a parameter to specify `tolerations` when deploying the agent-stack-k8s pods.
this can be required if we want to dedicate specific k8s nodes to run buildkite agent controller


triarius (approved):
Thanks, @kingatlas LGTM!

PR #198, opened 2023-08-16
Adding extraVolumeMounts option

kingatlas:
This PR adds the possibility to add extra `volumeMounts` to all the containers of the agent Pod. including `checkout` container (or other system containers).
This is needed for a specific use case: having a shared folder between all `checkout` containers, in order to use [git mirrors](https://buildkite.com/docs/agent/v3#promoted-experiments-git-mirrors)  feature.
The idea is to have a shared folder (per host), and to have all `checkout` containers source the git code from, using git mirrors. This is critical for situations where the `git clone` takes several minutes.

**Note**: I've added an integration test, but I don't see clearly how the integration tests can be executed (step by step).

### DIFF
--- a/internal/integration/fixtures/extra-volume-mounts.yaml
+++ b/internal/integration/fixtures/extra-volume-mounts.yaml
@@ -13,12 +13,12 @@ steps:
                 command: [ash]
                 args:
                   -  -c
-                  - "echo 'volume mounted' > /tmp/buildkite-git-mirrors/foo.txt"
+                  - '"echo volume mounted > /tmp/buildkite-git-mirrors/foo.txt"'
               - image: alpine:latest
                 command: [ash]
                 args:
                   - -c
-                  - "sleep 2; cat /tmp/buildkite-git-mirrors/foo.txt"
+                  - '"sleep 2; cat /tmp/buildkite-git-mirrors/foo.txt"'
             volumes:
               - name: host-volume
                 hostPath:
--- a/internal/integration/fixtures/extra-volume-mounts.yaml
+++ b/internal/integration/fixtures/extra-volume-mounts.yaml
@@ -3,7 +3,6 @@ steps:
     agents:
       queue: {{.queue}}
     env:
-      BUILDKITE_SHELL: /bin/sh -e -c
       BUILDKITE_GIT_CLONE_MIRROR_FLAGS: --mirror
       BUILDKITE_GIT_MIRRORS_PATH: /tmp/buildkite-git-mirrors
     plugins:
@@ -11,11 +10,15 @@ steps:
           podSpec:
             containers:
               - image: alpine:latest
-                command: [bash]
-                args: [-c "echo 'volume mounted' > /tmp/buildkite-git-mirrors/foo.txt"]
+                command: [ash]
+                args:
+                  -  -c
+                  - "echo 'volume mounted' > /tmp/buildkite-git-mirrors/foo.txt"
               - image: alpine:latest
-                command: [bash]
-                args: ["sleep 2; cat /tmp/buildkite-git-mirrors/foo.txt"]
+                command: [ash]
+                args:
+                  - -c
+                  - "sleep 2; cat /tmp/buildkite-git-mirrors/foo.txt"
             volumes:
               - name: host-volume
                 hostPath:
--- a/internal/integration/fixtures/extra-volume-mounts.yaml
+++ b/internal/integration/fixtures/extra-volume-mounts.yaml
@@ -11,11 +11,15 @@ steps:
           podSpec:
             containers:
               - image: alpine:latest
-                command: [touch]
-                args: [/tmp/buildkite-git-mirrors/somefile.txt]
+                command: [bash]
+                args: [-c "echo 'volume mounted' > /tmp/buildkite-git-mirrors/foo.txt"]
+              - image: alpine:latest
+                command: [bash]
+                args: ["sleep 2; cat /tmp/buildkite-git-mirrors/foo.txt"]
+            volumes:
               - name: host-volume
                 hostPath:
-                  path: /host/shared/folder
+                  path: /tmp
                   type: ''
           extraVolumeMounts:
               - name: host-volume
--- a/README.md
+++ b/README.md
@@ -127,6 +127,13 @@ Sidecar containers can be added to your job by specifying them under the top-lev
 
 There is no guarantee that your sidecars will have started before your job, so using retries or a tool like [wait-for-it](https://github.com/vishnubob/wait-for-it) is a good idea to avoid flaky tests.
 
+### Extra volume mounts
+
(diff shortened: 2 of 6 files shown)


internal/integration/fixtures/extra-volume-mounts.yaml
```diff
@@ -0,0 +1,27 @@
+steps:
+  - label: ":wave:"
+    agents:
+      queue: {{.queue}}
+    env:
+      BUILDKITE_SHELL: /bin/sh -e -c
```
triarius, line 6:
```suggestion
```
we don't need this anymore. It used to be necessary for alpine containers, (and it still is), but now, the controller injects this by default: https://github.com/kingatlas/agent-stack-k8s/blob/878c49c75a3af0cb4b0c8ccca1232409e820b42f/internal/controller/scheduler/scheduler.go#L258-L261

It's still around in older integration tests though, but that's due to inertia.

internal/integration/fixtures/extra-volume-mounts.yaml
```diff
@@ -0,0 +1,27 @@
...
+      BUILDKITE_GIT_CLONE_MIRROR_FLAGS: --mirror
+      BUILDKITE_GIT_MIRRORS_PATH: /tmp/buildkite-git-mirrors
+    plugins:
+      - kubernetes:
+          podSpec:
+            containers:
+              - image: alpine:latest
+                command: [bash]
```
triarius, line 14:
```suggestion
                command: [ash]
```

internal/integration/fixtures/extra-volume-mounts.yaml
```diff
@@ -0,0 +1,27 @@
...
+                args: [-c "echo 'volume mounted' > /tmp/buildkite-git-mirrors/foo.txt"]
+              - image: alpine:latest
+                command: [bash]
```
triarius, line 17:
```suggestion
                command: [ash]
```

internal/integration/fixtures/extra-volume-mounts.yaml
triarius: Bash isn't present on alpine, which might be why this integration test is failing.

internal/integration/fixtures/extra-volume-mounts.yaml
triarius: Ibid.

triarius (changes requested):
This is a brilliant idea @kingatlas, we would be glad to add the ability to use git mirrors with agent-stack-k8s.

It is true that integration tests are difficult to run. We have work scheduled in our backlog to improve the situation, but that requires a bit of a redesign of them.

So for now, the best we can do is for me to run the pipeline for you. And when I did, the integration test you added failed with the error:

```


Error:      	Not equal:
--


        	            	+(api.BuildStates) (len=6) "FAILED"
```

So it looks like the jobs started, but failed.

I suspect the fact that `bash` is not installed on alpine linux containers is the reason. You may use `ash` [instead](https://unix.stackexchange.com/a/692917/328750).

kingatlas:
Thank you @triarius for the feedback, 
I've made the requested changes, and added a missing `"-c"` for the second container command.
Please let me know if that works. Otherwise I would be glad to follow any other instructions to run the tests on my end.

internal/integration/fixtures/extra-volume-mounts.yaml
```diff
@@ -0,0 +1,30 @@
...
+              - image: alpine:latest
+                command: [ash]
+                args:
+                  -  -c
+                  - "echo 'volume mounted' > /tmp/buildkite-git-mirrors/foo.txt"
+              - image: alpine:latest
+                command: [ash]
+                args:
+                  - -c
+                  - "sleep 2; cat /tmp/buildkite-git-mirrors/foo.txt"
```
triarius, line 21:
```suggestion
                  - '"sleep 2; cat /tmp/buildkite-git-mirrors/foo.txt"'
```
Turns out, when the YAML is parsed, the double quotes are dropped. But they are needed so that `ash -c` receives one argument. I think escaping them will fix this.

internal/integration/fixtures/extra-volume-mounts.yaml
```diff
@@ -0,0 +1,30 @@
...
+      BUILDKITE_GIT_MIRRORS_PATH: /tmp/buildkite-git-mirrors
+    plugins:
+      - kubernetes:
+          podSpec:
+            containers:
+              - image: alpine:latest
+                command: [ash]
+                args:
+                  -  -c
+                  - "echo 'volume mounted' > /tmp/buildkite-git-mirrors/foo.txt"
```
triarius, line 16:
```suggestion
                  - '"echo volume mounted > /tmp/buildkite-git-mirrors/foo.txt"'
```
Turns out, when the YAML is parsed, the double quotes are dropped. But they are needed so that `ash -c` receives a single argument.

`echo` is perfectly capable of taking multiple arguments, so the single quotes aren't necessary, i.e., `echo 'hello world'` and `echo hello world` produce the same output.

So, I think if we remove the single quotes from inside the string, we can use single quotes to escape the double quotes that we need to be in the string.

triarius (commented):
Sorry @kingatlas I ended up hijacking your PR when trying to debug the tests. Apparently, GitHub allows upstream repo maintainers to push commits to downstream repos if you open a PR allowing maintainer edits: see the red warning [here](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork).

Once I did so, I found out that your tests were failing with this in the job logs:
```shell
BusyBox v1.36.1 (2023-07-27 17:12:24 UTC) multi-call binary.

Usage: sleep [N]...

Pause for a time equal to the total of the args given, where each arg can
have an optional suffix of (s)econds, (m)inutes, (h)ours, or (d)ays
```

internal/integration/fixtures/extra-volume-mounts.yaml
kingatlas: thank you @triarius for spending time on this.
I have pushed with the recommended syntax.
besides that, I'm not sure to understand what is blocking you from pushing to my downstream repo, I checked the PR and it has "Allow edits by maintainers" checked. I've also added you as a collaborator in the meantime.

triarius: Sorry, my message was a bit unclear:
1. I did have the ability to push to your downstream repo.
2. I did push commits to it.
3. I force pushed those commits away.
4. Adding me as a collaborator was not necessary, as you always had "Allow edits by maintainers" checked

I only brought it up because I think it could be a security risk for you to allow me to push commits to your downstream repo without explicitly allowing me to. But now that you added me as a collaborator, this point is moot.

BTW, feel free to remove me.

triarius (approved):
I think that did the trick 🎉

PR #199, opened 2023-08-17
revert the metadata change in rbac subjects

jiaquan1:
This is to revert the changes in the rbac subjects. Customized additional properties in metadata is not always allowed in rbac subjects, such as `annotations`, so revert the changes in the rbac subjects. Got the following error when helm install: 
```
error validating data: ValidationError(RoleBinding.subjects[0]): unknown field "annotations" in io.k8s.api.rbac.v1.Subject
```
Tested it works after revert the subject.


triarius (approved):
LGTM

PR #200, opened 2023-08-21
Add flag to preserve test pipelines

triarius:
Each integration test creates a new pipeline in the `buildkite-kubernetes-stack` organization. But these are deleted when the tests finish. It would be useful to debug failing integration tests if there were an option to not do this.

In this PR we do this if the environment has `TEST_PRESERVE_PIPELINE=true`

### DIFF
--- a/.buildkite/steps/tests.sh
+++ b/.buildkite/steps/tests.sh
@@ -11,11 +11,18 @@ branch="${BUILDKITE_BRANCH:-main}"
 IMAGE=$(buildkite-agent meta-data get "agent-image")
 export IMAGE
 
-gotestsum \
-  --junitfile "junit-${BUILDKITE_JOB_ID}.xml" \
+args=(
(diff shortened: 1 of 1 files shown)


.buildkite/steps/tests.sh
```diff
@@ -11,11 +11,18 @@ branch="${BUILDKITE_BRANCH:-main}"
...
 
-gotestsum \
-  --junitfile "junit-${BUILDKITE_JOB_ID}.xml" \
+args=(
+  --junitfile "junit-${BUILDKITE_JOB_ID}.xml"
   -- \
   -count=1 \
   -failfast \
   -ldflags="-X ${package}.branch=${branch}" \
-  "$@" \
```
triarius, line 20: Unfortunate to use this, but it's not used in CI. It's easy enough to run test locally through the `justfile` directly.

DrJosh9000 (approved):

PR #201, opened 2023-08-30
Log collector for agent-stack-k8s

nsuma8989:
This PR adds a log collector script for agent stack k8s which will help users of the stack to collect necessary logs for investigation incase of any issues


triarius (approved):
Thanks, @nsuma8989!

PR #205, opened 2023-10-08
Fix extraVolumeMounts integration test

triarius:
This was failing because the checkout was failing. It used git mirrors on a volume mount, and I suspect some checkout in a previous build corrupted the mirror directory. The corruption indicates a bug with how the agent uses git mirrors. As we have plans to fix that in the agent repo, I'm not going to investigate that here.

In any case, agent-stack-k8s can protect itself from such bugs in the agent by not using the same volume across builds. Instead, I've checked that the volume is shared between steps (and therefore pods) in the same build.

I've also used a `defer` in the integration tests to clean up the pipelines that are created. It seems to me that the tests attempted to delete the pipeline as soon as it was created with back off. Seems a bit racy to me, so I moved the cleanup into a `defer`.

Also, I've made the build message in an integration test have the name of the test. I considered making it the pipeline name too, but the pipeline name needs to be unique. It already has a random component, and the name is constrained, so I don't think putting the test name in there will work.

### DIFF
--- a/internal/integration/integration_test.go
+++ b/internal/integration/integration_test.go
@@ -20,7 +20,7 @@ func TestWalkingSkeleton(t *testing.T) {
 	}.Init()
 	ctx := context.Background()
 	pipelineID, cleanup := tc.CreatePipeline(ctx)
-	defer cleanup()
+	t.Cleanup(cleanup)
 	tc.StartController(ctx, cfg)
 	build := tc.TriggerBuild(ctx, pipelineID)
 	tc.AssertSuccess(ctx, build)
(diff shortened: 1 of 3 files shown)


internal/integration/integration_test.go
```diff
@@ -19,7 +19,8 @@ func TestWalkingSkeleton(t *testing.T) {
 		GraphQL: api.NewClient(cfg.BuildkiteToken),
 	}.Init()
 	ctx := context.Background()
-	pipelineID := tc.CreatePipeline(ctx)
+	pipelineID, cleanup := tc.CreatePipeline(ctx)
+	defer cleanup()
```
moskyb, line 23: these could potentially be swapped out for `t.Cleanup()` calls, but the actual distinction and when cleanup works better than defer escapes me... regardless, if this works fine then i probably wouldn't futz with it too much

moskyb (approved):

internal/integration/integration_test.go
triarius: Oh, TIL `t.Cleanup`. Apparently, `defer` can be buggy if you invoke `t.Run`, which a `t.Parallel` inside a test. The main test function can return (and run the defer) while the function is `t.Run` is still running 😱

We don't use `t.Run` in the integration tests, but I think we should still use `t.Cleanup` in case we do in the future and get puzzled by the flake.

triarius: Source: https://brandur.org/fragments/go-prefer-t-cleanup-with-parallel-subtests

DrJosh9000 (approved):

PR #207, opened 2023-12-03
Use volume mounts to cache go mod and build

triarius:
This takes the time to run some steps down from [3 min](https://buildkite.com/buildkite-kubernetes-stack/kubernetes-agent-stack/builds/873#018c31c2-b194-4a0e-b9fc-6f3e17a4ecd5) to [3 s](https://buildkite.com/buildkite-kubernetes-stack/kubernetes-agent-stack/builds/874#018c31e0-81ef-4b14-931e-882382ebb23c).

These will be shared between all pods on the same node, so cache misses will happen if the pod is scheduled on a different node. But the go caches are content addressable, so it should be safe to share them between multiple pods on the same node that are running concurrently.

### DIFF
--- a/.buildkite/pipeline.yaml
+++ b/.buildkite/pipeline.yaml
@@ -25,8 +25,27 @@ steps:
     - kubernetes:
         podSpec:
           containers:
-          - image: golang:1.20-alpine
+          - image: golang:1.21-alpine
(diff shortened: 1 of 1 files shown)


DrJosh9000 (approved):
Nice speedup!

PR #220, opened 2023-12-05
Use environment variables instead of arguments to control integration test main

triarius:
Sometimes, I want to run the integration tests in a way that skips cleanup up the pipelines that they create (to debug them). This was controlled by an environment variable, which cause the test running script to add an additional argument to the test command.

However, this agurment causes the non-integration tests to fail, meaning if the debug run had a successful run of the intetration tests, the build will still fail because the non-integration failed.

I could have separated the non-integration tests from the integration tests, but then take ~50ms to run, so the overhead of creating a separate step for them is far from worth it.


DrJosh9000 (approved):
Looks good, no notes

PR #243, opened 2024-01-25
PS-13 update readme to include default behaviour if no podspec is specified

SorchaAbel:
A support ticket requested clarification in our doco

Link to coda https://coda.io/d/_dHnUHNps1YO#Escalations-Table_tuS42/r224&view=modal

Link to code that I think does this
https://github.com/buildkite/agent-stack-k8s/blob/aa898e835b15116951a1cd84bcd3ca4fb239563f/internal/controller/scheduler/scheduler.go#L156-L164

It seems a little odd to be using k8s and not specifiy a podSpec but if I'm reading the code correctly then it will default to the step/job command


DrJosh9000 (approved):
Thanks! Looks right to me

PR #246, opened 2024-02-01
Replace command that copyied the .ssh directory and made the workspace world rwx

triarius:
Previously, a container in the job's pod would copy the .ssh directory and recursively chmod 777 the workspace directory, which contains the checkout directory.

This has directly or indirectly caused several issues:
- The checkout directory will also show a git diff: https://github.com/buildkite/agent-stack-k8s/blob/-/.buildkite/steps/check-code-generation.sh#L12
- The 777 permissions prevents private plugins from being checked out: https://3.basecamp.com/3453178/buckets/23112918/card_tables/cards/6211134454#__recording_6230141831
- Because the command was run through the agent, repository hooks were run in the checkout container, when (some of them) should be run in a job container: https://coda.io/d/Escalations-Feedback_dHnUHNps1YO/Pipelines-Escalations_su7FT#Pipelines-Escalations-Board_tu__K/r189

See also: https://linear.app/buildkite/issue/PDP-771/rethink-making-the-workspace-world-rwx where some of this is consolidated.

I believe copying the .ssh directory was done so that users using an ssh private key to checkout their repository will be able to use these credentials in their jobs as well. But I think the necessity to communicate with the repository after checkout is low, and if it's needed, there are alternative methods of mounting the credentials with the user should set up for their job containers. Furthermore, the fact that the permissions of the ssh private key was changed to be 777 makes it unlikely that ssh would have worked. It's [notoriously finicky](https://www.baeldung.com/linux/ssh-key-permissions#private-key-permissions) about the permissions of private keys.

So, in this PR, I have removed that command (the command phase entirely, which solves the issue of repository hooks running before the job). However, an issue arises if a job container does not run as root. To handle this, I've allowed the user to specify what user/group should own the workspace based on what they specific in the security context of the PodSpec. This does mean that all their containers should run as the same user, but I think that's an appropriate restriction. Using the security context to control this rather than creating another option seems more idiomatic to Kubernetes.

So now, the checkout container only performs the checkout phase, so hooks that are not part of the checkout phase will not run. Hooks that are part of the checkout phase will still run. We may decide to allow users to specify plugins to run in the checkout container too, but that's not done in this PR.

Along the way, I encountered several issues with the integration tests (a big theme in many of my PRs). I did break some of the fixes out into other PRs, but it is hard to that to some of the ones in this PR without failing tests.


internal/integration/testcase_test.go
```diff
@@ -89,9 +90,11 @@ func (t testcase) CreatePipeline(ctx context.Context) (string, func()) {
 	})
 	require.NoError(t, err)
 
-	return *pipeline.GraphQLID, func() {
-		if !preservePipelines {
-			t.deletePipeline(ctx)
+	return *pipeline.GraphQLID, func(t testcase) func() {
```
DrJosh9000, line 93: Why does the cleanup closure need to take the (same) `testcase` as an arg? afaict a copy of `testcase` will contain the same `*testing.T` and `*buildkite.Client`, so the cleanup closure should work on the captured `t` without having to pass it back in.

internal/controller/scheduler/scheduler.go
```diff
@@ -386,25 +396,53 @@ func (w *jobWrapper) Build() (*batchv1.Job, error) {
...
+				podUser,
+			)}
+		} else {
+			checkoutContainer.Args = []string{fmt.Sprintf(`set -exufo pipefail
+adduser -D -u %d -G root -h /workspace buildkite-agent
+su buildkite-agent -c "buildkite-agent-entrypoint bootstrap"`,
+				podUser,
+			)}
+		}
+	}
```
triarius, line 444:
```suggestion
	} else {
		checkoutContainer.Command = []string{"buildkite-agent-entrypoint"}
		checkoutContainer.Args = []string{"bootstrap"}
	}
```

internal/controller/scheduler/scheduler.go
```diff
@@ -386,25 +396,53 @@ func (w *jobWrapper) Build() (*batchv1.Job, error) {
 		WorkingDir:      "/workspace",
```
triarius, line 396: Remove setting args when initializing checkoutContainer above

DrJosh9000 (approved):
🥝 Nice work

internal/controller/scheduler/scheduler.go
```diff
@@ -413,7 +451,6 @@ func (w *jobWrapper) Build() (*batchv1.Job, error) {
 		Command:         []string{"cp"},
 		Args: []string{
 			"/usr/local/bin/buildkite-agent",
-			"/usr/local/bin/ssh-env-config.sh",
```
triarius, line 416: Undelete this

internal/controller/scheduler/scheduler.go
triarius: Not applicable now that's in a method.

internal/controller/scheduler/scheduler.go
```diff
@@ -378,33 +380,9 @@ func (w *jobWrapper) Build() (*batchv1.Job, error) {
...
-			Value: "checkout,command",
-		}, {
-			Name:  "BUILDKITE_AGENT_NAME",
-			Value: "buildkite",
-		}, {
-			Name:  "BUILDKITE_CONTAINER_ID",
-			Value: "0",
-		}, {
-			Name:  "BUILDKITE_COMMAND",
-			Value: "cp -r ~/.ssh /workspace/.ssh && chmod -R 777 /workspace",
```
triarius, line 403: Note: I've removed this command in the new definition of the checkout container below.

internal/controller/scheduler/scheduler.go
triarius: We're still 0.x, so I think it's ok to break for someone who relies on it. I think it's unlikely anyone would have relied on it as:
1. The role of this script is to write out the contents of an environment variable as a `.ssh/id_x` file (or equivalent).
2. That was already done in the checkout container, and the result was copied into the `/workspace/.ssh` directory, so there was no need to run this script in the previous versions.
3. Indeed, the script was not run in the job container as we defined them

If a customer does want to run this script in their job container, our recommendation will be to get if from the source: https://github.com/buildkite/docker-ssh-env-config

internal/controller/scheduler/scheduler.go
triarius: N/A because of the switch statment

internal/integration/fixtures/helloworld.yaml
triarius, line 1: This hello world was doing too much. I've broken that out into a different test.

internal/integration/testcase_test.go
triarius: I don't know what I was thinking when I wrote that. I was probably exhausted after many test failures that I could not debug!

I've simplified it significantly now.

internal/integration/testcase_test.go
```diff
@@ -188,10 +189,13 @@ func (t testcase) AssertArtifactsContain(build api.Build, expected ...string) {
 		nil,
 	)
 	require.NoError(t, err)
-	require.Len(t, artifacts, 2)
```
triarius, line 191: Seems a bit arbitrary to expect there to be exactly 2 artifacts 🤷‍♂️

Issue #222, opened 2023-12-10
Plugin assumes there is a working shell at /bin/sh and there often isn't one


c2h5oh:
This breaks one of the more common things that are going to be used with this plugin: Kaniko - only debug image has a shell, regular one doesn't.

This should at the very least be documented.

DrJosh9000:
Hi @c2h5oh, thanks for pointing that out. We'll look into documenting this requirement better.

Issue #221, opened 2023-12-10
Fetching SSH creds for git extemely unreliable in a seemingly random way


c2h5oh:
Changes to podSpec result in unpredictable checkout failures, example:

```yaml
steps:
  - label: ':pipeline: Pipeline Setup'
    agents:
      queue: my-ci
    plugins:
      - kubernetes:
          gitEnvFrom:
            - secretRef:
                name: buildkite-agent-ssh
          podSpec:
            containers:
              - image: 'buildkite/agent:latest'
                command:
                  - buildkite-agent
                args:
                  - pipeline upload
```
Works perfectly - 100% success rate

Adding an extra container results in 100% failure rate on checkout stage - missing creds - before any of the containers in spec are started. I have managed to trigger this by modifications as small as an additional space between args
```yaml
steps:
  - label: ':pipeline: Pipeline Setup'
    agents:
      queue: my-ci
    plugins:
      - kubernetes:
          gitEnvFrom:
            - secretRef:
                name: buildkite-agent-ssh
          podSpec:
            containers:
              - image: 'buildkite/agent:latest'
                command:
                  - buildkite-agent
                args:
                  - pipeline upload
              - image: 'buildkite/agent:latest'
                command:
                  - /bin/bash
                args:
                  - echo $${BUILDKITE_BRANCH}
```

Adding 
```yaml
env:
    GIT_SSH_COMMAND: ssh -vvv
```

yields 
```
[...]
debug1: Connection established.
debug1: identity file /root/.ssh/id_rsa type 0
debug1: identity file /root/.ssh/id_rsa-cert type -1
[...]
``` 
for the first example

```
[...]
debug1: Connection established.
debug1: identity file /root/.ssh/id_rsa type -1
debug1: identity file /root/.ssh/id_rsa-cert type -1
[...]
``` 
for the second one

With
```yaml
env:
    GIT_SSH_COMMAND: ssh -i /workspace/.ssh -vvv 
```
second example reports that `/workspace/.ssh` doesn't exist

With
```yaml
env:
    BUILDKITE_GIT_CLONE_FLAGS: "--depth=1"
```
That flag is passed to git command in first example, but not in the second.

triarius:
Hi @c2h5oh thanks for the detailed bug report.

I'm in the middle of overhauling how this works. I will try to make sure the new system works with the pipeline YAML you've provided. It would help immensely if you also send a link to the jobs to support@buildkite.com.

[...]

PR #249, opened 2024-02-11
Update README and help text to clarify a few things

triarius:
- clarifies that a single agent-stack-k8s controller can only listen to a single queue.
- more detail to the example pipeline, and switch the order so that the simpler example is first
- clarify how `gitEnvFrom` works
- clarify requirement for POSIX shell in image

Fixes: #222 #221


DrJosh9000 (approved):
-🌊

PR #251, opened 2024-02-11
Upload artifacts from artifact_paths as part of job execution

### LINKED ISSUE #203: [bug] artifacts dont upload on failed step
```yaml
steps:
  - label: build image
    artifact_paths:
      - "myfile"
    agents:
      queue: kubernetes
    plugins:
      - kubernetes:
          podSpec:
            containers:
              - image: alpine:latest
                command: [echo]
                args:
                - false
                - exit 1
```
On completion, the entire pod is torn down, rather than exiting its own container and continuing to go into the artifact container
That means that no artifacts are uploaded

triarius:
This removes the artifacts container and makes each job container upload artifacts.

Each job container runs `buildkite-agent bootstrap` with a command phase.
Searching artifact paths and uploading the results is [part of the definition](https://github.com/buildkite/agent/blob/c7693c49ad7ac52bfaf1270dab35dfc447f23228/internal/job/executor.go#L184-L225) of the command phase.

So this should fix #203. To verify this, I've added an integration test that an artifact is uploaded from a job container whose command fails.

Fixes: #203
Closes: #238

### DIFF
--- a/internal/controller/scheduler/completions.go
+++ b/internal/controller/scheduler/completions.go
@@ -61,7 +61,7 @@ func (w *completionsWatcher) OnUpdate(old interface{}, new interface{}) {
 func (w *completionsWatcher) cleanupSidecars(pod *v1.Pod) {
 	if terminated := getTermination(pod); terminated != nil {
 		if err := retry.RetryOnConflict(retry.DefaultRetry, func() error {
-			ctx := context.Background()
+			ctx := context.TODO()
 			job, err := w.k8s.BatchV1().Jobs(pod.Namespace).Get(ctx, pod.Labels["job-name"], metav1.GetOptions{})
 			if err != nil {
 				return err
--- a/internal/controller/scheduler/completions.go
+++ b/internal/controller/scheduler/completions.go
@@ -5,7 +5,7 @@ import (
 
 	"github.com/buildkite/agent-stack-k8s/v2/internal/controller/config"
 	"go.uber.org/zap"
-	v1 "k8s.io/api/core/v1"
+	"k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/informers"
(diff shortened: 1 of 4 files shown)


internal/controller/scheduler/completions.go
```diff
@@ -61,14 +61,14 @@ func (w *completionsWatcher) OnUpdate(old interface{}, new interface{}) {
...
 			if err != nil {
 				return err
-			} else {
-				job.Spec.ActiveDeadlineSeconds = pointer.Int64(1)
-				_, err = w.k8s.BatchV1().Jobs(pod.Namespace).Update(context.TODO(), job, metav1.UpdateOptions{})
-				return err
 			}
+			job.Spec.ActiveDeadlineSeconds = pointer.Int64(defaultTermGracePeriodSeconds)
+			_, err = w.k8s.BatchV1().Jobs(pod.Namespace).Update(ctx, job, metav1.UpdateOptions{})
+			return err
```
triarius, line 71: This was a bit hard to read, so I've refactored it.

internal/controller/scheduler/completions.go
```diff
@@ -61,14 +61,14 @@ func (w *completionsWatcher) OnUpdate(old interface{}, new interface{}) {
 func (w *completionsWatcher) cleanupSidecars(pod *v1.Pod) {
 	if terminated := getTermination(pod); terminated != nil {
 		if err := retry.RetryOnConflict(retry.DefaultRetry, func() error {
-			job, err := w.k8s.BatchV1().Jobs(pod.Namespace).Get(context.TODO(), pod.Labels["job-name"], metav1.GetOptions{})
+			ctx := context.Background()
```
DrJosh9000, line 64: Another thing for a "cleanup TODOs" list... until then I would leave it as `context.TODO()`

```suggestion
			ctx := context.TODO()
```

DrJosh9000 (approved):

PR #252, opened 2024-02-11
Name integration test pipeline after test name

triarius:
The pipeline that each integration test creates will now be named after the test, which makes it earlier to find when debugging. The format is something like: `<test name>-<job id>`, so tests with the same name from multiple jobs will be time ordered (since job id is time ordered).

I've also cleaned up the integration tests a bit, and fixed a few issues with them.

### DIFF
--- a/internal/integration/fixtures/extra-volume-mounts.yaml
+++ b/internal/integration/fixtures/extra-volume-mounts.yaml
@@ -10,29 +10,26 @@ steps:
       podSpec:
         containers:
         - image: alpine:latest
-          command: [ash, -c]
+          command: [touch]
           args:
-          - |-
-            echo volume_mounted > "/tmp/buildkite-git-mirrors/foo-$${BUILDKITE_JOB_ID}.txt"
-            ls -lA /tmp/buildkite-git-mirrors
-            cat "/tmp/buildkite-git-mirrors/foo-$${BUILDKITE_JOB_ID}.txt"
+          - /tmp/buildkite-git-mirrors/foo-$$(BUILDKITE_JOB_ID).txt
         - image: alpine:latest
           command: [ash, -c]
           args:
           - |-
             COUNT=0
             until [[ $$((COUNT++)) == 9 ]]; do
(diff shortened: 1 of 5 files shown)


internal/integration/fixtures/extra-volume-mounts.yaml
triarius, line 1: This test was failing because one job would write to a node-local volume and another pod would read from it. But there's no guarantee that the pods would be scheduled on the same node! 🤦‍♂️

Now, we do the reading and writing in different containers on the same pod. The `extraVolumeMounts` syntax will ensure they both have the node-local volume mounted at the same path.

internal/integration/testcase_test.go
```diff
@@ -39,19 +39,23 @@ type testcase struct {
...
-		nameVariable := fmt.Sprintf("%s-%d", strings.ToLower(t.Name()), time.Now().UnixNano())
-		hash := fmt.Sprintf("%x", sha256.Sum256([]byte(nameVariable)))
-
-		// k8s labels are limited to length 63, we use the pipeline name as a label.
-		// So we need to limit the length of the pipeline name too.
-		t.PipelineName = fmt.Sprintf("%s%s", namePrefix, hash[:63-len(namePrefix)-len(queuePrefix)])
+		namePrefix := t.Name()
+		jobID := os.Getenv("BUILDKITE_JOB_ID")
+		if jobID == "" {
+			jobID = strconv.FormatInt(time.Now().Unix(), 10)
```
DrJosh9000, line 56: Is that possibly going to cause a problem? Two tests with the same name in the same second with an empty `BUILDKITE_JOB_ID`?

triarius: Whoops, I should have made it `UnixNano()`. The only time this should be run without a job id is if you're running it locally, in which case you just fix it 🤷‍♂️

triarius: Its nanoseconds now.

DrJosh9000 (approved):

PR #258, opened 2024-02-23
Set BUILDKITE_HOOKS_PATH by default

bpoland:
I was having trouble getting hooks to run with k8s agents, then finally realized that although the agent containers have a `/buildkite/hooks` directory created in them, nothing was pointing the agent at that directory.  I was able to get the hooks working by simply setting this env var -- seems like this is something that should be done automatically but let me know if I'm off base!

### DIFF
--- a/internal/controller/scheduler/scheduler.go
+++ b/internal/controller/scheduler/scheduler.go
@@ -197,6 +197,9 @@ func (w *jobWrapper) Build() (*batchv1.Job, error) {
 		}, {
 			Name:  "BUILDKITE_BIN_PATH",
 			Value: "/workspace",
+		}, {	
+			Name:  "BUILDKITE_HOOKS_PATH",
+			Value: "/buildkite/hooks",
 		}, {
 			Name: agentTokenKey,
 			ValueFrom: &corev1.EnvVarSource{
--- a/README.md
+++ b/README.md
@@ -205,6 +205,45 @@ steps:
                   - --image=ttl.sh/example:1h
 ```
 
+## Agent Hooks
+The agent looks for hooks in the `/buildkite/hooks` directory by default. You can either build your own agent image with hooks baked there, or mount a volume there which includes the hooks you want to run. For example, if you have a `my-agent-hooks` configmap like this:
+```yaml
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: my-agent-hooks
+data:
+  pre-checkout: |
+    #!/usr/bin/env bash
+
+    echo "Pre-checkout hook"
+```
+
+Then you could mount it like this:
+```yaml
+steps:
+  - label: build image
+    agents:
+      queue: kubernetes
+    plugins:
+      - kubernetes:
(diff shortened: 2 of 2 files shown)


bpoland:
Also just a note for anyone that sees this, I was able to get it working in the mean time by adding `BUILDKITE_HOOKS_PATH: /buildkite/hooks` in the secret referenced by `gitEnvFrom` alongside the SSH key, since the entire secret gets mounted as env vars for the containers :)

https://github.com/buildkite/agent-stack-k8s/tree/main?tab=readme-ov-file#cloning-repos-via-ssh

internal/controller/scheduler/scheduler.go
```diff
@@ -197,6 +197,9 @@ func (w *jobWrapper) Build() (*batchv1.Job, error) {
 		}, {
 			Name:  "BUILDKITE_BIN_PATH",
 			Value: "/workspace",
+		}, {	
+			Name:  "BUILDKITE_HOOKS_PATH",
+			Value: "/buildkite/hooks",
```
triarius, line 202: I think the reason we didn't set this is that this directory won't necessarily exist. So setting it to some value won't work unless if users have some way to getting hooks into that directory.

If you're using a volume mount, then, you will need to do something at the step level to mount that volume into your container. (e.g. the example YAML in the file above). If you're going to the trouble of doing that, then you may as well specify `BUILDKITE_HOOKS` in that YAML. A user could mount their hooks volume anywhere, so it's not appropriate for the controller to inject a value for BUILDKITE_HOOKS that does not match the mount path.

So the example YAML should look something like:
```yaml
steps:
  - label: build image
    agents:
      queue: kubernetes
    plugins:
      - kubernetes:
          podSpec:
            containers:
              - image: gradle:latest
                command: [ gradle ]
                args:
                  - jib
                  - --image=ttl.sh/example:1h
                env:
                  - name: BUILDKITE_HOOKS
                    value: /buildkite/hooks  # <---- This value
            volumes:
              - name: agent-hooks
                configMap:
                  name: my-agent-hooks
                  defaultMode: 0755
          extraVolumeMounts:
            - name: agent-hooks
              mountPath: /buildkite/hooks   # <---- needs to be the same as this
```


If you're using a custom container image that contains these hooks, I think you should be able to put
```
ENV BUILDKITE_HOOKS /buildkite/hooks
```
in the Dockerfile. That might be more ergonomic.

triarius (commented):
Sorry it took a long time to review this, but I had some nagging feeling that something wasn't right at the back of my mind and I think I've realised what it is.

I've provided two alternative ways to get agent hooks to work. LMK what you think.

internal/controller/scheduler/scheduler.go
bpoland: Sorry I missed this comment previously! I agree that someone could put their hooks anywhere but it seems like having a convention would be nice, particularly with the README example for new users.  But these options definitely work so I'm fine to drop it if you disagree.

PR #263, opened 2024-02-28
Fix k8s-stack only getting jobs that have equal set of agent query rules

moskyb:
The k8s stack as it currently stands has a bug:

Let's say that we launch a k8s stack with the following tags:
```
  queue=cotopaxi
  colour=chartreuse
  lake=quilotoa
```

The way Buildkite normally works, and the way we would expect the k8s stack to work, is that the k8s stack would pick up any job that has agent query rules that cover a __subset__ of these tags. That is, we might expect that a jobs with agent query rules:
```
  queue=cotopaxi
  lake=quilotoa
```
or
```
  queue=cotopaxi
```
would be picked up by the k8s stack.

Unfortunately, there's a bug in the way we query for jobs where, essentially, agent query rules get ANDed together, and as a result, the k8s stack will only pick up jobs that have __the complete set__ of agent query rules matching the agent's tags. That is, the agent mentioned above will __only__ pick up jobs that have all of `queue=cotopaxi, colour=chartreuse, lake=quilotoa`. This is a bug.

To fix this, i've implemented the following fix:
- instead of explicitly querying jobs that match the tags for our agent, we'll __only__ query for the ones that match the queue for our instance
- from here we'll decide if we're a valid instance to execute that job - eg, if the job has an agent query rule that excludes us, we'll ignore it. Previously, we were relying on the backend to do this for us, but this is where the query is broken for our use case

### DIFF
--- a/internal/controller/monitor/monitor.go
+++ b/internal/controller/monitor/monitor.go
@@ -3,6 +3,7 @@ package monitor
 import (
 	"context"
 	"encoding/base64"
+	"errors"
 	"fmt"
 	"sort"
 	"time"
@@ -83,14 +84,14 @@ func (r clusteredJobResp) CommandJobs() []*api.JobJobTypeCommand {
 
 // getScheduledCommandJobs calls either the clustered or unclustered GraphQL API
 // methods, depending on if a cluster uuid was provided in the config
-func (m *Monitor) getScheduledCommandJobs(ctx context.Context, tags []string) (jobResp, error) {
+func (m *Monitor) getScheduledCommandJobs(ctx context.Context, queue string) (jobResp, error) {
 	if m.cfg.ClusterUUID == "" {
-		resp, err := api.GetScheduledJobs(ctx, m.gql, m.cfg.Org, tags)
+		resp, err := api.GetScheduledJobs(ctx, m.gql, m.cfg.Org, []string{fmt.Sprintf("queue=%s", queue)})
 		return unclusteredJobResp(*resp), err
 	}
 
 	resp, err := api.GetScheduledJobsClustered(
-		ctx, m.gql, m.cfg.Org, tags, encodeClusterGraphQLID(m.cfg.ClusterUUID),
+		ctx, m.gql, m.cfg.Org, []string{fmt.Sprintf("queue=%s", queue)}, encodeClusterGraphQLID(m.cfg.ClusterUUID),
 	)
 	return clusteredJobResp(*resp), err
 }
@@ -109,13 +110,20 @@ func (m *Monitor) Start(ctx context.Context, handler JobHandler) <-chan error {
 
 	agentTags := toMapAndLogErrors(logger, m.cfg.Tags)
 
+	var queue string
+	var ok bool
+	if queue, ok = agentTags["queue"]; !ok {
+		errs <- errors.New("missing required tag: queue")
+		return errs
+	}
+
 	go func() {
 		logger.Info("started")
 		ticker := time.NewTicker(time.Second)
 		defer ticker.Stop()
 
 		for {
-			resp, err := m.getScheduledCommandJobs(ctx, m.cfg.Tags)
+			resp, err := m.getScheduledCommandJobs(ctx, queue)
 			if err != nil {
 				logger.Warn("failed to get scheduled command jobs", zap.Error(err))
 				continue
--- a/internal/controller/scheduler/scheduler.go
+++ b/internal/controller/scheduler/scheduler.go
@@ -242,34 +242,47 @@ func (w *jobWrapper) Build() (*batchv1.Job, error) {
 		c.Args = []string{"bootstrap"}
 		c.ImagePullPolicy = corev1.PullAlways
 		c.Env = append(c.Env, env...)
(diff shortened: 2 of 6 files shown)


internal/controller/scheduler/scheduler.go
```diff
@@ -242,34 +242,47 @@ func (w *jobWrapper) Build() (*batchv1.Job, error) {
...
+				Value: strings.Join(clicommand.RedactedVars.Value.Value(), ","),
+			},
+			corev1.EnvVar{
+				Name:  "BUILDKITE_SHELL",
+				Value: "/bin/sh -ec",
+			},
+			corev1.EnvVar{
+				Name:  "BUILDKITE_ARTIFACT_PATHS",
+				Value: w.envMap["BUILDKITE_ARTIFACT_PATHS"],
+			})
```
triarius, line 285: 🙏💖

I'd always been too scared to make this change 😂

internal/controller/scheduler/scheduler.go
```diff
@@ -242,34 +242,47 @@ func (w *jobWrapper) Build() (*batchv1.Job, error) {
...
+				Value: strconv.Itoa(i + systemContainers),
+			},
+			corev1.EnvVar{
+				Name:  "BUILDKITE_PLUGINS_PATH",
+				Value: "/tmp",
+			},
+			corev1.EnvVar{
+				Name:  "BUILDKITE_SOCKETS_PATH",
+				Value: "/workspace/sockets",
+			},
```
triarius, line 273: Ah, so this is needed because of the job api deexperimentification yeah?

The underlying issue is that agent-stack-k8s picks up the latest released agent at build time. We should fix that, but not in this PR>

internal/integration/testcase_test.go
```diff
@@ -51,11 +51,17 @@ func (t testcase) Init() testcase {
...
+		buildNum := os.Getenv("BUILDKITE_BUILD_NUMBER")
+		if buildNum == "" {
+			buildNum = strconv.FormatInt(time.Now().UnixNano(), 10)
+		}
+
+		t.PipelineName = strings.ToLower(fmt.Sprintf("test-%s-%s", namePrefix, buildNum))
+
+		retryNum := os.Getenv("BUILDKITE_RETRY_COUNT")
+		if retryNum != "0" {
+			t.PipelineName = fmt.Sprintf("%s-retry-%s", t.PipelineName, retryNum)
```
triarius, line 63: Ok looks good. The reason I chose job id is that it's guaranteed to be unique. This is too, I guess. I usually just copy the job id from the URL anchor when I need. But this should be easier to search for on the page.

internal/integration/testcase_test.go
```diff
@@ -114,8 +120,18 @@ func (t testcase) StartController(ctx context.Context, cfg config.Config) {
 func (t testcase) TriggerBuild(ctx context.Context, pipelineID string) api.Build {
 	t.Helper()
 
+	authorEmail := os.Getenv("BUILDKITE_BUILD_CREATOR_EMAIL")
+	authorName := os.Getenv("BUILDKITE_BUILD_CREATOR")
+	if authorName == "" {
+		authorName = "Agent Stack K8s Integration Test"
+	}
```
triarius, line 127: Does this actually mean Josh isn't making all the test builds? 😂

triarius (approved):
Excellent work! Thanks for fixing this 💜

LMK if you need help debugging the tests.

internal/integration/testcase_test.go
moskyb: yeah, i also changed it so because the UUIDs are really long and non-human-readable, and get truncated on my builds page:
CleanShot 2024-02-28 at 16 08 12@2x

internal/controller/scheduler/scheduler.go
moskyb: yep, because of that. it was previously defaulting to a home directory (`/`) that it didn't have permissions to create subdirectories in

internal/controller/monitor/monitor.go
```diff
@@ -109,13 +110,20 @@ func (m *Monitor) Start(ctx context.Context, handler JobHandler) <-chan error {
 
 	agentTags := toMapAndLogErrors(logger, m.cfg.Tags)
 
+	var queue string
+	var ok bool
+	if queue, ok = agentTags["queue"]; !ok {
+		errs <- errors.New("missing required tag: queue")
+		return errs
+	}
```
patrobinson, line 118: This is different to non-K8s agents, which default to `queue=default` https://buildkite.com/docs/agent/v3/queues#the-default-queue

But that's probably for historical reasons and I reckon it's better to be explicit.

patrobinson (approved):
There's still some differences to how non-k8s agents behave, any job without a `queue` tag will never be assigned, instead of [being assigned to any agent queue](https://buildkite.com/docs/agent/v3/cli-start#agent-targeting)

But again I reckon that's legacy behaviour we may not want to maintain.

internal/controller/monitor/monitor.go
moskyb: yeah, if this launches without a queue tag, it defaults to `queue=kubernetes`, this is just an extra layer of defense

PR #274, opened 2024-03-14
Update examples to clarify role of `command` and `args`

triarius:
In k8s containers, `command` is the executable to run, and `args` are the arguments to pass to it. However, in `agent-stack-k8s`, these are substituted with `buildkite-agent` and various commands like `start` and `bootstrap`.

The `command` and `args` that the user specifies in their `podSpec` is joined with spaces and set as the `BUILDKITE_COMMAND` environment variable for the `bootstrap` command: https://github.com/buildkite/agent-stack-k8s/blob/aa3cddf123470b7788426e3c49a02f873d451319/internal/controller/scheduler/scheduler.go#L240

This can lead to confusing situations, so I've updated the docs and the test fixture examples to discourage specifying both. Users should stick to just specifying `command`, but feel free to write a shell script with multiple commands.


DrJosh9000 (approved):
LGTM

Issue #233, opened 2023-12-19
Underlying parse error does not print if using gitEnvFrom for Git credentials


mbarrien:
In testing agent-stack-k8s, some of my podSpecs I wrote ended up being incorrectly formatted. When this happens, agent-stack-k8s tries to replace the container with a build failure job which just echoes the underlying error to the console, as defined in https://github.com/buildkite/agent-stack-k8s/blob/main/internal/controller/scheduler/scheduler.go#L433-L447.

However, when using gitEnvFrom for SSH credentials to checkout git, the underlying error gets masked, and instead the console just shows failure to checkout the Git repo (failing after 3 retries), and the underlying error is never printed. This is because the checkout container (nor any other container) do not get the gitEnvFrom attached to it when there a build failure. Thus checkout fails and the container echoing the error never prints.

The only workaround to this is to run the job, then identify and examine the pod it creates while it is in flight, and examine the pod manifest looking for the error text in the BUILDKITE_COMMAND environment variable. This is obviously not optimal.

DrJosh9000:
Thanks @mbarrien - you found the code, so you understand what's going on here. The goal of `BuildFailureJob` was a "best effort last ditch" attempt to surface validation problems, but clearly has issues.

We have ideas for a more comprehensive solution, but it won't be quick. In the meantime we'll accept PRs that would skip the checkout so that `BuildFailureJob` actually works.

Issue #273, opened 2024-03-13
"failure" job uses private image causing imagePullBackoff


bpoland:
Hey, I'm running into a bug consistently but it's pretty weird :)

In my podspec I am using an `imagePullSecrets` section since I'm using a private image.  It looks something like this (I've removed some extra sections that don't seem relevant, but let me know if you can't reproduce and I can share more details):

```
  - label: Build
    plugins:
      - kubernetes:
          gitEnvFrom:
            - secretRef:
                name: buildkite-agent-ssh
          podSpec:
            imagePullSecrets:
              - name: image-pull-secret
            containers:
              - image: <my-private-image>
                command:
                  - |
                    echo "+++ build"
                    <my build command>
            volumes:
              - name: agent-hooks
                configMap:
                  name: buildkite-agent-hooks
                  defaultMode: 0755
          extraVolumeMounts:
            - name: agent-hooks
              mountPath: /buildkite/hooks
```

This worked fine and created a pod with the correct imagePullSecrets. But, we use a yaml linter that told me to replace the `defaultMode: 0755` with `defaultMode: "0755"` and when I did that, the pod created by agent-stack-k8s was now missing the `imagePullSecrets` and it failed to pull the image :(

I don't understand why adding quotes would have broken this but hopefully you can reproduce and fix it! :)

bpoland:
And just to be clear, adding the quotes causes the resulting pod to be completely missing the `imagePullSecrets` block...

triarius:
Thanks for raising this @bpoland,

I think the linter is giving you some suspicious advice. The `defaultMode` field of a config map projection in a volume is [supposed to have](https://kubernetes.io/docs/reference/kubernetes-api/config-and-storage-resources/volume/#projections) an `int32` type, so `0755` should be the value rather than `"0755"`, which is a `string` type. Maybe the linter is just hypercorrecting for the fact that `0755` is octal notation, i.e. 493 in decimal. But the k8s docs (and POSIX conventions) explicitly allow for file permission to be specified as octals, so I think it's preferable to leave it as `0755`. I think YAML 1.2 supports `0o755` as a more explicit octal notation, maybe your linter won't complain about it then? Hopefully the layers of parsing that the pipeline YAML goes through allows for that.

[...]

PR #275, opened 2024-03-14
Skip checkout when there are parse errors

triarius:
When the controller attempts to parse the Kubernetes plugin config and encounters an error, it will run a "failure job" that echos the error message as the command so that it is surfaced in the job logs. However, for private repositories that need secrets for checkout, this will fail because the container is not given access to those secrets.

Rather than give the container access to the secrets, in this PR, we skip the checkout step entirely when there are such errors: it is not needed and can only slow down the job or cause some other error.

The error message container is run with the default agent image, which is public, so it should not need any more resources like `imagePullSecrets` to run. I guess this means that it won't necessarily work on k8s clusters that are air-gapped, but this image is needed for other containers like the `agent start` container and the `init` container, so no other jobs would work on such clusters anyway.

In any case, I don't think transferring things like `gitEnvFrom` and `imagePullSecret` are going to work if there was an error in the specification of those fields. We want this failure job to be as bare bones as possible to reduce the chance of errors in the Pipeline YAML from preventing it from running.

Fixes: #233, #273

#### Note to reviews
I've implemented this as a `skipCheckout` flag on the `Build` method. I'm aware that such practices don't align with our coding standards. But, I think the entire `Build` method needs to refactored, and I don't want to do that yet. There are a few PRs in flight (#262, #248, #258) that modify the `Build` method, so a significant refactor would introduce too many merge conflicts.

After they are merged, we should refactor this method into smaller, composable pieces so that when there is a parse error, the checkout container is omitted and an error message container is substituted for the job command containers.

### DIFF
--- a/.buildkite/steps/tests.sh
+++ b/.buildkite/steps/tests.sh
@@ -13,6 +13,5 @@ export IMAGE
 
 gotestsum --junitfile "junit-${BUILDKITE_JOB_ID}.xml" -- \
   -count=1 \
-  -failfast \
   -ldflags="-X ${package}.branch=${branch}" \
   ./...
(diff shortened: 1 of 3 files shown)


.buildkite/steps/tests.sh
```diff
@@ -13,6 +13,5 @@ export IMAGE
 
 gotestsum --junitfile "junit-${BUILDKITE_JOB_ID}.xml" -- \
   -count=1 \
-  -failfast \
```
triarius, line 16: I think we want to see all the failures at once.

moskyb (approved):
🤙 send it

PR #279, opened 2024-03-23
feat: implement customizable ticker interval to mitigate buildkite api rate limit Issues

42atomys:
This pull request introduces the capability to customize the ticker interval within our system. Previously, a static ticker interval of 1 second significantly restricted operations, especially when running more than 10 controllers, by exceeding Buildkite’s GraphQL API complexity rate limits. This enhancement provides a workaround that allows users to adjust the ticker interval, thereby optimizing their usage to avoid hitting these limits.

### DIFF
--- a/cmd/controller/controller.go
+++ b/cmd/controller/controller.go
@@ -67,6 +67,11 @@ func AddConfigFlags(cmd *cobra.Command) {
 		10*time.Minute,
 		"time to retain kubernetes jobs after completion",
 	)
+	cmd.Flags().Duration(
+		"poll-interval",
+		time.Second,
+		"time to wait between polling for new jobs",
+	)
 	cmd.Flags().String(
 		"cluster-uuid",
 		"",
--- a/cmd/controller/controller.go
+++ b/cmd/controller/controller.go
@@ -70,7 +70,7 @@ func AddConfigFlags(cmd *cobra.Command) {
 	cmd.Flags().Duration(
 		"poll-interval",
 		time.Second,
-		"time to wait between polling for new jobs",
+		"time to wait between polling for new jobs (minimum 1s); note that increasing this causes jobs to be slower to start",
 	)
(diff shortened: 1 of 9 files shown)


internal/controller/monitor/monitor.go
```diff
@@ -37,6 +38,10 @@ type JobHandler interface {
 func New(logger *zap.Logger, k8s kubernetes.Interface, cfg Config) (*Monitor, error) {
 	graphqlClient := api.NewClient(cfg.Token)
 
+	if cfg.PollInterval == 0 {
```
DrJosh9000, line 41: 1 second should be a lower bound, so let's do this:

```suggestion
	if cfg.PollInterval < time.Second {
```

cmd/controller/controller.go
```diff
@@ -67,6 +67,11 @@ func AddConfigFlags(cmd *cobra.Command) {
 		10*time.Minute,
 		"time to retain kubernetes jobs after completion",
 	)
+	cmd.Flags().Duration(
+		"poll-interval",
+		time.Second,
+		"time to wait between polling for new jobs",
```
DrJosh9000, line 73: I think it's important to call out the effect on speed to pick up jobs:

```suggestion
		"time to wait between polling for new jobs (minimum 1s); note that increasing this causes jobs to be slower to start",
```

DrJosh9000 (approved):
Thanks for raising this, and sorry for the delay in getting around to it. Our main discussions have been around how this would negatively impact the speed at which jobs would be picked up, when a short-term fix is to increase the GraphQL quota, and longer-term improvements are being planned. But as we discussed yesterday, we'll accept this as a potentially useful configuration option for all customers.

I've rebased and pushed a handful of small tweaks. I'm planning on merging this now, but if you have any comments or things you would have done differently, please let us know!

Issue #239, opened 2024-01-10
[bug] post-checkout hooks are inoperable


agates4:
Because there isn't a checkout step (it's actually a command step under the hood), any configured post checkout hooks won't execute because there isn't an official checkout

moskyb:
hey @agates4! can you go into a bit more detail as to what this behaviour is preventing you from doing? are there specific bits that need to happen at post-checkout for you, or is there an existing plugin you're trying to use that uses this hook?

part of the k8s stack's plugin's role is to take control of the checkout and exec process, so it might not be feasible for post-checkout hooks to run, but we can take a look.

agates4:
> hey @agates4! can you go into a bit more detail as to what this behaviour is preventing you from doing? are there specific bits that need to happen at post-checkout for you, or is there an existing plugin you're trying to use that uses this hook?
> 
> part of the k8s stack's plugin's role is to take control of the checkout and exec process, so it might not be feasible for post-checkout hooks to run, but we can take a look.

Hey @moskyb I don't have a specific use case for this hook just wanted to point out what does not have parity with the longstanding Buildkite agent functionality!

triarius:
@agates4 There's another [issue](https://github.com/buildkite/agent-stack-k8s/issues/227) for a more customiseable checkout, and I think this could fall under that. Are you happy for use to consolidate these?

I think if we were going to start work in this area, one option I'm favouring is to allow the user to specify their own container definition to do the checkout, and bypass the default checkout container.

That container could just run a buildkite plugin and use the hooks in that, or it could do something else entirely, but we should establish a contract along the lines of it should check out the repo to `/workspace/$BUILDKITE_AGENT_NAME/$BUILDKITE_PIPELINE/` or something like that so that the job conatiners know where to find it.

bpoland:
Just circling back here, I can get global post-checkout hooks to work using the method I linked above but it seems like plugin post-checkout hooks aren't working for me.  Specifically, [this plugin](https://github.com/seek-oss/aws-sm-buildkite-plugin) -- the environment hook gets called but not the post-checkout one to write secrets to a file.

Looking closer at what's being executed and in what order, it doesn't appear to line up with [the hooks documentation](https://buildkite.com/docs/agent/v3/hooks#job-lifecycle-hooks).

<img width="466" alt="image" src="https://github.com/buildkite/agent-stack-k8s/assets/24273909/25daa3e3-ae76-4420-9212-743f52096f9e">

Specifically:
[...]

Issue #227, opened 2023-12-11
Make checkout configurable


c2h5oh:
- [ ] allow disabling checkout
- [ ] allow setting checkout-specific env vars (`BUILDKITE_GIT_CLONE_FLAGS`, `BUILDKITE_GIT_FETCH_FLAGS`) - you'd normally do it using a plugin but since this plugins also does checkout you can't

DrJosh9000:
Hi @c2h5oh, for disabling checkout, would setting `BUILDKITE_REPO=""` be possible for you?

c2h5oh:
Sure. Where would I be doing that? in step env?

DrJosh9000:
Currently, changing `BUILDKITE_REPO` must be done using either an environment or pre-checkout agent hook. The pipeline's configured repo would override attempts to set it with e.g. step env.

c2h5oh:
Wouldn't that disable checkout for all steps?

DrJosh9000:
I agree that this is probably not ideal, but since hooks are executed in each step, you could conceivably vary `BUILDKITE_REPO` based on some other variable set in the step env:

```shell
#!/usr/bin/env bash
if [[ "${SKIP_CHECKOUT:false}" == "true" ]]; then
    export BUILDKITE_REPO=""
fi
```

c2h5oh:
Well, since we agree this is not ideal would you accept a PR that would add checkout control to this plugin, or do you consider it out of scope?

DrJosh9000:
I was trying to figure out the best place to add these controls, since it might be a more general agent issue. We've discussed it and I think we agree there ought to be better control over checkout specifically for the k8s stack. We're happy for, say, a `checkout` section in the `kubernetes` plugin definition, but we don't have a firm idea of what that looks like yet. If you're willing to make a PR exploring this, please do!

c2h5oh:
I was thinking about combining features present in plugins normally used to implement this.

```yaml
checkout:
    disableCheckout: false
    cloneFlags: ""
    fetchFlags: ""
```

- setting `disableCheckout` to true would replace checkout hook with NOP, like this plugin does https://github.com/uber-workflow/run-without-clone-buildkite-plugin/blob/master/hooks/checkout
[...]

PR #281, opened 2024-03-27
Enable plugin phase in checkout container

DrJosh9000:
This implements the suggestion of @moskyb in #239

> wrt. plugin hooks not firing, i believe this can be fixed by changing [this line](https://github.com/buildkite/agent-stack-k8s/blob/main/internal/controller/scheduler/scheduler.go#L405) to Value: "plugin,checkout", - this will make the checkout container check out and execute plugins hooks, though only for the environment and (pre-, post- ) checkout hooks.

henceforth people won't have to fork the stack in order to enable plugins in the checkout container. This should address #227 as well - a plugin with a checkout hook should be able to customise the checkout.

Note there is still no way to customise the specifics (that should be coming with PodSpecPatch™ #282)

In case of other plugins, this PR removes the need to check out plugins in the command container, since plugins are enabled in the checkout container and saved in `/workspace/plugins` instead of `/tmp`. This should allow the use of command container images that don't contain `git` in that scenario.

The integration test uses improbable-eng/metahook (awesome!). Although only two hooks are set (`pre-checkout`, `post-checkout`), metahook hooks almost everything, and its hooks all start with `#!/usr/bin/env bash`. So the integration test needs a command container image with bash in it.

### DIFF
--- a/internal/controller/scheduler/scheduler.go
+++ b/internal/controller/scheduler/scheduler.go
@@ -283,6 +283,10 @@ func (w *jobWrapper) Build(skipCheckout bool) (*batchv1.Job, error) {
 			Value: w.envMap["BUILDKITE_ARTIFACT_PATHS"],
 		},
 		{
+			Name:  "BUILDKITE_PLUGINS_PATH",
+			Value: "/workspace/plugins",
+		},
+		{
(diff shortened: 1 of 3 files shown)


triarius (approved):
Looks great! Sorry for the merge conflicts

PR #282, opened 2024-03-27
Allow patching pod spec from controller configuration and steps

triarius:
This is our spin on @42atomys' excellent suggestion to allow applying a [strategic merge patch](https://kubernetes.io/docs/tasks/manage-kubernetes-objects/update-api-object-kubectl-patch/#use-strategic-merge-patch-to-update-a-deployment-using-the-retainkeys-strategy) of the PodSpec. It builds on their work in #262 but adds a few things we suggested in PR review.

Notably
1. Adds integration and unit tests
2. Added `pod-spec-patch` to the helm value file's JSON schema
3. Parsing of the user provided PodSpecPatch YAML as `corev1.PodSpec`
4. It will **NOT** be possible to set `Command` and `Args` on a container via a pod spec patch. This is because we've found that the interaction between a step's `command:` attribute and these containers fields is too complex to expose to users. Now, using pod spec patch, the steps's `command:` will be the only way to customise the process executed in a container. However, the existing `sidecarContainers` functionality is not affected by this, so for users that need additional containers for which they need to customise `Command` and `Args`, we recommend they use that instead.

Now instead of specifying a verbose PodSpec in every step, users can specify a common patch in their helm chart. The controller will take the `command` of a step and attempt to create a JK8s ob around it. Further customisation of the PodSpec is supported by applying the patches. Thus, most of the complexity of the Kubernetes YAML can be shifted to the Helm chart's values rather than the Buildkite YAML. Any further step level customization should be performed with the step level `podSpecPatch` rather than specifying a step level `podSpec`.

Fixes: #247
Closes: #262

Reviewers should note that a significant detour had to be made in order to reuse the `corev1.PodSpec` go struct in parsing the config level podSpecPatch. The essence of the issue that the library used to parse configuration, `viper`, [does not support](https://github.com/spf13/viper/issues/1050#issuecomment-747703285) the struct tag `json:",inline"` because it relies on [`mapstructure`](https://github.com/mitchellh/mapstructure) for which the equivalent struct tag is `mapstructure:",squash"`. I've made a fork of `mapstructure` that rectifies this, and used a `replace` directive in `go.mod` to force `viper` to use that the forked version of `mapstructure`. I plan on contributing the fork back upstream, so hopefully this is a temporary situation. There is a more detailed explanation in the commit message for https://github.com/buildkite/agent-stack-k8s/pull/282/commits/8f253f55b807acf309c112157a9e6df235c62552


cmd/controller/controller.go
```diff
@@ -73,32 +77,72 @@ func addFlags(cmd *cobra.Command) {
 	)
 }
 
-func ParseConfig(cmd *cobra.Command, args []string) (config.Config, error) {
```
triarius, line 76: I've refactored and split up this method to make it easier to test.

internal/controller/scheduler/scheduler.go
```diff
@@ -243,56 +251,68 @@ func (w *jobWrapper) Build(skipCheckout bool) (*batchv1.Job, error) {
...
-			},
-			corev1.EnvVar{
-				Name:  "BUILDKITE_SHELL",
-				Value: "/bin/sh -ec",
-			},
-			corev1.EnvVar{
-				Name:  "BUILDKITE_ARTIFACT_PATHS",
-				Value: w.envMap["BUILDKITE_ARTIFACT_PATHS"],
-			})
+		)
```
triarius, line 310: Most of these were constant in `i` so were moved into `containerEnv` outside the loop.

DrJosh9000 (approved):
LGTM

I have various minor quibbles, but this is much better shipped than not.

artem-zinnatullin:
I just wanted to say that this is a very good change, makes common infra configuration much easier 👍

PR #301, opened 2024-04-08
Fix resource.Quantity parsing

### LINKED ISSUE #293: Default resources on pod-spec-patch throws validation error
I tried to implement the exact yaml example supplied in:
https://github.com/buildkite/agent-stack-k8s?tab=readme-ov-file#default-resources

The `helm install` operation works, but all of the containers fail with:
```
- config.pod-spec-patch.containers.0.resources.limits.memory: Invalid type. Expected: map, given: string
- config.pod-spec-patch.containers.0.resources.requests.cpu: Invalid type. Expected: map, given: string
- config.pod-spec-patch.containers.0.resources.requests.memory: Invalid type. Expected: map, given: string
```

Here is the exact `values.yaml` file I'm supplying
```
config:
    org: ...
    debug: true
    cluster-uuid: ...
    config:
      pod-spec-patch:
        containers:
          - name: agent          # this container acquires the job
            resources:
              requests:
                cpu: 100m
                memory: 50Mi
              limits:
                memory: 1Gi
          - name: checkout       # this container clones the repo
            envFrom:
              - secretRef:
                  name: secret-key
            resources:
              requests:
                cpu: 100m
                memory: 50Mi
              limits:
                memory: 1Gi
          - name: container-0    # the job runs in a container with this name by default
            resources:
              requests:
                cpu: 100m
                memory: 50Mi
              limits:
                memory: 1Gi
```

DrJosh9000:
Fixes #293

Fortunately this error was easy enough to replicate without Helm or having to set up a local registry: `just run --org (org) --buildkite-token (token) --debug -f config.yaml` where config.yaml was:

```yaml
pod-spec-patch:
  containers:
    - name: container-0
      resources:
        requests:
          cpu: 100m
```

For parsing our config once the controller starts, Viper gets the first bite (pun intended), and it uses mapstructure. Immediately before the failing call (`v.UnmarshalExact(...)`), Viper contains something like (obtained with `v.Debug()`):

```
Config:
map[string]interface {}{"pod-spec-patch":map[string]interface {}{"containers":[]interface {}{map[string]interface {}{"name":"container-0", "resources":map[string]interface {}{"requests":map[string]interface {}{"cpu":"100m"}}}}}}
```

Viper doesn't know how to put `"100m"` into a `resource.Quantity`. `resource.Quantity` knows how to unmarshal itself from JSON, but neither mapstructure nor Viper call that. Instead we must alter the decode hook used by Viper from its default (that conveniently understands `time.Duration` strings and comma-separated lists).

### DIFF
--- a/examples/config.yaml
+++ b/examples/config.yaml
@@ -27,3 +27,9 @@ pod-spec-patch:
         secretKeyRef:
           name: github-secrets
           key: github-token
+    resources:
+      requests:
+        cpu: 100m
+        mem: 50Mi
+      limits:
+        mem: 1Gi
--- a/cmd/controller/controller.go
+++ b/cmd/controller/controller.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"log"
 	"os"
+	"reflect"
 	"strings"
 	"time"
 
@@ -22,6 +23,7 @@ import (
 	"github.com/spf13/pflag"
 	"github.com/spf13/viper"
 	"go.uber.org/zap"
+	"k8s.io/apimachinery/pkg/api/resource"
 	"k8s.io/client-go/kubernetes"
 	restconfig "sigs.k8s.io/controller-runtime/pkg/client/config"
 	"sigs.k8s.io/controller-runtime/pkg/manager/signals"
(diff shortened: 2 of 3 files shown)


examples/config.yaml
```diff
@@ -27,3 +27,9 @@ pod-spec-patch:
         secretKeyRef:
           name: github-secrets
           key: github-token
+    resources:
+      requests:
+        cpu: 100m
+        mem: 50Mi
+      limits:
+        mem: 1Gi
```
tessereth, line 35: Did you want a test that includes something comma separated just to cover that case? Or does it not really matter that much?

tessereth (approved):
Nice catch! LGTM.

PR #303, opened 2024-04-09
Add checkout config block to plugin config

### LINKED ISSUE #227: Make checkout configurable

DrJosh9000:
Fixes #227 

The `kubernetes` plugin will now accept

```yaml
- kubernetes:
    checkout:
      skip: true
```

and this will completely skip scheduling a checkout container. 

git clone and fetch flags _could_ be configured with `podSpecPatch`. However the configuration logically belongs under `checkout`, so why not.


tessereth (approved):
I wonder if we want to make this config a first-class thing rather than a k8s-specific thing. But it does the job and it's useful so :shipit:

PR #309, opened 2024-04-18
[WIP] Add multi_command option

DrJosh9000:
The `kubernetes` plugin uses the Kubernetes definition of `podSpec`. [Kubernetes defines](https://kubernetes.io/docs/tasks/inject-data-application/define-command-argument-container/#running-a-command-in-a-shell) the `command` and `args` of a `container` to mean a single command entrypoint for the container (`command` is called `entrypoint` in some other container runtimes). To be as compatible as possible with Kubernetes and existing users, the default interpretation of `command` and `args` should remain consistent whether or not it is being passed through `BUILDKITE_COMMAND`.

For example:

```yaml
plugins:
- kubernetes:
    podSpec:
      containers:
        - name: echo
          command: ["my_command", "subcommand"]
          args: ["hello", "123", "foo"]
```

results in `$BUILDKITE_COMMAND` equal to `my_command subcommand hello 123 foo`.

Some people would like `command` to be interpreted as it would for a (non-Kubernetes) command step: it can be one or more commands (as a list). These are joined with `\n` before being passed through `BUILDKITE_COMMAND`.

If you have only one container in the `podSpec`, or multiple containers that run the same command, then there's no need for this new option: just use the `command` at the `step` level. It will be interpreted in the Buildkite way (multiple commands).

However, if `podSpec` defines multiple containers, _and_ they need to run different commands, _and_ a single command entrypoint is undesirable, _and_ they don't want to [explicitly call a shell to process multiple commands](https://kubernetes.io/docs/tasks/inject-data-application/define-command-argument-container/#run-a-command-in-a-shell)...

...then this new `multi_command` option can be used to interpret the `command` list as multiple commands.

For example:

```yaml
plugins:
- kubernetes:
    multi_command: true
    podSpec:
      containers:
        - name: echo
          command: ["my first command", "my other command"]
          args: ["hello", "123", "foo"]
```

results in `$BUILDKITE_COMMAND` equal to 
```
my first command
my other command hello 123 foo
```

### DIFF
--- a/internal/controller/scheduler/scheduler.go
+++ b/internal/controller/scheduler/scheduler.go
@@ -56,6 +56,7 @@ type KubernetesPlugin struct {
 	Metadata          Metadata               `json:"metadata,omitempty"`
 	ExtraVolumeMounts []corev1.VolumeMount   `json:"extraVolumeMounts,omitempty"`
 	Checkout          Checkout               `json:"checkout,omitempty"`
+	MultiCommand      bool                   `json:"multi_command,omitempty"`
 }
 
 type Checkout struct {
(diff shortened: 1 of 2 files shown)


DrJosh9000:
Closing not because it's a bad idea, but because it ought to be done from scratch (merge conflicts, probably worth re-thinking the specifics, etc).

PR #331, opened 2024-05-22
Make imagePullBackOffWatcher grace period configurable

DrJosh9000:
30 seconds may be too short for some customers. It should at least be configurable.

Also, as pointed out, pod creation time isn't a good moment to base the grace period - there could be a long period between creating the pod and the kubelet actually starting it. Pod start time is more useful.

### DIFF
--- a/cmd/controller/controller.go
+++ b/cmd/controller/controller.go
@@ -78,9 +78,9 @@ func AddConfigFlags(cmd *cobra.Command) {
 		"Bind address to expose the pprof profiler (e.g. localhost:6060)",
 	)
 	cmd.Flags().Duration(
-		"startup-grace-period",
-		config.DefaultStartupGracePeriod,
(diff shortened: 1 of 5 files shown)


cmd/controller/controller.go
```diff
@@ -77,6 +77,11 @@ func AddConfigFlags(cmd *cobra.Command) {
 		"",
 		"Bind address to expose the pprof profiler (e.g. localhost:6060)",
 	)
+	cmd.Flags().Duration(
+		"startup-grace-period",
```
zhming0, line 81: Is it better to use a name more specific to image pull back off watcher? The current name might confuse reader with the start up graceful period of the controller itself (a.k.a the `initialDelaySeconds`). 

```suggestion
		"image-pull-backoff-watcher-startup-grace-period",
```

DrJosh9000: Good point

zhming0 (approved):

PR #332, opened 2024-05-22
Some scheduler cleanup

DrJosh9000:
The main improvement is the removal of `err` as a _field_ from `jobWrapper`. The second is clearer logic for the error path (fallback to failure job).


zhming0 (approved):
✅ much easier to read now. 

I was confused with the term "job registration" before.

PR #336, opened 2024-05-31
PENT-103-part-1: refactor integration test so it works on clusterd org

zhming0:
This PR refactors integration test startup and teardown so it supports clustered organizations in Buildkite, which is the majority. 

Concretely, now each integration test will spin up an ephemeral agent queue + pipeline. 

Combined with the improvement in `development.md`, hopefully this PR also unblocks members of public to execute our integration tests in their own k8s cluster. 
Previously, nobody can run the integration unless they owns an unclustered Buildkite organization. 

---

Beside the main theme, this PR also slightly increase the ergonomic of test setup by moving the cleanup invocation into the test helpers.

---

If this PR is merged, the next step is to migrate our CI org to clusterd org and potentially marking cluster-id as mandatory.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,6 @@ Brewfile.lock.json
 .vscode
 
 dist/
+
+# For all glorious direnv users.
+.envrc
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -16,12 +16,72 @@ just --list
 
 # Integration Tests
 
+## Architecture
+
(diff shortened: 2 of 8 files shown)


.gitignore
```diff
@@ -4,3 +4,6 @@ Brewfile.lock.json
 .vscode
 
 dist/
+
+# For all glorious direnv users.
+.envrc
```
DrJosh9000, line 9: 👍

DrJosh9000 (approved):
This is a solid improvement! Good job!

PR #341, opened 2024-06-04
Configurable checkout/command/sidecar defaults

DrJosh9000:
Particularly, the ability to set a global `envFrom` that applies to each of the three flavours of container that can be run, and also the same thing but within the plugin config.


wolfeidau (approved):
Looks good 👍🏻

PR #342, opened 2024-06-04
Add CODEOWNERS

zhming0:
self-explanatory. 

So we can get automatically assigned when there are pull requests. 

The Github complains that the codeowner file has error (team doesn't exist?), I don't know if it's working... so let's give it shot.


DrJosh9000:
Good idea, is the team correct?

DrJosh9000 (approved):
YOLO

PR #344, opened 2024-06-07
PS-67: stop image pull backoff error handling for sidecars

zhming0:
Currently, our backoff error watcher monitors all containers in a pod, which is problematic for customers who heavily rely on sidecars.

Since sidecar errors theoretically do not impact the health of pipeline jobs, canceling an entire job based on the status of a sidecar only adds unnecessary trouble.

At the moment, we don't provide governance support for sidecars, meaning customers can't see logs from sidecars. When we kill a job due to a sidecar problem, customers aren't given a proper reason. Sometimes, their CI workload is functioning correctly, but some sidecars have a delayed start, leading to the job being killed. From the customers' perspective, everything appears to be working fine until something randomly terminates the job, which is frustrating.

Customers can debug sidecar issues themselves through their Kubernetes platform.

This PR reduces the scope of the image pull backoff error watcher so it only monitors containers that we actively govern.

NOTE: A longer-term solution is being planned to address the observability issues comprehensively, so this PR is a temporary solution.

### DIFF
--- a/internal/controller/scheduler/imagePullBackOffWatcher.go
+++ b/internal/controller/scheduler/imagePullBackOffWatcher.go
@@ -2,6 +2,8 @@ package scheduler
 
 import (
 	"context"
+	"regexp"
+	"slices"
 	"time"
 
 	"github.com/Khan/genqlient/graphql"
@@ -111,6 +113,10 @@ func (w *imagePullBackOffWatcher) cancelImagePullBackOff(ctx context.Context, po
 		if !shouldCancel(&containerStatus) {
 			continue
 		}
+		if !isSystemContainer(&containerStatus) {
+			log.Info("Ignoring sidecar container during ImagePullBackOff watch.", zap.String("name", containerStatus.Name))
+			continue
+		}
 
 		log.Info("Job has a container in ImagePullBackOff. Cancelling.")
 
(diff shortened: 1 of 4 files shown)


internal/controller/scheduler/imagePullBackOffWatcher.go
```diff
@@ -147,3 +152,12 @@ func shouldCancel(containerStatus *v1.ContainerStatus) bool {
 	return containerStatus.State.Waiting != nil &&
 		containerStatus.State.Waiting.Reason == "ImagePullBackOff"
 }
+
+// We don't want to kill jobs because of sidecars error.
+// The name "system container" is subject to more debate.
+func isSystemContainer(containerStatus *v1.ContainerStatus) bool {
+	name := containerStatus.Name
+	// This will arguably cause false positive, but we will replace this soon.
+	matched, _ := regexp.MatchString(`container-\d+`, name)
```
zhming0, line 161: It's probably wise to capture error from 
`agent`, `copy-agent` and `checkout` too. But I want to double check if there are already pre-existing error detection logic for these error path.

internal/controller/scheduler/imagePullBackOffWatcher.go
```diff
@@ -147,3 +152,25 @@ func shouldCancel(containerStatus *v1.ContainerStatus) bool {
...
+// to prevent customer confusion.
+//
+// Most importantly, the CI can still pass (in theory) even if sidecars fail.
+//
+// (The name "system container" is subject to more debate.)
+func isSystemContainer(containerStatus *v1.ContainerStatus) bool {
+	name := containerStatus.Name
+	if name == AgentContainerName || name == CopyAgentContainerName {
+		return true
+	}
```
moskyb, line 170: should we include the checkout container here as well? we'll probably want to alert if we can't pull that image... something like:

```golang
	if slices.Contains([]slice{AgentContainerName, CopyAgentContainerName, CheckoutContainerName}, name) {
		return true
	}
```

moskyb (changes requested):
one blocking question, but other than that LGTM!

internal/controller/scheduler/imagePullBackOffWatcher.go
zhming0: Good spot! Please take another look 🙏🏿 .

moskyb (approved):

PR #347, opened 2024-06-17
Move integration test on CI to use a different namespace

zhming0:
Our integration test runs our k8s controller outsides of our k8s cluster. It will spin up lots of jobs/pods in a k8s namespace X. 

Right now, X = `buildkite`, it happens to be where our CI is running. It becomes a problem now if we saw a pod in `buildkite` namespace, we don't know if the pod is from CI workload or integration tests. 

This change makes sure when we run our CI, workload generated by integration will run on its own namespace, reducing confusion and hopefully easier to manage on the long run.

### DIFF
--- a/.buildkite/steps/tests.sh
+++ b/.buildkite/steps/tests.sh
@@ -10,6 +10,11 @@ package="github.com/buildkite/agent-stack-k8s/v2/internal/integration_test"
 branch="${BUILDKITE_BRANCH:-main}"
 IMAGE=$(buildkite-agent meta-data get agent-image)
 export IMAGE
+# Do NOT allow integration test workloads in the buildkite namespace.
+# It causes confusion and complicates cleanup.
+# The buildkite namespace is reserved for CI workloads.
+# Integration tests will use their own namespace and out-of-cluster k8s controller to manage job pods.
+export NAMESPACE=buildkite-k8s-integration-test
 
 gotestsum --junitfile "junit-${BUILDKITE_JOB_ID}.xml" -- \
   -count=1 \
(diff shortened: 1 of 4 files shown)


DrJosh9000 (commented):
Looks good, one minor comment

.buildkite/steps/tests.sh
```diff
@@ -10,6 +10,12 @@ package="github.com/buildkite/agent-stack-k8s/v2/internal/integration_test"
 branch="${BUILDKITE_BRANCH:-main}"
 IMAGE=$(buildkite-agent meta-data get agent-image)
 export IMAGE
+# Do NOT allow integration test workloads in the buildkite namespace.
+# It causes confusion and complicates cleanup.
+# The buildkite namespace is reserved for CI workloads.
+# Integration tests will use their own namespace and out-of-cluster k8s controller to manage job pods.
+NAMESPACE=buildkite-k8s-integration-test
+export NAMESPACE
```
DrJosh9000, line 18: Not sure why we're not doing this:

```suggestion
export NAMESPACE=buildkite-k8s-integration-test
```

and above:

```diff
-IMAGE=$(buildkite-agent meta-data get agent-image)
-export IMAGE
export IMAGE="$(buildkite-agent meta-data get agent-image)"
```

zhming0: It's a good question. This is one of shellcheck's recommendations.

Please see https://www.shellcheck.net/wiki/SC2155

DrJosh9000: Oh, that makes sense.

DrJosh9000 (approved):

PR #351, opened 2024-06-18
Stop using old k8s exec envvar

zhming0:
This change will prevent users from being spammed with warning messages.

## Reviewer please read: 

The risk of this change is that users won't be able be downgrade agent to anything below `v3.74.0`. I don't expect users would be manually pinging agent version but if it's a risk we can do something smarter like: "if agent version > X then don't use the old experimental flag".


moskyb (approved):
LGTM!

re: not being able to downgrade, i'm totally fine with that, i think it's reasonable to assume that the k8s stack relies on agent features

PR #355, opened 2024-07-25
Refer to the correct section of the UI when finding the UUID for a queue

areitz:
Here is what I see in the UI when I attempt to follow these directions:

Screenshot 2024-07-25 at 3 37 22 PM

It seems like maybe the UI has changed since this document was written? I'm new to Buildkite, so it's possible that my UI is different than the norm.


wolfeidau (approved):
👍🏻 good suggestion.

PR #356, opened 2024-07-29
allow picking work off default cluster queue

wolfeidau:
Allow specifying empty query rules.

To use the default queue we can pass `--tags 'queue='` which configures the agent to omit the queue parameter from graphql calls, which in turn retrieves jobs from the default queue.

**NOTE:** The integration tests allow parallel testing within a given cluster, however the default queue is global to a cluster so some changes will be required to integration test this without breaking concurrent builds.


DrJosh9000 (approved):
Good stuff

PR #360, opened 2024-08-01
Replace fetch flags env with checkout param

DrJosh9000:
As seen often in the build logs, `BUILDKITE_GIT_FETCH_FLAGS` is not actually permitted. But the option is available through `kubernetes: checkout: fetchFlags:`.

Also, my editor decided to auto-format the YAML.


wolfeidau (approved):
👍🏻 🔨

PR #361, opened 2024-08-02
Adjust defaults for container0 (build container) as it is too frugal

wolfeidau:
As discovered while debugging some build issues, we could probably be a bit more generous with CPU and memory. Locally release builds are taking ~20 seconds, in this container they are taking over ~5 minutes threshold for jobs.. 

Currently the build container only has 1/10th of a CPU and 50mb of RAM.

### DIFF
--- a/cmd/controller/controller_test.go
+++ b/cmd/controller/controller_test.go
@@ -82,11 +82,8 @@ func TestReadAndParseConfig(t *testing.T) {
 					},
 					Resources: corev1.ResourceRequirements{
 						Requests: corev1.ResourceList{
-							"cpu": resource.MustParse("100m"),
-							"mem": resource.MustParse("50Mi"),
-						},
-						Limits: corev1.ResourceList{
-							"mem": resource.MustParse("1Gi"),
+							"cpu": resource.MustParse("1000m"),
+							"mem": resource.MustParse("4Gi"),
 						},
 					},
 				},
--- a/examples/config.yaml
+++ b/examples/config.yaml
@@ -57,7 +57,5 @@ pod-spec-patch:
               key: github-token
       resources:
         requests:
-          cpu: 100m
-          mem: 50Mi
-        limits:
-          mem: 1Gi
+          cpu: 1000m # one core
+          mem: 4Gi


cmd/controller/controller_test.go
```diff
@@ -82,11 +82,8 @@ func TestReadAndParseConfig(t *testing.T) {
 					},
 					Resources: corev1.ResourceRequirements{
 						Requests: corev1.ResourceList{
-							"cpu": resource.MustParse("100m"),
-							"mem": resource.MustParse("50Mi"),
-						},
-						Limits: corev1.ResourceList{
-							"mem": resource.MustParse("1Gi"),
+							"cpu": resource.MustParse("1000m"),
+							"mem": resource.MustParse("4Gi"),
```
zhming0, line 86: During CI, we run our test controller in the same k8s cluster with our production k8s controller. So setting a limit will be wise. 

Setting namespace resource limit is another option, I don't I remember I set it when I create the namespace.

wolfeidau: I am only modifying the podspec for container 0, sadly this isn't visible in the test diff or YAML diff... 

```
# This will be applied to the job's podSpec as a strategic merge patch
# See https://kubernetes.io/docs/tasks/manage-kubernetes-objects/update-api-object-kubectl-patch
pod-spec-patch:
  serviceAccountName: buildkite-agent-sa
  automountServiceAccountToken: true
  containers:
    - name: container-0
      env:
        - name: GITHUB_TOKEN
          valueFrom:
            secretKeyRef:
              name: github-secrets
              key: github-token
      resources:
        requests:
          cpu: 1000m # one core
          mem: 4Gi
```

zhming0: yeah I see it's container-0 being changed. I may have confused you in my comment I think (sorry). What I mean is that it might be safer to do this on container-0 spec: 

```
resources: 
  requests: 
    cpu: 1000m # one core 
    mem: 4Gi
  limites: 
    mem: 4Gi # This will prevent one bad test case impacting other tenants on the cluster
```

DrJosh9000 (approved):

cmd/controller/controller_test.go
wolfeidau: @zhming0 yeah I am definitely going to dig into how this works 🙏🏻 

So very good raising these sorts of questions!

PR #362, opened 2024-08-05
Added some notes about local testing

wolfeidau:
Just capturing some of the notes I have locally for developing code in this project.

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -74,10 +74,16 @@ just run --org $ORG --buildkite-token $BUILDKITE_TOKEN --debug --tags 'queue=,os
 
 # Testing Locally
 
-Before you start, check which kubernetes cluster your using by default.
+Before you start, check which kubernetes cluster configuration your using by default.
 
 ```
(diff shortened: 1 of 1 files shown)


DEVELOPMENT.md
```diff
@@ -52,18 +52,46 @@ Concretely, to get the integration test running locally, you will need:
...
+```
+just run --org $ORG --buildkite-token $BUILDKITE_TOKEN --debug --tags 'queue=,os=linux'
+```
+
+# Testing Locally
+
+Before you start, check which kubernetes cluster your using by default.
+
+```
+kubectl cluster-info
```
DrJosh9000, line 80: I recommend checking `kubectl config current-context` first. Orbstack magically created an `orbstack` context for me, but (as an example) right now I'm in a different context where `kubectl cluster-info` doesn't work.

wolfeidau: @DrJosh9000 good call, adjusted commands.

DrJosh9000 (approved):

PR #363, opened 2024-08-06
Fail jobs without starting pods

### LINKED ISSUE #273: "failure" job uses private image causing imagePullBackoff

DrJosh9000:
So a job is nonsense, a pod can't be constructed, or Kubernetes won't accept it. Currently, we build a whole 'nother pod containing an agent, a command container, an init container, etc  using the default agent image, and tell it to...

```shell
echo "the error message" && exit 1
```

which smacks of cracking a walnut with the Mesta 50000.

This PR changes it so that the scheduler uses the new agent functionality introduced in buildkite/agent#2915 - `core.Controller` and `core.JobController` - to acquire and fail the job directly within the agent-stack-k8s controller.

Fixes #273


wolfeidau (approved):
👍🏻 🚀

PR #371, opened 2024-08-26
Add choice of "command interposer"

DrJosh9000:
### What
Change how `podSpec/command` and `podSpec/args` is interpreted. The choice of "interposer" is up to the user. If they allow the Kubernetes plugin, the choice can be made per-step, otherwise there is a default that can be set in `values.yaml`. 

The default (if no choice is made) is changing.

`buildkite` mode is more convenient if you are used to pipeline.yaml syntax, and want to pass multiple commands as a list. `vector` mode is much more like how Kubernetes treats `command` and `args`, and is intended to execute a single command (that, per the Kubernetes docs, can be `["sh", "-c", "big script here"]`. There's also `legacy` mode, equivalent to what it does before this PR, to ease transitioning pipelines over to new modes. 

### Why
Joining the container command and args with `strings.Join(..., " ")` was never exactly a good idea. 

For example, before this PR, this doesn't do what it looks like it does:

```yaml
podSpec:
  containers:
    - command:
        - echo First command
        - echo Second command
```

Because each command component was joined with spaces, this would run the command `echo First command echo Second command`, and echo `First command echo Second command`, not two things over two lines. 

If we tried harder to interpret it the way Kubernetes does (as a raw arg vector executed directly, not a string interpreted by a shell), then we should be using the vector `["echo First command", "echo Second command"]`, which would fail because while `echo` could be an executable file (that takes one or more arguments), it's unlikely that `echo First command` is the name of an executable.

With no apologies to Terry Pratchet: If we're breaking Kubernetes' rules, we ought to break 'em good an' 'ard.


wolfeidau:
@DrJosh9000 not a fan of the readme changes, was there a reason for this?

Happy to defer to others but without some sort of automated formatter this feels like it is just adding friction.

DrJosh9000:
@wolfeidau it's not important here, so I'll back out the 100 column change.

wolfeidau (approved):
Looks good 👍🏻 🚀

PR #375, opened 2024-09-03
Improve the development docs

CerealBoy:
- Changing the kubernetes secret name of the SSH key used for the integration test.
- Added a section about debugging the `BUILDKITE_AGENT_TOKEN` if `BUILDKITE_TOKEN` works.
- Added a small usage of `--cluster-uuid ${CLUSTER_UUID}` for a specific to make it obvious the value is required.
- Added a bit about running a single integration test from local.
- Removed the cluster queues reference from the `cleanup-orphans` command as the command does not in fact clean up cluster queues.


wolfeidau (approved):
Nice work!

PR #376, opened 2024-09-05
Reduce MaxInFlight integration test flakes

DrJosh9000:
The MaxInFlightUnlimited integration test is flaky. This PR doesn't make it completely unflaky (I have ideas but it adds significant complication), but should reduce the chances of flakiness. While here, I tidied up some nits I had with the style, and made MaxInFlightLimited consistent with those changes.


CerealBoy (approved):
LGTM

PR #383, opened 2024-09-27
Log result of patched podSpec

artem-zinnatullin:
I'm debugging a bug where after podSpec patching some configuration options are lost, this should be helpful for others too.


DrJosh9000 (approved):
✅ 
Noting the change from `Info` log to `Debug` - some of us chatted and it seems reasonable (the original log probably wasn't adding much value).

PR #385, opened 2024-09-30
Make Buildkite endpoints configurable

DrJosh9000:
This could be useful when running against a dev instance of Buildkite itself.

### DIFF
--- a/api/client.go
+++ b/api/client.go
@@ -13,7 +13,10 @@ import (
 	"github.com/Khan/genqlient/graphql"
 )
 
-func NewClient(token string) graphql.Client {
+func NewClient(token, endpoint string) graphql.Client {
+	if endpoint == "" {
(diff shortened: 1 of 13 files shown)


SorchaAbel:
@DrJosh9000 this is amazing! I can't believe how fast you did this

SorchaAbel (approved):
I'm not the best person to approve but selfishly I want this and it all LGTM

DrJosh9000:
I tested it and it works. The main wrinkle is that the `graphql-endpoint` and `agent-endpoint` hosts (`agent.buildkite.localhost` and `graphql.buildkite.localhost` in the example) can't map to 127.0.0.1 in `/etc/hosts`. That would cause the controller and any pods it creates to connect to itself, not a dev instance. Running dev Buildkite locally, I edited /etc/hosts to use the LAN address of the machine instead of a loopback address.

PR #389, opened 2024-10-01
Don't schedule using stale job info

### LINKED ISSUE #382: Controller keeps scheduling Pods from a cancelled Build
Steps to reproduce:

1. Set up buildkite/agent-stack-k8s with max-in-flight: 2
1. Push a new commit to an open PR which, it triggers 10 new jobs, controller schedules 2 Pods at a time respecting `max-in-flight: 2`
1. Cancel the build in Buildkite UI / API, all 10 new jobs display as Cancelled in Buildkite UI
1. Observe controller still scheduling Pods for that build flooding the K8S cluster (kubectl describe pod shows BUILDKITE_COMMIT which is cancelled) and preventing other Jobs from being handled

Environment:

- AWS EKS v1.25
- buildkite/agent-stack-k8s v0.15.0

DrJosh9000:
### Why
Fixes #382 

### What
Change the `Handler` interface to accept a new `Job` that includes a channel that is closed when the job information becomes stale. This can be used to time out in the limiter waiting for a token to become available. That way, stale jobs won't be given to the scheduler.

### DIFF
--- a/internal/controller/monitor/monitor.go
+++ b/internal/controller/monitor/monitor.go
@@ -167,55 +167,54 @@ func (m *Monitor) Start(ctx context.Context, handler JobHandler) <-chan error {
 				return
 			}
 
-			jobs := resp.CommandJobs()
-
-			// Wrapped so staleCancel is called before the next loop iteration.
-			func() {
-				// A sneaky way to create a channel that is closed after a duration.
-				// Why not pass directly to handler.Create? Because that might
-				// interrupt scheduling a pod, when all we want is to bound the
-				// time spent waiting for the limiter.
-				staleCtx, staleCancel := context.WithTimeout(ctx, m.cfg.PollInterval)
-				defer staleCancel()
-
-				// TODO: sort by ScheduledAt in the API
-				sort.Slice(jobs, func(i, j int) bool {
-					return jobs[i].ScheduledAt.Before(jobs[j].ScheduledAt)
-				})
-
-				for _, j := range jobs {
-					if staleCtx.Err() != nil {
-						// Results already stale; try again later.
-						return
-					}
-
-					jobTags := toMapAndLogErrors(logger, j.AgentQueryRules)
-
-					// The api returns jobs that match ANY agent tags (the agent query rules)
-					// However, we can only acquire jobs that match ALL agent tags
-					if !agenttags.JobTagsMatchAgentTags(jobTags, agentTags) {
-						logger.Debug("skipping job because it did not match all tags", zap.Any("job", j))
-						continue
-					}
-
-					job := Job{
-						CommandJob: &j.CommandJob,
-						StaleCh:    staleCtx.Done(),
-					}
-
-					logger.Debug("creating job", zap.String("uuid", j.Uuid))
-					if err := handler.Create(ctx, job); err != nil {
-						if ctx.Err() != nil {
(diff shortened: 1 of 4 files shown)


internal/controller/monitor/monitor.go
```diff
@@ -32,8 +32,16 @@ type Config struct {
 	Tags            []string
 }
 
+type Job struct {
+	// The job information.
+	*api.CommandJob
+
+	// Closed when the job information becomes stale.
+	StaleCh <-chan struct{}
```
artem-zinnatullin, line 40: Hm, I thought the fix would be to double-check the cancellation/etc state of the Job/Build before actually scheduling it onto K8S

Having a callback (Channel) when information becomes stale seems like a much more complicated way to achieve that with unclear benefits and requires running a polling loop to keep the channel updated which will load up Buildkite backend quite a lot with these polling checks…

Sorry I'm not very proficient with Golang, can you please help me understand why this approach is better? Thanks!

internal/controller/scheduler/limiter.go
```diff
@@ -93,7 +92,8 @@ func (l *MaxInFlightLimiter) Create(ctx context.Context, job *api.CommandJob) er
 		return nil
 	}
 
-	// Block until there's a token in the bucket.
+	// Block until there's a token in the bucket, or the job information becomes
```
artem-zinnatullin, line 95: I think the intent of the code here doesn't really align with the comment

To block until something is stale is quite meaningless. Perhaps something like this wording could be used
```suggestion
	// Block until there's a token in the bucket, or cancel if the job information becomes stale.
```

internal/controller/monitor/monitor.go
DrJosh9000: If there was no limiter in place then Buildkite would receive a query every `PollInterval` for jobs to run (call this the query loop), and monitor would loop over the result (the inner loop), calling `handler.Create` for each of them. 

In the previous PR on the limiter, I changed how the limiter worked to introduce blocking on the "token bucket", but this causes the bug: a job that was returned from the query a long time ago could be next in the inner loop in the Monitor around `handler.Create`, so that when the limiter can finally get a token, the job could now be cancelled.

With the `StaleCh` approach, the limiter can block until either there's a token, or the job information is stale. If it's stale, then the limiter shouldn't pass the job onto the scheduler and it can return early. Then the query loop can wait until the next `PollInterval` to run the main query.

To double-check the cancellation state of the job before passing it on to the scheduler, or double-checking within scheduler, would mean making another query to Buildkite at that point. 

So while I think my approach doesn't look particularly clean (I could probably make it look nicer), it does avoid the extra query.

artem-zinnatullin: I see, so the idea is to keep the query loop as the main source of truth for up-to-date data and avoid creating a separate code path for explicit job state refresh? Sounds reasonable!

I guess coming from Reactive Streams I'd expect one Channel to be enough to either provide a valid object or indicate cancellation/stale by closing it, The "query loop" logic would then be distributing the updates about jobs to the interested channels or closing those channels when job is cancelled.

Your approach seems to solve that so there is no issue, just speaking out loud :)

artem-zinnatullin (approved):

wolfeidau (approved):
Tough problem.

PR #390, opened 2024-10-02
Tidyups in .buildkite

DrJosh9000:
### What

- Shell tidyups: handlebars, quotes
- Update Go to 1.23
- Refactor env, volumes, and volume mounts using YAML anchors
- Give release step access to the Go caches

### Why

The release step is slow, maybe this will help?


CerealBoy (approved):
LGTM!

PR #391, opened 2024-10-07
Add many more agent config options

DrJosh9000:
### What
Add a subset of agent config options to `values.yaml`, and more git-related options to checkout params.

### Why
There are huge gaps in the configurability of the agent, that so far require:

- building a custom agent image to bake in a custom config, or
- complicated chains of configuration in a pipeline or values file (env var + volume with podSpecPatch + extraVolumeMounts), or
- putting agent config in pipeline-level env vars, that outside this stack are usually prevented from having an effect

There's a need to supply many of the same config options, and also to put a k8s twist on some of them (e.g. instead of `hooks-path`, what about `hooks-volume-source`?)


internal/controller/config/agent_config.go
```diff
@@ -0,0 +1,137 @@
...
+	TraceContextEncoding      *string  `json:"trace-context-encoding,omitempty"`
+
+	// Applies differently depending on the container (agent server vs checkout/
+	// command container).
+	NoPTY            bool `json:"no-pty,omitempty"`
+	NoCommandEval    bool `json:"no-command-eval,omitempty"`
+	NoLocalHooks     bool `json:"no-local-hooks,omitempty"`
+	NoPlugins        bool `json:"no-plugins,omitempty"`
+	PluginValidation bool `json:"plugin-validation,omitempty"`
+
```
wolfeidau, line 33: Would it be good to have the signing warn option? 

This way if you mix agents you can at least avoid 🔥

wolfeidau (changes requested):
Only one observation / question, great work!

README.md
```diff
@@ -547,70 +547,213 @@ steps:
...
     queue: kubernetes
   plugins:
-  - kubernetes:
-      checkout:
-        cloneFlags: -v --depth 1
-        fetchFlags: -v --prune --tags
+    - kubernetes:
+        checkout:
+          cloneFlags: -v --depth 1
+          fetchFlags: -v --prune --tags
```
dabarrell, line 566: Noting that the indent of the `steps` array and `plugins` array don't match with this change

internal/controller/config/agent_config.go
DrJosh9000: I added most of the signing and verification options, and made it default to `warn` if no key is specified. (If a key is specified, they could want `block` by default). WDYT?

README.md
DrJosh9000: Fixed.

internal/controller/config/agent_config.go
wolfeidau: 👍🏻 sounds great, this layered default is a bit more complex but also safer for those who don't even know what signing is which is a big plus.

wolfeidau (approved):

PR #393, opened 2024-10-08
Suppress 'ignored env' warning

DrJosh9000:
### What

Set `BUILDKITE_IGNORED_ENV` to a blank var, to suppress the error message.
Note that this behaviour depends on the next agent release (v3.83.0?)

### Why

agent-stack-k8s circumvents the normal flow of env vars through the agent - it sets all the job vars on the checkout/command containers, when (outside of the k8s stack) the agent wouldn't do this. This never caused a warning until the agent started passing more env vars through the socket, including `BUILDKITE_IGNORED_ENV`, which the agent populates, causing the warning to appear even though the vars are not actually ignored.

What we should do is provide a way to set all the agent configuration that is normally set as "protected" env vars (#391), and then un-circumvent the env var shenanigans the scheduler does (future work). Until then, we can ignore the ignored-env warning.


123sarahj123 (approved):
LGTM :)

CerealBoy (approved):

PR #394, opened 2024-10-08
Fix s/"bool"/"boolean"/ in JSON Schema

DrJosh9000:
Oops. "boolean" is valid, "bool" is not.


DrJosh9000:
> do we need a checker for this file?

Yeah that would probably save some time...

PR #395, opened 2024-10-08
Fix VolumeSource in JSON Schema

DrJosh9000:
Turns out `VolumeSource` is just an artefact of the Go library implementation - the JSON Schema has no concept of `VolumeSource`, only `Volume`. (And "name" is required, per the schema.)

I suppose the user should be able to change the volume name if they want, anyway.

I tested it and it works:

Screenshot 2024-10-08 at 5 39 10 PM


wolfeidau (approved):
👍🏻

PR #406, opened 2024-10-29
Clean up pending pods for cancelled jobs

### LINKED ISSUE #392: Controller does not Delete Pending Jobs from K8S if they're cancelled on Buildkite — flooding the cluster
Testing a fix for #382 with `controller:0.15.0-14-g68932d3` build I found that `buildkite/agent-stack-k8s` apparently does not have any (?) logic to delete Pending Jobs/Pods for cancelled jobs/builds!

We heavily rely on [`Cancel Intermediate Builds` setting in Buildkite (see docs)](https://buildkite.com/docs/pipelines/skipping#cancel-running-intermediate-builds) which cancels in-flight builds on same branch when a new commit is pushed to a PR.

Google Chrome 2024-10-07 15 27 24

Current behavior of the `buildkite/agent-stack-k8s` controller keeps Pending Jobs/Pods in K8S even after after a Buildkite job/build cancelled thus flooding the K8S cluster with resource allocations, then actually starts those jobs and consumes CPU time leading to overspending $$$.

---

Expected behavior:

`buildkite/agent-stack-k8s` controller should send Job/Pod "Delete" request to K8S for a cancelled Buildkite Job that is not in Running state on K8S.

DrJosh9000:
### Why
Fixes #392.

### How
`imagePullBackOffWatcher` is now `podWatcher`, and has a new responsibility: starting and stopping goroutines that poll Buildkite for job state. Each new goroutine should only run for a pod that is in the Pending phase: k8s has accepted the pod, but it isn't running yet. When it enters Running phase, then the agent within the pod can be responsible for handling cancellation.

Starting the cancel checker could be done from the scheduler after it has submitted the job to k8s, but this spreads responsibility between two somewhat distinct parts of the code. Putting it in the artist-formerly-known-as-imagePullBackOffWatcher centralises it, and also means it has a chance to find and clean up pods that were created before a controller restart.

The main downside to this approach is that the additional queries will eat into the GraphQL quota for the user. One possible solution might be the "job handover" idea (i.e. the stack controller acquires jobs like an agent, polls the Agent REST API for cancellation, then hands over the job to the agent in the pod when ready...)


internal/controller/scheduler/pod_watcher.go
```diff
@@ -19,111 +20,166 @@ import (
...
-func (w *imagePullBackOffWatcher) RegisterInformer(
-	ctx context.Context,
-	factory informers.SharedInformerFactory,
-) error {
+func (w *podWatcher) RegisterInformer(ctx context.Context, factory informers.SharedInformerFactory) error {
 	informer := factory.Core().V1().Pods().Informer()
 	if _, err := informer.AddEventHandler(w); err != nil {
 		return err
 	}
+	w.resourceEventHandlerCtx = ctx // 😡
```
CerealBoy, line 88: Is `RegisterInformer` only ever called once? Could the value be clobbered here and / or does it matter?

CerealBoy (approved):
1 question from me, otherwise looks great!

internal/controller/scheduler/pod_watcher.go
DrJosh9000: `RegisterInformer` should be called once from `controller.Run`, and a quick look through the code shows no other calls, so I'm not concerned.

PR #409, opened 2024-10-31
Fix parsing configs with dots in mapping keys

DrJosh9000:
### What
Change viper's key delimiter to something other than `.`.

### Why
Viper unmarshals keys with `.`s as a series of nested maps, instead of as a single string key. Which is frustrating when the Kubernetes convention is to use domain names within keys.


wolfeidau:
@DrJosh9000 did you consider `viper.KeyDelimiter("::")` as per their docs? https://pkg.go.dev/github.com/spf13/viper#readme-unmarshaling

DrJosh9000:
I suppose `::` is unlikely to occur, and it is easier to type than `·`.

wolfeidau (approved):
👍🏻 🐍 🚀

PR #412, opened 2024-11-06
PullIfNotPresent for app containers

### LINKED ISSUE #411: Optimization possible to avoid pulling each container image twice
I think there’s an optimization that can be done in the agent-stack-k8s regarding image pulling. I understand the reasoning for the init-containers/preFlightContainers (see https://github.com/buildkite/agent-stack-k8s/blob/00a778db02818843253416d6833ff0199d8ea13c/internal/controller/scheduler/scheduler.go#L510) and I think that’s a reasonable solution. 

Then however we do another pull for each image here: https://github.com/buildkite/agent-stack-k8s/blob/00a778db02818843253416d6833ff0199d8ea13c/internal/controller/scheduler/scheduler.go#L357 which I believe is not necessary since we just pulled all these images via the init-containers. 

In our clusters a no-op pull takes roughly a second, so this adds a second for each container used. When using many containers (we often have 5+), this adds up to quite a lot of extra time waiting. I believe it’s possible to just change the PullPolicy to Never to avoid this extra pulling, which is fine since we just pre-pulled the images in the init containers.

DrJosh9000:
Open to debating whether or not it should be `PullNever`, but either way, this should shave off some time spent double-pulling containers.

Fixes #411

### DIFF
--- a/internal/controller/scheduler/scheduler.go
+++ b/internal/controller/scheduler/scheduler.go
@@ -354,7 +354,9 @@ func (w *worker) Build(podSpec *corev1.PodSpec, skipCheckout bool, inputs buildI
 		c.Command = []string{"/workspace/tini-static"}
 		c.Args = []string{"--", "/workspace/buildkite-agent", "bootstrap"}
 
-		c.ImagePullPolicy = corev1.PullAlways
+		// The image *should* be present since we just pulled it with an init
+		// container, but weirder things have happened.
+		c.ImagePullPolicy = corev1.PullIfNotPresent
 		c.Env = append(c.Env, containerEnv...)
 		c.Env = append(c.Env,
 			corev1.EnvVar{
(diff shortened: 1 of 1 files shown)


internal/controller/scheduler/scheduler.go
```diff
@@ -354,7 +354,9 @@ func (w *worker) Build(podSpec *corev1.PodSpec, skipCheckout bool, inputs buildI
 		c.Command = []string{"/workspace/tini-static"}
 		c.Args = []string{"--", "/workspace/buildkite-agent", "bootstrap"}
 
-		c.ImagePullPolicy = corev1.PullAlways
+		// The image *should* be present since we just pulled it with an init
+		// container, but weirder things have happened.
+		c.ImagePullPolicy = corev1.PullIfNotPresent
```
wolfeidau, line 359: Does this come with the normal caveat that if you update an image tag the new version won't be pulled? 

So you kind of need to be careful of `latest` tags or the like.

DrJosh9000: Each distinct image is also pulled by an init container with the policy `PullAlways`, so the node should have e.g. "latest" as of a few seconds ago. 

I think changing this is an improvement with respect to concurrently-changing image tags, because if multiple command containers all use the same image, they will now use the one pulled by the init container, instead of all pulling separately and potentially having different images to one another.

wolfeidau: @DrJosh9000 should we include a note in the release notes to avoid surprising people when this behaviour changes?

wolfeidau (approved):
👍🏻 🚂

internal/controller/scheduler/scheduler.go
DrJosh9000: Will do.

PR #413, opened 2024-11-07
Configurable default job metadata

### LINKED ISSUE #408: Feature Request: Ability to add arbitrary labels to the pods that get spun up to execute jobs
I have the build kite agent running in my cluster that is managed by ArgoCD. It doesn't seem like there is any way currently to have the actual jobs that get spun up by the agent have their labels altered? If I was able to add labels to those pods, I would be able to go ArgoCD resource tracking working properly.

DrJosh9000:
### What
Provide a way to supply default metadata for jobs.

### Why
Fixes #408

### DIFF
--- a/README.md
+++ b/README.md
@@ -381,6 +381,40 @@ default-checkout-params:
 ...
 ```
 
+### Default job metadata
+agent-stack-k8s can automatically add labels and annotations to the Kubernetes jobs it creates.
+Default labels and annotations can be set in `values.yaml` with `default-metadata`, e.g.:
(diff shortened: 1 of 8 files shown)


wolfeidau (changes requested):
Looks good 👍🏻 

Is this something we should add to the README?

Seems handy

DrJosh9000:
> Is this something we should add to the README?

Done

wolfeidau (approved):
👍🏻 🚀

PR #417, opened 2024-11-14
Configurable stale job data timeout

DrJosh9000:
`poll-interval` is probably not be the ideal timeout for stale job data in many cases, so make it configurable and default it to 10s (duration chosen by guesswork). 

Increasing the timeout means more jobs from the same GraphQL query can be created before the next poll, but increases the chances of scheduling an already-cancelled job as a pod.


SorchaAbel (approved):
Looks great!

Issue #278, opened 2024-03-21
feat: add monitoring to the agent-stack and job (/metrics and labels)


42atomys:
The goal of this feature request is to enhance the observability and monitoring capabilities of agent-stack within a Kubernetes environment. By introducing a metrics endpoint for each agent-stack controller and enriching jobs with labels and annotations, we aim to align with the Kubernetes metrics API server. This alignment will empower users with comprehensive monitoring tools, facilitating better management and analysis of their deployments.

### Objectives
**Metrics Endpoint for Agent-Stack Controllers**: Implement a `/metrics` endpoint for controller within the agent-stack. This endpoint will expose various metrics relevant to the operations and performance of the agent-stack wit prometheus (duplicate of #102), making it possible for monitoring tools to scrape and aggregate this data.

**Labels and Annotations for Jobs**: Enhance jobs running in Kubernetes with appropriate labels and annotations. This will ensure our jobs are fully compatible and discoverable by the Kubernetes metrics API server, allowing for seamless integration into users' existing monitoring setups.

**Probes for Agent Stack**: Integrate readiness and liveness probes into the agent stack. These probes will improve the reliability and stability of the agent-stack by enabling Kubernetes to automatically manage the lifecycle of pods based on their health status.

42atomys:
I will open a PR in few days :100:

DrJosh9000:
Sounds good, looking forward to it!

DrJosh9000:
Hi @42atomys, we're wondering if you're still intending to open a PR for this? I think they would all be wonderful additions, even as separate parts!

42atomys:
Hi @DrJosh9000, I work on buildkite repo only on work hours and I'm currently out of office for few weeks. If you want to contribute before my return, you can ^^

42atomys:
UPDATE: I no longer work with Buildkite, so I will not be integrating this proposal.

PR #419, opened 2024-11-14
Add Prometheus metrics

### LINKED ISSUE #102: Expose metrics
Via [prometheus/client_golang](https://prometheus.io/docs/guides/go-application/) probably.

- Jobs processed
- GraphQL query latency
- API error rates
- and more?

DrJosh9000:
### What
Adds a Prometheus metrics endpoint, and a whole heap of metrics.

### Why
Fixes #102.

Addresses the metric part of #278.

Among other reasons, extra observability into the controller will likely be needed to keep digging into #302.

### Show me the charts
Two ways to get the number of jobs the deduper thinks is running:
Screenshot 2024-11-27 at 1 28 01 PM
Screenshot 2024-11-27 at 1 29 36 PM

Available limiter tokens:
Screenshot 2024-11-27 at 1 30 16 PM

Median time between querying a job and scheduling it:
Screenshot 2024-11-27 at 1 31 15 PM

### DIFF
--- a/internal/controller/deduper/metrics.go
+++ b/internal/controller/deduper/metrics.go
@@ -1,0 +1,80 @@
+package deduper
+
+import (
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
+)
+
+const (
+	promNamespace = "buildkite"
+	promSubsystem = "deduper"
+)
+
+// Overridden by New to return len(inFlight).
+var jobsRunningGaugeFunc = func() int { return 0 }
+
+var (
+	_ = promauto.NewGaugeFunc(prometheus.GaugeOpts{
+		Namespace: promNamespace,
+		Subsystem: promSubsystem,
+		Name:      "jobs_running",
+		Help:      "Current number of running jobs according to deduper",
+	}, func() float64 { return float64(jobsRunningGaugeFunc()) })
+
+	jobHandlerCallsCounter = promauto.NewCounter(prometheus.CounterOpts{
+		Namespace: promNamespace,
+		Subsystem: promSubsystem,
+		Name:      "job_handler_calls_total",
+		Help:      "Count of jobs that were passed to the next handler in the chain",
+	})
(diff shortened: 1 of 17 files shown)


internal/controller/deduper/metrics.go
```diff
@@ -0,0 +1,80 @@
...
+// Overridden by New to return len(inFlight).
+var jobsRunningGaugeFunc = func() int { return 0 }
+
+var (
+	_ = promauto.NewGaugeFunc(prometheus.GaugeOpts{
+		Namespace: promNamespace,
+		Subsystem: promSubsystem,
+		Name:      "jobs_running",
+		Help:      "Current number of running jobs according to deduper",
+	}, func() float64 { return float64(jobsRunningGaugeFunc()) })
```
CerealBoy, line 22: With the `_ = ` is this merely a template for adding the rest of the vars below?

CerealBoy (approved):
This is great! I had 1 inline question, and there's other similar examples that I didn't add the same question.

As an aside, have we thought about other providers / services besides prometheus? Would we be able to use OpenTelemetry to achieve the same thing without being tied specifically to prometheus?

internal/controller/deduper/metrics.go
DrJosh9000: It's `_` merely because this gauge doesn't need to be referred to within the package later on - `promauto` registers the metric, then the value is obtained during metric scrape through the callback (unlike most of the other metrics, where the value is set or incremented or added or observed on the metric).

DrJosh9000:
> As an aside, have we thought about other providers / services besides prometheus?

Not particularly. I'm more familiar with Prometheus than anything else 🤷 

> Would we be able to use OpenTelemetry to achieve the same thing without being tied specifically to prometheus?

I believe so. The types of metrics defined here are a subset of the types supported in OTel. Moving across shouldn't be too difficult given the OTel Go SDK has a Prometheus Exporter.

internal/controller/deduper/metrics.go
CerealBoy: Excellent, thank you for the explanation!

PR #421, opened 2024-11-15
Allow custom mount path

mcncl:
## Changes
I don't know if this will work, but I don't know if we can test this stack locally either, so this is a PR in the hope that someone tears it apart and tells me what I did wrong, so that I can learn from it.

- Allows the custom setting of mount path
- Uses a default mount path of "workspace" if none is provided
- Adds a test to ensure that the mount path is set correctly when a custom path is provided


mcncl:
This isn't the thing they wanted, I misunderstood. Closing for now.

PR #423, opened 2024-11-17
Rename Create -> Handle, + more logs and comments to aid debugging

DrJosh9000:
Rename the `Create` method of `JobHandler` to `Handle`, to make it clear that not every `JobHandler` implementation actually creates pods.
Add extra comments and some more debug log calls to make the flow of jobs easier to understand.

### DIFF
--- a/internal/controller/controller.go
+++ b/internal/controller/controller.go
@@ -37,6 +37,8 @@ func Run(
 		}()
 	}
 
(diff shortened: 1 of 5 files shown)


SorchaAbel (approved):
The comments are so good. And great idea to rename from create to handle.

PR #424, opened 2024-11-18
Split limiter and deduplicator

DrJosh9000:
Limiter shouldn't do two things at once. Separate the two concerns (limiting the maximum number of jobs, and preventing duplicate jobs).

Also rearranges some related code - there was too much happening within the scheduler package.


wolfeidau (approved):
👍🏻 Nice work

PR #422, opened 2024-11-15
Implement a configurable workspace volume specification

jfmeachum:

PR #428, opened 2024-11-19
Configurable workspace volume

DrJosh9000:
This is #422, but the primary difference is that the workspace volume configuration is in the top-level config instead of inside `agent-config`. I think the workspace volume is central enough to the operation of the stack that it ought to be at that level, even though there are similar volumes (hooks, plugins). (Happy to revise this.)

Other than that, I've made some fixes, minor rearranging, and added to the example config and test.


wolfeidau (approved):
nice 🚀 👍🏻

PR #429, opened 2024-11-19
Add waitgroup around jobHandlerWorker

DrJosh9000:
In #427 I forgot to wait on the worker goroutines. This might interfere with the intended improvement, since the context and channel would close much sooner causing the workers to return early.


SorchaAbel (approved):
LGTM

PR #431, opened 2024-11-21
Match tags in pod watcher

DrJosh9000:
### What
The pod watcher should only take ownership of pods whose labels match the agent tags configured on the controller. 

### Why
The controller queries for jobs matching any tags, but then filters to ensure a match for all tags. So each controller only really "owns" _those_ jobs. 

Currently the pod watcher does no such filtering, so it runs cancel checkers and looks for ImagePullBackoff on _all_ pods with job UUIDs. That's fine if there's only a single controller in the cluster, but if there are multiple controllers in a cluster (with distinct tags) then there is redundant pod watcher work.

We could try to record all job UUIDs started by this controller and look for those, but such a list would become invalid on a controller restart.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,8 +1,6 @@
 module github.com/buildkite/agent-stack-k8s/v2
 
-go 1.22.6
-
-toolchain go1.22.7
+go 1.23.3
 
 require (
 	github.com/Khan/genqlient v0.7.0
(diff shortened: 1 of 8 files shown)


wolfeidau (approved):
Looks good ✨ 🚀

PR #433, opened 2024-11-22
adding pull policy to agent container

srichardbd:
We are seeing QPS image pull errors with our internal registry, we would like to add the abilty to specify imagePullPolicy on the agent containers


srichardbd:
we are trying podSpecPatch

PR #436, opened 2024-11-26
Refine informer usage in deduper

DrJosh9000:
Apply things learned in #432 to the deduper, mainly to reduce debug log noise.

Since we can use the information provided by Informer to tell when a job transitions between running and not-running without having to consult the deduping map, we can avoid logging "was already {,not-}in-flight!" when the lack of a state change is expected.

### DIFF
--- a/internal/controller/deduper/deduper.go
+++ b/internal/controller/deduper/deduper.go
@@ -32,12 +32,12 @@ type Deduper struct {
 
 // New creates a Deduper.
 func New(logger *zap.Logger, handler model.JobHandler) *Deduper {
-	l := &Deduper{
+	d := &Deduper{
 		handler:  handler,
 		logger:   logger,
 		inFlight: make(map[uuid.UUID]bool),
 	}
-	return l
+	return d
 }
 
 // RegisterInformer registers the limiter to listen for Kubernetes job events,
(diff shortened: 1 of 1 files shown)


internal/controller/deduper/deduper.go
```diff
@@ -156,24 +170,27 @@ func (d *Deduper) markRunning(id uuid.UUID) {
...
 
-// markComplete records a job as not in-flight.
-func (d *Deduper) markComplete(id uuid.UUID) {
+// markNotRunning records a job as not in-flight.
+func (d *Deduper) markNotRunning(id uuid.UUID, source string) {
 	// Change state from in-flight to not in-flight.
 	numInFlight, ok := d.casa(id, false)
 	if !ok {
-		d.logger.Debug("markComplete: job was already not-in-flight!",
+		d.logger.Debug("markNotRunning: job was already not-in-flight!",
```
wolfeidau, line 183: Is this double negative something which is easy to understand?

wolfeidau (changes requested):
Just one observation.

internal/controller/deduper/deduper.go
DrJosh9000: Fair point, I've renamed it back to `markFinished` (`markComplete` sounded to me as though the job succeeded, when it might not have).

DrJosh9000: Or maybe it should be `unmarkRunning`? 🤔

DrJosh9000: Maybe including the method name is not useful, given that zap includes the file and line number.

wolfeidau (approved):
👍🏻 🚀

PR #438, opened 2024-11-27
Catch duplicate jobs in scheduler

DrJosh9000:
There's definitely some flow in here where jobs that were already created try to get created again. While we figure that out, catch the `AlreadyExists` reason and make the deduper treat it as though the job is running (because there _is_ a job)


wolfeidau (approved):
👍🏻

PR #442, opened 2024-12-03
Add job watcher

### LINKED ISSUE #302: Controller stops accepting jobs from the cluster queue
We have the `agent-stack-k8s` up and running and works fine for a while. However, it suddenly stops accepting new jobs and the last thing it outputs is (we turned on debug):
```
2024-04-08T11:38:23.100Z	DEBUG	limiter	scheduler/limiter.go:77	max-in-flight reached	{"in-flight": 25}
```

We currently only have a single pipeline, single cluster and single queue. When this happens there are no jobs or pods named `buildkite-${UUID}` in the k8s cluster. Executing `kubectl -n buildkite rollout restart deployment agent-stack-k8s` makes the controller happy again and it starts jobs from the queue.

I suspect that there is something that should decrement the `in-flight` number, but fails to do so. We are now running a test where this number is set to 0 to see if that works around the problem.

DrJosh9000:
### What
Add a component that looks for k8s jobs in two troublesome states:
* Jobs that finish without ever creating a pod
* Jobs that stall for too long without creating a pod

When such a job is detected, it grabs k8s events for the job and puts them in a failure message to fail the corresponding BK job.

### Why
May fix #302, as well as configuration issues that don't block job creation but prevent running the pod (e.g. missing user, invalid labels, ...)

### DIFF
--- a/.buildkite/rbac.yaml
+++ b/.buildkite/rbac.yaml
@@ -34,6 +34,12 @@ rules:
       - pods/eviction
     verbs:
       - create
+  - apiGroups:
+      - ""
+    resources:
+      - events
+    verbs:
+      - list
 ---
 apiVersion: v1
 kind: ServiceAccount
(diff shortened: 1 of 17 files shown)


CerealBoy (approved):
🚀

PR #449, opened 2024-12-11
Add config option to disable command modification check

DrJosh9000:
### What
Provide the option to allow PodSpecPatch to modify container commands and args that are normally protected.

### Why
It's a recurring request to be able to do this with PodSpecPatch. #461 provided a means to do it for command containers (by transforming the patch), but sometimes people want to alter the other containers for whatever reason. Allowing the direct modification of any container command (that has been carefully arranged by the scheduler) remains a "foot-gun", so it is now an option that must be enabled.

### Notes
I've regenerated the TOC in the README because it was slightly out of date.


defgenx:
Does this PR is still ongoing ? I had to duplicate DinD image to be able to run the container. Would be nice to be able to do that through pod-spec-patch. :)

DrJosh9000:
@defgenx Between annual leave and merging #461 (which allows podSpecPatch-ing commands for some containers) I haven't been pushing this one as hard. What does your use case look like more specifically? Which containers are you trying to alter and how?

DrJosh9000:
> We have a DinD container running next to the agent and the default command of the DinD image is not on localhost; making it impossible to use as is.

@defgenx I'll look into rebasing this PR and making it work, but your use case does sound as though a [sidecar container](https://github.com/buildkite/agent-stack-k8s?tab=readme-ov-file#sidecars) could work - the controller makes no attempt to change the command or args of sidecars.

defgenx:
> sidecar container

Mhhhh I understand your point, but not really.
Just knowing we are not sure it's started before the job and the need to add `wait-for-it` script is a no-go. We just want a DinD container to be started every time an agent is started without having to specify anything in my pipeline yaml.

Thanks for updating it again. Hope it will be added soon. 💪

tomowatt (approved):
LGTM 🚀

PR #384, opened 2024-09-27
Include PR info into release version

artem-zinnatullin:
Apparently, any PR to this repo whether made by untrusted 3rd-party or by a member of Buildkite team which successfully ran CI publishes:

- [Helm charts](https://github.com/buildkite/agent-stack-k8s/pkgs/container/helm%2Fagent-stack-k8s) 
- [Docker Image of Controller](https://github.com/buildkite/agent-stack-k8s/pkgs/container/agent-stack-k8s%2Fcontroller)

They get versioned via `git describe` like so:

image
image

Where `g59acf83` in `ghcr.io/buildkite/agent-stack-k8s/controller:0.15.0-5-g59acf83` indicates a short commit sha.

The security issue here is that for a typical end user who wants to grab what they think is latest build they just go to the [packages](https://github.com/orgs/buildkite/packages?repo_name=agent-stack-k8s) linked right on GitHub and grab a latest build like the `controller:0.15.0-5-g59acf83`

But in reality `controller:0.15.0-5-g59acf83` is build from an unmerged PR https://github.com/buildkite/agent-stack-k8s/pull/381! 

In worst case scenario such PRs can be made by a malicious 3rd-party doing god knows what like leaking user credentials or doing Remote Code Execution in their CI setup, PRs are not yet merged onto main branch of original repo but already appear in the available packages list: here is controller build from my old PR as I found out now https://github.com/buildkite/agent-stack-k8s/pkgs/container/agent-stack-k8s%2Fcontroller/218244275?tag=0.11.0-1-g43e4777

As of right now if I understand correctly a member of Buildkite team needs to manually approve CI to run on a 3rd-party PR, which should prevent malicious code from being published, but for the end user nothing indicates the difference between the releases on the package page.

This PR adds PR information into the release version name so it's clear that you're not taking some latest main branch commit release: `controller:0.15.0-5-g59acf83` → `controller:0.15.0-5-g59acf83-PR-381`

---

I do want to emphasize that having 3rd-party PRs release as packages is actually very useful for testing the changes by the community and I'd love Buildkite to keep that feature, but to do so in a secure manner the release names need to clearly indicate that it is an untrusted 3rd-party PR release.

PR #450, opened 2024-12-11
Push images to Public ECR

DrJosh9000:
### What
Publish dev builds (anything CI has run on but is not tagged with a version) to Public ECR only - in specific `-dev` repo paths.

Publish release builds to both Public ECR and GHCR.

### Why
Addresses one of the key issues raised in #384 - currently it is not easy to disambiguate supported release builds from random dev builds just by looking at the GitHub Packages page. Moving the dev builds to separate dev repos accomplishes this.

Public ECR was also chosen because it is easy to set up with Terraform, and supports OIDC-assumable roles for pushing. 

At some point we should also push releases to Buildkite Packages 😄


yob (approved):
LGTM, let's try it and iterate if we need to 👍

PR #451, opened 2024-12-12
Schedule jobs in priority order

### LINKED ISSUE #445: Feature request: support Buildkite "Job prioritization"
As discussed with @DrJosh9000 controller should eventually add support for Buildkite ["Job prioritization"](https://buildkite.com/docs/pipelines/configure/workflows/managing-priorities) so that jobs are scheduled onto Kubernetes with respect to their Buildkite priority.

>Job priority is considered before jobs are dispatched to [agent queues](https://buildkite.com/docs/agent/v3/queues), so jobs with higher priority are assigned before jobs with lower priority, regardless of which has been longest in the queue. Priority only applies to command jobs, including plugin commands.

Right now controller seems to take jobs from Buildkite API in batches and each batch is shuffled (…?) with no respect neither to the creation time nor priority:

https://github.com/buildkite/agent-stack-k8s/blob/06a36017bb4d01d42322e678a9e45b7d271c3c1e/internal/controller/monitor/monitor.go#L216-L226

DrJosh9000:
### What
Sort jobs by priority (descending) before sending them through to be scheduled. Highest priority jobs should be scheduled first, lowest priority jobs scheduled last.

### Why
Fixes #445

### Notes
The graphQL queries remove `scheduledAt` (since it stopped being used with #427) and adds `priority`. My text editor also reformatted the GraphQL. The integration tests also hit a nil panic (these are sometimes flaky), so I have fixed the two instances of that particular path.

### DIFF
--- a/api/generated.go
+++ b/api/generated.go
@@ -6,7 +6,6 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"time"
 
 	"github.com/Khan/genqlient/graphql"
 )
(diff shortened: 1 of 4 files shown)


wolfeidau (approved):
👍🏻 NICE 🚀

PR #453, opened 2024-12-13
Allow running command containers as non-root user with EBS backed (and other) workspace volumes

patrobinson:
When using a storage provisioner, such as aws-ebs-csi, the default permission mask for the volume mount is 755 and [there's no way to change it](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/2211).

That means when you specify a workspace from one of these provisioners you get a volume owned by root, which is not-writable by non-root users:

config.yaml snipped
```yaml
config:
  agent-config:
    hooksVolume:
      name: buildkite-hooks
      configMap:
        defaultMode: 493
        name: buildkite-agent-hooks
  workspace-volume:
    ephemeral:
      volumeClaimTemplate:
        spec:
          accessModes: ["ReadWriteOnce"]
          resources:
            requests:
              storage: 1Gi
    name: workspace-volume
```

configMap pre-checkout hook:
```
#!/bin/sh

ls -l /
ls -l /workspace
ls -l /workspace/sockets || true
```

Screenshot 2024-12-13 at 3 13 02 pm

This means using a command container that has the `USER` set to something other than root isn't possible.
Firstly our agent can't start, because it needs access to write a job-api socket file. You get:
```
Error: Error setting up Job API: creating job API server: creating socket server: creating socket directory: mkdir /workspace/sockets: permission denied
```

This can be worked around by mounting an in-memory container to the `/workspace/sockets` directory with the following config snipped:
```yaml
  pod-spec-patch:
    containers:
      - name: container-0
        volumeMounts:
          - name: socket-dir
            mountPath: /workspace/sockets
    volumes:
      - name: socket-dir
        emptyDir:
          medium: Memory
```

But the command container still has no ability to write to /workspace directory, which is a common use case.

We already handle this issue within the checkout container by ensuring all files are created as the same user as the SecurityContext provided https://github.com/buildkite/agent-stack-k8s/blob/main/internal/controller/scheduler/scheduler.go#L837-L838

But since the volume as mounted as 755, the checkout container has no permission to write to the directory.

I tried to patch this by telling the checkout container to change ownership of the directory to the relevant user before it changes users,
```
                checkoutContainer.Args = []string{fmt.Sprintf(`set -exufo pipefail
 addgroup -g %d buildkite-agent
 adduser -D -u %d -G buildkite-agent -h /workspace buildkite-agent
+chown -R %d:%d /workspace
 su buildkite-agent -c "%s && buildkite-agent-entrypoint bootstrap"`,
+                       podGroup,
+                       podUser,
                        podGroup,
                        podUser,
                        gitConfigCmd,
```

But it causes the agent to fail to start with:
```
2024-12-13 04:00:30 INFO   Build Path doesn't exist, creating it (/workspace/build)
fatal: failed to create builds path: mkdir /workspace/build: permission denied
```

So I landed on this solution, copying the existing implementation that checkout uses to ensure it runs as root then chown the directory in the init container.

Result:
Screenshot 2024-12-13 at 3 25 38 pm


internal/controller/scheduler/scheduler.go
```diff
@@ -718,6 +654,126 @@ func PatchPodSpec(original *corev1.PodSpec, patch *corev1.PodSpec) (*corev1.PodS
...
+			RunAsGroup:   ptr.To[int64](0),
+			RunAsNonRoot: ptr.To[bool](false),
+		}
+
+		fmt.Fprintf(&containerArgs, `set -exufo pipefail
+adduser -D -u %d -G root -h /workspace buildkite-agent
+chown -R %d /workspace`,
+			podUser,
+			podUser,
+		)
```
patrobinson, line 705: This could be refactored, it's pretty similar to the createCheckoutContainer code

But gathering early feedback first.

internal/controller/scheduler/scheduler.go
```diff
@@ -718,6 +654,126 @@ func PatchPodSpec(original *corev1.PodSpec, patch *corev1.PodSpec) (*corev1.PodS
...
+
+	// If the group is not root, but the user is root, I don't think we NEED to do anything. It's fine
+	// for the user and group to be root for the checked out repo, even though the Pod's security
+	// context has a non-root group.
+	default:
+		securityContext = nil
+	}
+	// Init containers. These run in order before the regular containers.
```
DrJosh9000, line 713: Most of this (very long) comment applies to the image-pull-check init containers or init containers as a whole, but this method is only creating the `copy-agent` init container. So I don't think the comment should be moved from where it was, or at least, maybe it could be moved to some sort of notes-for-devs file...

internal/controller/scheduler/scheduler.go
```diff
@@ -718,6 +654,126 @@ func PatchPodSpec(original *corev1.PodSpec, patch *corev1.PodSpec) (*corev1.PodS
...
+		// The init container needs to be run as root to create the user and give it ownership to the workspace directory
+		securityContext = &corev1.SecurityContext{
+			RunAsUser:    ptr.To[int64](0),
+			RunAsGroup:   ptr.To[int64](0),
+			RunAsNonRoot: ptr.To(false),
+		}
+
+		fmt.Fprintf(&containerArgs, `set -eufo pipefail
+addgroup -g %d buildkite-agent
+adduser -D -u %d -G buildkite-agent -h /workspace buildkite-agent
```
DrJosh9000, line 684: `chown` ought to work without `addgroup`/`adduser`, and nothing within this container is going to refer to the UID/GID by name, so these can be skipped.

internal/controller/scheduler/scheduler.go
```diff
@@ -718,6 +654,126 @@ func PatchPodSpec(original *corev1.PodSpec, patch *corev1.PodSpec) (*corev1.PodS
...
+	case podUser != 0 && podGroup == 0:
+		//The init container needs to be run as root to create the user and give it ownership to the workspace directory
+		securityContext = &corev1.SecurityContext{
+			RunAsUser:    ptr.To[int64](0),
+			RunAsGroup:   ptr.To[int64](0),
+			RunAsNonRoot: ptr.To[bool](false),
+		}
+
+		fmt.Fprintf(&containerArgs, `set -exufo pipefail
+adduser -D -u %d -G root -h /workspace buildkite-agent
```
DrJosh9000, line 701: Similarly

DrJosh9000 (commented):
This seems like a good solution.

patrobinson:
@DrJosh9000 thanks, I've marked this as ready for review

internal/controller/scheduler/scheduler.go
```diff
@@ -718,6 +654,79 @@ func PatchPodSpec(original *corev1.PodSpec, patch *corev1.PodSpec) (*corev1.PodS
...
+			RunAsUser:    ptr.To[int64](0),
+			RunAsGroup:   ptr.To[int64](0),
+			RunAsNonRoot: ptr.To(false),
+		}
+
+		fmt.Fprintf(&containerArgs, `set -eufo pipefail
+chown -R %d:%d /workspace`,
+			podUser,
+			podGroup,
+		)
```
DrJosh9000, line 686: `-u`: unneeded since there are no variable expansions
`-o pipefail`: probably unneeded since there are no pipes AFAICT

Either way, all of the shell options can be set as args to `ash` itself instead of needing to call `set` within the script.

```suggestion
		fmt.Fprintf(&containerArgs, "chown -R %d:%d /workspace\n", podUser, podGroup)
```

internal/controller/scheduler/scheduler.go
```diff
@@ -718,6 +654,79 @@ func PatchPodSpec(original *corev1.PodSpec, patch *corev1.PodSpec) (*corev1.PodS
...
+		securityContext = &corev1.SecurityContext{
+			RunAsUser:    ptr.To[int64](0),
+			RunAsGroup:   ptr.To[int64](0),
+			RunAsNonRoot: ptr.To[bool](false),
+		}
+
+		fmt.Fprintf(&containerArgs, `set -exufo pipefail
+chown -R %d /workspace`,
+			podUser,
+		)
```
DrJosh9000, line 699: Similarly
```suggestion
		fmt.Fprintf(&containerArgs, "chown -R %d /workspace\n", podUser)
```

internal/controller/scheduler/scheduler.go
```diff
@@ -718,6 +654,79 @@ func PatchPodSpec(original *corev1.PodSpec, patch *corev1.PodSpec) (*corev1.PodS
 	return &patchedSpec, nil
 }
 
+func (w *worker) createInitContainer(podSpec *corev1.PodSpec, workspaceVolume *corev1.Volume) corev1.Container {
```
DrJosh9000, line 657: Since init containers are a broader topic than the single init container returned here:

```suggestion
func (w *worker) createWorkspaceSetupContainer(podSpec *corev1.PodSpec, workspaceVolume *corev1.Volume) corev1.Container {
```

or maybe `createCopyAgentContainer` ?

DrJosh9000 (approved):
LGTM with one name change. I also have some minor cleanup suggestions but these also apply to the existing `createCheckoutContainer`, so feel free to split off into a followup.

PR #459, opened 2024-12-17
Image check improvements

### LINKED ISSUE #415: Use PullIfNotPresent for images that have a digest?
Follow-up to: https://github.com/buildkite/agent-stack-k8s/issues/411

In the init-container logic https://github.com/buildkite/agent-stack-k8s/blob/00a778db02818843253416d6833ff0199d8ea13c/internal/controller/scheduler/scheduler.go#L510 we use `PullAlways` to ensure the images are present. Pulling an image that already exists takes ~1s in our clusters, which adds up quickly when using many containers. 

We could optimize this further by using `PullIfNotPresent` instead `PullAlways` for any images that contain a digest, e.g. `ghcr.io/buildkite/agent@sha256:17fbf3e3d5574ba4dc6e7436391635b0a4a4a7f78a751a372e2456d2d72f541d`. Since the exact version is pinned by the hash, we don't need to pull if the image is present since it's guaranteed that we already have the correct version.

This would change the observed behavior slightly in the case where the image doesn't exist in the upstream registry anymore. Instead of failing consistently, any nodes that have the image cached would keep using it.

(Feel free to reject this idea, it might not be worth the extra complexity.)

DrJosh9000:
### What

1. Make the default image pull policy configurable. Stop setting it to `PullIfNotPresent` unilaterally - set it to the default if not set already.
2. Move the creation of image check init containers to _after_ applying PodSpecPatches, in order to catch images changed by the patches.
3. Check each image ref is well-formed by parsing it.
4. Derive the image pull policy to use for an image check from the policy set on the container. If none is set there, use a new configurable default. If that is not set either, default to `PullAlways` unless the image has a digest, in which case use `PullIfNotPresent` (#415)
5. Since a container with pull policy `Never` can fail for a new reason (`ErrImageNeverPull`), check for this new reason.
6. A minor consistency cleanup in `jobWatcher` - `stalledJobs` is keyed by job UUID.

### Why
1. If someone goes to the trouble of setting `imagePullPolicy` on a container, we should respect that.
2. PodSpecPatch can introduce images that slip through the pre-flight checks. (Nobody should be using PodSpecPatch to mutate the image checks currently - it would be hard to do reliably when there is more than 1 distinct image, because iteration order is essentially random.)
3. We can avoid wasting k8s resources by parsing image refs ourselves prior to creating a k8s job, and the parsed ref is needed to decide the pull policy based on whether there is a digest or not.
4. Fixes #415 - to control what the pull policy for an image is in an image-check init container, one merely needs to set it on the corresponding containers in their `podSpec{,Patch}`, or configure a default when deploying the stack.
5. While testing I noticed that the informer seems to inform `podWatcher` of a pending pod with waiting container in this status only once, which makes it hard to apply a grace period in the same way as `ImagePullBackOff`. Thus the image pull checks now work in a separate goroutine looping over the current list of pending and running pods, rather than only on informer events. However this means `ImagePullBackOff` is failed slightly faster!


internal/controller/scheduler/job_watcher.go
```diff
@@ -58,7 +58,7 @@ func NewJobWatcher(logger *zap.Logger, k8sClient kubernetes.Interface, cfg *conf
 		logger:       logger,
 		k8s:          k8sClient,
 		cfg:          cfg,
-		stallingJobs: make(map[*batchv1.Job]struct{}),
+		stallingJobs: make(map[uuid.UUID]*batchv1.Job),
```
wolfeidau, line 61: Is there a mutex for stalling jobs or is covered by another?

internal/controller/scheduler/scheduler.go
```diff
@@ -19,6 +21,8 @@ import (
 
 	"github.com/buildkite/agent/v3/clicommand"
 
+	"github.com/distribution/reference"
```
wolfeidau, line 24: Handy library, have starred that one ⭐

wolfeidau (changes requested):
Just a wee bit of feedback.

internal/controller/scheduler/job_watcher.go
DrJosh9000: Yes, `stallingJobsMu` on line 40.

internal/controller/scheduler/scheduler.go
DrJosh9000: I tried finding one among the k8s libraries, and discovered they were using this under the hood.

wolfeidau (approved):

PR #461, opened 2024-12-19
Allow patching some container commands/args

DrJosh9000:
It came to me last night in a dream... wait, why am I dreaming about work? 🫤

### What
Allow patching some container commands and args with PodSpecPatch: sidecars and new containers added by patch are allowed to make any command/args modification, and command container patches are interposed back into `BUILDKITE_COMMAND` form as we do for non-patch containers. 

### Why
The command modification check was added to prevent breaking the pod. I wrote an explanation in #449 - if `buildkite-agent start` isn't run in the agent container, or `buildkite-agent bootstrap` isn't run in the checkout or command containers, the pod _will_ break in various ways.

However, it was always too strict: 
* sidecar containers never have `buildkite-agent` added to it, so they should be modifiable
* new containers added via PodSpecPatch also don't have the command interposition applied, so they should be allowed to have commands

What about command containers? I think if a user tried to override the command/args of a command container, they probably expect the patch to be _interpreted the same way_ as we do currently for command containers specified with `podSpec`: substitute `buildkite-agent bootstrap` for the command and interpose the supplied command into `BUILDKITE_COMMAND`.

Original command containers can be identified by having had this transform already applied, so we just need to match them against the patch by the same process that strategic merge patch uses: container name.

Patching the checkout container is more complicated - it's a container we add rather than one the user supplies anywhere else, with scripts for adjusting user and group IDs, etc. What do we do with a user supplied command? (schlep it into a checkout hook? that could work... 🤔 but what if the user supplied a checkout hook already?)

### DIFF
--- a/internal/controller/scheduler/scheduler.go
+++ b/internal/controller/scheduler/scheduler.go
@@ -43,6 +43,11 @@ const (
 
 var errK8sPluginProhibited = errors.New("the kubernetes plugin is prohibited by this controller, but was configured on this job")
 
+var (
+	commandContainerCommand = []string{"/workspace/tini-static"}
+	commandContainerArgs    = []string{"--", "/workspace/buildkite-agent", "bootstrap"}
+)
+
 type Config struct {
 	Namespace                   string
 	Image                       string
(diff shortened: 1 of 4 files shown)


wolfeidau (approved):
Looks good, pretty straight forward 🎉 ☁️ 💤

PR #471, opened 2025-01-14
Add note about DSA keys deprecation, change example to RSA

petetomasik:
Deprecation of DSA keys from OpenSSH can be found [here](https://www.openssh.com/txt/release-9.9). This took effect in the `buildkite/agent-base` OS images we use to build the `buildkite/agent` image as of [`v3.88.0`](https://github.com/buildkite/agent/compare/v3.87.1...v3.88.0).

### DIFF
--- a/README.md
+++ b/README.md
@@ -288,12 +288,14 @@ If you need to use git ssh credentials in your job containers, we recommend one 
 1. Use a container image that's based on the default `buildkite/agent` docker image and preserve the default entrypoint by not overriding the command in the job spec.
 2. Include or reproduce the functionality of the [`ssh-env-config.sh`](https://github.com/buildkite/docker-ssh-env-config/blob/-/ssh-env-config.sh) script in the entrypoint for your job container image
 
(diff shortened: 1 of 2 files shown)


DrJosh9000 (approved):
I assume `SSH_PRIVATE_RSA_KEY` works the same as `SSH_PRIVATE_DSA_KEY`? I haven't tested this.

petetomasik:
> I assume `SSH_PRIVATE_RSA_KEY` works the same as `SSH_PRIVATE_DSA_KEY`? I haven't tested this.

Yes, any of the env vars defined here will work - https://github.com/buildkite/docker-ssh-env-config?tab=readme-ov-file#docker-ssh-env-config. A [PR](https://github.com/buildkite/docker-ssh-env-config/pull/6) was already by created by a customer to remove support for `SSH_PRIVATE_DSA_KEY`. We can start with warnings/notes on all the repos that use this wrapper before finally removing DSA key support from the wrapper itself.

PR #484, opened 2025-01-24
fix(scheduler): pod-spec-patch was not working on default init containers

defgenx:
### Description
This PR resolves an issue where the `PodSpecPatch` for default `initContainers` was not functioning correctly in the scheduler.

### Changes
- Fixed duplication of init containers in `podSpecPatch` when the init-container exists.
- Update the documentation for patching resources.
- Ensure the documentation of pod-spec-patch is aligned with what the code is doing.

### DIFF
--- a/internal/controller/scheduler/scheduler.go
+++ b/internal/controller/scheduler/scheduler.go
@@ -560,7 +560,7 @@ func (w *worker) Build(podSpec *corev1.PodSpec, skipCheckout bool, inputs buildI
 		)
 	}
 
-	initContainers := []corev1.Container{w.createWorkspaceSetupContainer(podSpec, workspaceVolume)}
+	podSpec.InitContainers = []corev1.Container{w.createWorkspaceSetupContainer(podSpec, workspaceVolume)}
 
 	// Only attempt the job once.
 	podSpec.RestartPolicy = corev1.RestartPolicyNever
@@ -671,7 +671,6 @@ func (w *worker) Build(podSpec *corev1.PodSpec, skipCheckout bool, inputs buildI
 		}
 	}
 
-	cullCheckForExisting(initContainers[0])    // the workspace setup container
 	for _, c := range podSpec.InitContainers { // user-defined init containers
 		cullCheckForExisting(c)
 	}
@@ -726,7 +725,7 @@ func (w *worker) Build(podSpec *corev1.PodSpec, skipCheckout bool, inputs buildI
 			zap.String("image", image),
 			zap.String("policy", string(policy)),
 		)
-		initContainers = append(initContainers, corev1.Container{
+		podSpec.InitContainers = append(podSpec.InitContainers, corev1.Container{
 			Name:            name,
 			Image:           image,
 			ImagePullPolicy: policy,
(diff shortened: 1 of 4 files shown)


internal/controller/scheduler/scheduler.go
```diff
@@ -560,7 +560,7 @@ func (w *worker) Build(podSpec *corev1.PodSpec, skipCheckout bool, inputs buildI
 		)
 	}
 
-	initContainers := []corev1.Container{w.createWorkspaceSetupContainer(podSpec, workspaceVolume)}
+	podSpec.InitContainers = []corev1.Container{w.createWorkspaceSetupContainer(podSpec, workspaceVolume)}
```
DrJosh9000, line 563: If the user has supplied a podSpec that already defines one or more init containers, this would replace them with the workspace setup container.

It would also make the workspace setup container patchable, which would make it easy to break the way the agent is designed to work in the pod. Is that the intent of this PR? What's the use case you have in mind?

DrJosh9000 (commented):
Thanks for raising this PR, @defgenx ! I have a question, see below.

internal/controller/scheduler/scheduler.go
defgenx: Thanks for reviewing and answering !

Yes true, my bad. We can move to an `append" here instead. I forgot about plugins to be honest.

From my understanding, `pod-spec-patch` is an advanced feature allowing anyone to patch the podSpec default values at runtime.
In our case we want to change the resources of the `copy-agent` init container. It looks like it's nothing new because it was possible with the old `pod-spec-patch` implementation; making me think it's more a regression than a feature.
It's also possible to patch the containers of the agent right now so it would make sense to allow patching the init containers too. What do you think about it ?

Let me know if you have anything else in mind, but controlling the cost is really important for us and since we bumped from an old agent version to the last one, we cannot do that anymore so I tried to making it works again. :)

DrJosh9000: I see, thanks for clarifying. I think the init container commands and args should still be protected from modification (with an override in the form of #449), but allowing the resources to be limited makes total sense. 

Since these default containers should require bare minimum resources on pretty much any system, I wonder if they should be generated with default limits - that way you wouldn't have to patch them at all!

defgenx: Yes If I could have a configuration to define the resources by default for the init container `copy-agent` it would do the trick and I think it would be cleaner.
I can try to do that if it's ok for you.

defgenx: @DrJosh9000 any updates?

DrJosh9000: > Yes we can keep the command protected but I feel we should be able to patch the rest (as for containers). I can update the PR so we raise an error if the command is touched. Is it ok for you ?

> Can we keep this PR to make it works like for containers and then add the feature to configure the copy-agent in another PR later?

Both sound good to me.

defgenx: @DrJosh9000 cool ! I let you review my changes then 🙏

defgenx:
@DrJosh9000 looks like a few merged PRs implement fixes to avoid patching init containers. It's still impossible to set resources for the `copy-agent` but at least the default value looks ok.

If you want to close this PR just let me know, but if Buildkite doesn't want users to patch init-containers I think the documentation should be explicit about it. Right now the example is really misleading. :)

defgenx:
@mcncl you want me to close the PR? :)

defgenx:
Just closing this PR, try to achieve what we wanted to do with current fixes done and see if it needs a new PR for that.

PR #490, opened 2025-01-30
enable name overrides

alexr56k:
**Description**

This simple change will allow you to assert a new chart name for the helm installation. It does not introduce any regressions with the normal behavior of the chart.

**Reasoning**

I found a need for this while trying to deploy multiple Buildkite queues with the same parent chart. I essentially made a parent that calls this chart (n) times as a subchart dependency. This allows me to manage multiple queues with wildly different configurations without managing individual charts.

If requested, I'd be happy to provide a readme example for my specific use case.


rayalan (approved):

charts/agent-stack-k8s/templates/_helper.tpl
```diff
@@ -1,6 +1,23 @@
...
+{{- .Values.fullnameOverride | trunc 63 | trimSuffix "-" -}}
+{{- else -}}
+{{- $name := default .Chart.Name .Values.nameOverride -}}
+{{- if contains $name .Release.Name -}}
+{{- .Release.Name | trunc 63 | trimSuffix "-" -}}
+{{- else -}}
+{{- printf "%s-%s" .Release.Name $name | trunc 63 | trimSuffix "-" -}}
+{{- end -}}
+{{- end -}}
+{{- end -}}
```
DrJosh9000, line 16: Since all the template actions use `-` on both sides, you should be able to indent these to make the nested structures clearer without changing what it does:

```suggestion
{{- define "agent-stack-k8s.name" -}}
  {{- default .Chart.Name .Values.nameOverride | trunc 63 | trimSuffix "-" -}}
{{- end -}}

{{- define "agent-stack-k8s.fullname" -}}
  {{- if .Values.fullnameOverride -}}
    {{- .Values.fullnameOverride | trunc 63 | trimSuffix "-" -}}
  {{- else -}}
    {{- $name := default .Chart.Name .Values.nameOverride -}}
    {{- if contains $name .Release.Name -}}
      {{- .Release.Name | trunc 63 | trimSuffix "-" -}}
    {{- else -}}
      {{- printf "%s-%s" .Release.Name $name | trunc 63 | trimSuffix "-" -}}
    {{- end -}}
  {{- end -}}
{{- end -}}
```

DrJosh9000 (changes requested):
Thanks for the contribution! I think this is overall good. 

I think at minimum you will need to define the new `nameOverride` and `fullnameOverride` in `charts/agent-stack-k8s/values.schema.json`, otherwise Helm install may fail to validate `values.yaml`.

alexr56k:
Thanks for the suggestions @DrJosh9000 . I've made the requisite changes to the chart.

charts/agent-stack-k8s/templates/_helper.tpl
alexr56k: Added!

DrJosh9000 (approved):
LGTM! Thanks!

PR #494, opened 2025-02-03
Update CODEOWNERS - adding Support Engineers as CODEOWNERS

karensawrey:
Adding CODEOWNERS file with Support Engineers as codeowners.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,1 +1,2 @@
 * @buildkite/pipelines-enterprise
+* @buildkite/support-engineering


DrJosh9000 (approved):
LGTM

Maybe the defunct pipelines-enterprise team should be cleaned up too?

karensawrey:
Thanks, @DrJosh9000 - I'll merge this.
I was following the instructions in the Linear task that explicitly stated not to remove any existing codeowners. So a clean up of those is up to you.
Thanks again for the review!

PR #495, opened 2025-02-04
Fix resources and port parsing for non-strings

mikekap:
resources can be specified as floats or ints; similarly, liveness checks use a port that's either a string or an int (i.e. named port from the ports: list).

Overall, this hand-coded approach is likely to continue causing issues, but this at least fixes the two issues I ran into. I think to fix this completely, you'd need to rewrite config parsing to use the same machinery as what kube uses, but that feels overkill.

### DIFF
--- a/cmd/controller/controller.go
+++ b/cmd/controller/controller.go
@@ -6,6 +6,7 @@ import (
 	"log"
 	"os"
 	"reflect"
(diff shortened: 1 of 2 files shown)


DrJosh9000 (approved):
Thanks for the contribution! LGTM!

Issue #480, opened 2025-01-23
Controller should always set Job.spec.activeDeadlineSeconds


artem-zinnatullin:
Follow-up for https://github.com/buildkite/agent-stack-k8s/issues/479

I strongly believe that for such a critical dev-infrastructure software as CI/CD system controller all measures should be taken to prevent  customer resource overuse and overspend in all or any of following error scenarios:

- Buildkite API goes partially or fully [down as on Jan 5-7 2025](https://www.buildkitestatus.com/incidents/txxkzf4r262c)
- Controller itself gets into `CrashLoopBackOff` or otherhow loses ability to communicate with K8S API
- One of current 5 containers used in each Job: `copy-agent`, `imagecheck-0`, `checkout`, `agent` or `container-0` gets into unrecoverable error-state such as https://github.com/buildkite/agent/issues/3149
- Etc

K8S itself should be able to eventually kill the stuck CI Jobs from the cluster, otherwise as happened with us https://github.com/buildkite/agent-stack-k8s/issues/479 we had 96 jobs stuck in partial-error state for over a week and we only found out due to billing overspend detection.

K8S Jobs have a fairly robust and simple mechanic: [`Job.spec.activeDeadlineSeconds`](https://kubernetes.io/docs/concepts/workloads/controllers/job/) which if set allows K8S itself to terminate Jobs that exceed active deadline which would have helped us avoid overspend.

Currently, https://github.com/buildkite/agent-stack-k8s only sets `activeDeadlineSeconds` as part of  `cleanupSidecars()` which as far 
I understand happens after controller detects job being done and as described here is an unreliable mechanic: https://github.com/buildkite/agent-stack-k8s/blob/734dfdd78d182a8b4fe16cd7e7bd6dbd5a41d31a/internal/controller/scheduler/completions.go#L97

I think https://github.com/buildkite/agent-stack-k8s must **_always_** set a good default `Job.spec.activeDeadlineSeconds` (say 3-6 hours) and allow users to override it on per Job basis for those customers who have long running jobs or other need to customize this behavior.

Issue #479, opened 2025-01-23
Job keeps running when checkout container fails causing overspend


artem-zinnatullin:
After the recent Buildkite [outage on Jan 5-7 2025](https://github.com/buildkite/agent/issues/3149) we've found a number (96) of CI Jobs created by https://github.com/buildkite/agent-stack-k8s being stuck in the cluster for **_over a week_** in a partially failed state thus causing a noticeable overspend on our AWS.

My investigation into the problem suggests that `agent` container created as part of the K8S CI Job was still running despite `checkout` and `container-0` (user build) containers being failed within the Job. 

Due to `agent` container still running neither the https://github.com/buildkite/agent-stack-k8s, nor https://github.com/lwolf/kube-cleanup-operator considered such Jobs/Pods as finished and didn't kill them thus they kept "running" for over a week allocating significant resources in the cluster.

We also didn't see this on Buildkite side, it was the cluster overspend alarm that raised our attention, there still might be other Buildkite customers with these "failed-but-running" Jobs stuck in their K8S clusters causing overspending and unnecessary resource usage.

`kubectl get pods`:

```js
NAME                                                              READY   STATUS      RESTARTS   AGE
buildkite-019436c4-c39d-49a6-941c-acd9614ffd4a-pj4qk              1/3     Error       0          9d
buildkite-0194370b-f565-4d5a-968f-e797c26a35e6-5m8q8              1/3     Error       0          9d
buildkite-01943769-c3bf-474b-896d-0fbecc6b6249-ttcnh              1/3     Error       0          9d
buildkite-0194376c-9e09-4d23-94f1-6b7871c21101-pc9zg              1/3     Error       0          9d
buildkite-01943d74-2dc1-4c1c-917d-c44bb8551d78-czpmt              1/3     Error       0          7d20h
buildkite-01943d74-2dc8-4f12-8f19-9eb635f0c5d2-fsfbp              1/3     Error       0          7d20h
buildkite-01943d74-b43e-450b-9bd8-eea8dbe9b671-v5rr4              1/3     Error       0          7d20h
buildkite-01943d74-b43f-4679-87bb-798b856d1e90-xttmz              1/3     Error       0          7d20h
buildkite-01943d74-b440-49c1-ae5e-39e774413c32-qkvk4              1/3     Error       0          7d20h
buildkite-01943d78-9f13-4f95-a94e-84bed679c3fe-wnnfm              1/3     Error       0          7d20h
[...]

Issue #463, opened 2024-12-28
Controller Keeps Reattempting Failed Jobs


CH-BrianJurgess:
In our cluster, the agent controller will keep reattempting a number of jobs despite these jobs either being failed or cancelled within the UI. It will recreate these pods over and over again no matter how many times I manually clean them up. When I grab the logs from the agent container I see that the agent container properly exits. However, when I grab the logs from container-0, I see that the buildkite agent has seg faulted and the container never exists. Every single one of these pipelines had eventually failed for one reason or another on the original run. 

**Agent Container Logs**
```
2024-12-28 21:52:34 NOTICE Starting buildkite-agent v3.87.0 with PID: 8
2024-12-28 21:52:34 NOTICE The agent source code can be found here: https://github.com/buildkite/agent
2024-12-28 21:52:34 NOTICE For questions and support, email us at: hello@buildkite.com
2024-12-28 21:52:34 INFO   Configuration loaded path=/home/agent/buildkite/config
2024-12-28 21:52:34 INFO   Build Path doesn't exist, creating it (/workspace/build)
2024-12-28 21:52:34 INFO   Registering agent with Buildkite...
2024-12-28 21:52:34 INFO   Successfully registered agent "buildkite-01940a29-f1c4-44bf-88d2-672719abb132-9f4f2" with tags [k8s:agent-stack-version=v0.20.1, k8s:service-account=buildkite-build-agent, k8s:namespace=jenkins, k8s:node=ip-10-16-70-42.us-west-2.compute.internal, queue=default-queue]
2024-12-28 21:52:34 INFO   Starting 1 Agent(s)
2024-12-28 21:52:34 INFO   You can press Ctrl-C to stop the agents
2024-12-28 21:52:34 INFO   buildkite-01940a29-f1c4-44bf-88d2-672719abb132-9f4f2 Connecting to Buildkite...
2024-12-28 21:52:34 INFO   buildkite-01940a29-f1c4-44bf-88d2-672719abb132-9f4f2 Attempting to acquire job 01940a29-f1c4-44bf-88d2-672719abb132...
[...]

PR #497, opened 2025-02-06
SUP-3258 - Implement .spec.job.activeDeadlineSeconds

petetomasik:
Implements `.spec.job.activeDeadlineSeconds` as a means of [setting runtime boundaries](https://kubernetes.io/docs/concepts/workloads/controllers/job/#job-termination-and-cleanup) for Kubernetes jobs that are unable to complete within a "reasonable" amount of time.

This is configured with a default value of `21600` seconds (6 hours) in the controller configuration. This value can be overridden by the user via the command line `--job-active-deadline-seconds 43200`, directly in the controller configuration `job-active-deadline-seconds: 43200`, or on a per Buildkite job basis via the `kubernetes` plugin:
```
steps:
  - label: "12 hour nap, override default 6 hour job-active-deadline-seconds config"
    command: sleep 43200
    plugins:
      - kubernetes:
          jobActiveDeadlineSeconds: 46800
```
At this time `.spec.job.activeDeadlineSeconds` will not be able to be disabled, as we [agree](https://github.com/buildkite/agent-stack-k8s/issues/480#issue-2807735870) that a job runtime boundary should be defined at all times.

Fixes https://github.com/buildkite/agent-stack-k8s/issues/480
Fixes https://github.com/buildkite/agent-stack-k8s/issues/479
Fixes https://github.com/buildkite/agent-stack-k8s/issues/463

### DIFF
--- a/internal/controller/scheduler/scheduler.go
+++ b/internal/controller/scheduler/scheduler.go
@@ -54,6 +54,7 @@ type Config struct {
 	Image                       string
 	AgentTokenSecretName        string
 	JobTTL                      time.Duration
+	JobActiveDeadlineSeconds    int
 	AdditionalRedactedVars      []string
 	WorkspaceVolume             *corev1.Volume
 	AgentConfig                 *config.AgentConfig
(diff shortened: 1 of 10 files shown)


internal/controller/scheduler/scheduler.go
```diff
@@ -75,6 +76,7 @@ func New(logger *zap.Logger, client kubernetes.Interface, cfg Config) *worker {
 }
 
 type KubernetesPlugin struct {
+	Job               *batchv1.Job           `json:"job,omitempty"`
```
DrJosh9000, line 79: Using `*batchv1.Job` here could get confusing - someone might try to write:

```yaml
plugins:
- kubernetes:
    job:
      template:
        spec:
          containers:
            ...
```

which wouldn't be rejected for any failure to parse, and the user might expect that defining containers that way would work but instead it would be silently ignored. 

(Maybe one day the whole `KubernetesPlugin` should be changed to revolve around a `*batchv1.Job` instead of defining pieces of it. Needs thought.)

In the meantime I would suggest either just adding `JobActiveDeadlineSeconds` as a `KubernetesPlugin` field directly, or maybe wrapping it in a new `Job` struct that can be grown over time to include other job parameters.

DrJosh9000 (approved):
LGTM

Other than one comment below, ideally this would also have a test. The easiest way currently is probably an integration test that uses a pipeline YAML fixture with a small `activeDeadlineSeconds` and a command that sleeps longer than it. But it may be tricky to take image pull time into account, so the deadline and sleep would probably be at least 30 seconds to avoid flakiness.

internal/controller/scheduler/scheduler.go
petetomasik: This is step 1 of a long journey to fully transition to `*batchv1.Job`. I went back and forth on whether to implement it this way and possibly cause confusion for users that try to leverage the entire Job spec. There are additional tickets we have that will continue to build off of the Job spec (`backoffLimit`, `podFailurePolicy`, etc.) bit by bit. We could implement this (and the future additional fields) as fields in the plugin and eventually bring them back into the full Job spec when the time comes to more precisely control the feature set.

DrJosh9000 (approved):
LGTM ✅

PR #499, opened 2025-02-07
SUP-3283: Enable configuring GraphQL Result Limits

tomowatt:
By allowing configuring the GraphQL Result Limits the amount of Jobs being retrieived by queries can be increased/decreased for scaling the amount of Jobs the controller can retreive and then scheduled as Kubernetes Jobs

### DIFF
--- a/README.md
+++ b/README.md
@@ -122,24 +122,33 @@ Available Commands:
   version     Prints the version
 
 Flags:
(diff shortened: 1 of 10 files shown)


petetomasik (approved):
Tested locally with `graphql-results-limit: 1` and everything worked as expected! 🚀

PR #500, opened 2025-02-07
RFC: Use different exit statuses for different failures

mikekap:
The problem that I'm having is that I get occasional image pull "failures." That is to say - these are not persistent, retrying them usually fixes things.

I'd like to automate the retries but the current retry code makes that difficult (1 is too generic). `docker run` returns 125 for a missing image. This patch makes buildkite return the same and similarly updates exit codes in other places. This isn't obviously the right approach, so feedback welcome.

### DIFF
--- a/internal/controller/scheduler/fail_job.go
+++ b/internal/controller/scheduler/fail_job.go
@@ -26,6 +26,7 @@ func acquireAndFailForObject(
 	k8sClient kubernetes.Interface,
 	cfg *config.Config,
 	obj metav1.Object,
+	code int,
 	message string,
 ) error {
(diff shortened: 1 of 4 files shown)


zhming0:
@mikekap thank you for raising this PR to use. We adopted your idea [in another PR](https://github.com/buildkite/agent-stack-k8s/pull/628), which I think mostly does the same time and achieve the same goal. 

Along with our other recent changes in [v0.29.0](https://github.com/buildkite/agent-stack-k8s/releases/tag/v0.29.0), like the introduction of `stack_error` error reason, hopefully the error handling is slightly easier now 🙏🏿 . 

The only exception is exit code 125 for image pull error, we didn't make this change because there are more error categories in image pull failure like authentication error etc. We welcome contribution in making a more granular error classification system there 🙌🏿 .

PR #502, opened 2025-02-10
SUP-3216 - Remove checkout container from podSpec, if skipped

petetomasik:
If a Buildkite job is configured to skip checkout AND the `checkout` container is configured in the controller's `pod-spec-patch` or the plugin's `podSpecPatch`, remove the `checkout` container from the resulting merged `podSpec` before the controller creates the Kubernetes Job.

### DIFF
--- a/internal/controller/scheduler/scheduler.go
+++ b/internal/controller/scheduler/scheduler.go
@@ -586,11 +586,14 @@ func (w *worker) Build(podSpec *corev1.PodSpec, skipCheckout bool, inputs buildI
 		w.logger.Debug("Applied podSpec patch from k8s plugin", zap.Any("patched", patched))
 	}
 
+	// Removes all containers named "checkout" when checkout disabled via controller config or plugin
+	// This will also remove containers named "checkout" added via PodSpecPatch
 	if skipCheckout {
 		for i, pod := range podSpec.Containers {
 			switch pod.Name {
 			case CheckoutContainerName:
-				podSpec.Containers = append(podSpec.Containers[:i], podSpec.Containers[i+1:]...)
+				podSpec.Containers = slices.Delete(podSpec.Containers, i, i+1)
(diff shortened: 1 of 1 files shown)


internal/controller/scheduler/scheduler.go
```diff
@@ -586,6 +586,18 @@ func (w *worker) Build(podSpec *corev1.PodSpec, skipCheckout bool, inputs buildI
 		w.logger.Debug("Applied podSpec patch from k8s plugin", zap.Any("patched", patched))
 	}
 
+	if skipCheckout {
+		for i, pod := range podSpec.Containers {
+			switch pod.Name {
+			case CheckoutContainerName:
+				podSpec.Containers = append(podSpec.Containers[:i], podSpec.Containers[i+1:]...)
```
DrJosh9000, line 593: [slices.Delete](https://pkg.go.dev/slices#Delete) would be more readable here:

```suggestion
				podSpec.Containers = slices.Delete(podSpec.Containers, i, i+1)
```

internal/controller/scheduler/scheduler.go
```diff
@@ -586,6 +586,18 @@ func (w *worker) Build(podSpec *corev1.PodSpec, skipCheckout bool, inputs buildI
 		w.logger.Debug("Applied podSpec patch from k8s plugin", zap.Any("patched", patched))
 	}
 
+	if skipCheckout {
```
DrJosh9000, line 589: Given the complexity in this file I think there ought to be a comment added that explains what's going on:

```suggestion
	// PodSpecPatch can be used to add a container named "checkout" even when
	// checkout has been disabled, so this removes it.
	if skipCheckout {
```

DrJosh9000 (commented):
Is it worth producing a log line when this happens? A customer that has used podSpecPatch to add a container named "checkout" might be confused if it mysteriously disappears.

internal/controller/scheduler/scheduler.go
petetomasik: Thanks for the feedback. Changes implemented!

tomowatt (approved):

PR #507, opened 2025-02-12
[SUP-3361] Controller to handle queue paused events and prevent processing jobs

nsuma8989:
This PR is for controller to check if the queue it is picking up the jobs for is paused or not. If the queue is paused then controller should not pick up the jobs and process them. It should wait until the queue is unpaused and then process the jobs

### DIFF
--- a/internal/controller/monitor/monitor.go
+++ b/internal/controller/monitor/monitor.go
@@ -41,6 +41,7 @@ type Config struct {
 	Org                    string
 	Tags                   []string
 	GraphQLResultsLimit    int
+	EnableQueuePause       bool
 }
 
 func New(logger *zap.Logger, k8s kubernetes.Interface, cfg Config) (*Monitor, error) {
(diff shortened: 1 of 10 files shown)


petetomasik:
We should include information about this specific new behavior in the README. The ability to pause/resume queues is also only [available](https://buildkite.com/docs/pipelines/clusters/manage-queues#pause-and-resume-a-queue) on the latest plans.

nsuma8989:
> We should include information about this specific new behavior in the README. The ability to pause/resume queues is also only [available](https://buildkite.com/docs/pipelines/clusters/manage-queues#pause-and-resume-a-queue) on the latest plans.

@petetomasik I was planning to do that after the fix is merged and release is made so in the readme I can mention a note specifying that from release X of agent-stack-k8s if queue is paused controller will respect that and stop picking up jobs until queue is unpaused.

internal/controller/monitor/monitor.go
```diff
@@ -126,6 +126,23 @@ func (m *Monitor) getScheduledCommandJobs(ctx context.Context, queue string) (jo
...
+	isQueuePaused := false
+	for _, edge := range queues.Organization.Cluster.Queues.Edges {
+		if edge.Node.Key == queue {
+			isQueuePaused = edge.Node.DispatchPaused
+			break
+		}
+	}
+
+	if isQueuePaused {
+		return nil, fmt.Errorf("the queue '%s' is paused", queue)
```
DrJosh9000, line 143: `%q` is useful for wrapping strings in quotes for you:

```suggestion
		return nil, fmt.Errorf("the queue %q is paused", queue)
```

internal/controller/monitor/monitor.go
```diff
@@ -126,6 +126,23 @@ func (m *Monitor) getScheduledCommandJobs(ctx context.Context, queue string) (jo
 		agentQueryRule = append(agentQueryRule, fmt.Sprintf("queue=%s", queue))
 	}
 
+	queues, err := api.GetClusterQueues(ctx, m.gql, m.cfg.Org, m.cfg.ClusterUUID, 100)
```
DrJosh9000, line 129: I think it may be worth a TODO comment to change this approach once there is a way to get info for a particular queue in a targeted query

```suggestion
	// TODO: use a more targeted query once one becomes available
	queues, err := api.GetClusterQueues(ctx, m.gql, m.cfg.Org, m.cfg.ClusterUUID, 100)
```

DrJosh9000 (approved):
LGTM with two small suggestions

petetomasik (changes requested):
In order to prevent unexpected behavior with the controller, this feature should be disabled by default in the config and can be enabled by those that would like to use it.

nsuma8989:
> In order to prevent unexpected behavior with the controller, this feature should be disabled by default in the config and can be enabled by those that would like to use it.

@petetomasik Why would this be unexpected behavior? If the queue is paused, the user did it on purpose to stop processing new jobs, but with agent-stack-k8s today, it does not respect that. So I would say we are fixing a gap in our solution. I thought we call this out in the release notes and also in the readme documentation instead of making it disabled by default.

petetomasik (approved):

PR #508, opened 2025-02-13
Fix unusable image pull policy options

### LINKED ISSUE #486: ImageCheckPullPolicy gets overriden by ImagePullPolicy
The recent changes of image pull policies in https://github.com/buildkite/agent-stack-k8s/pull/459 created an issue where the CheckPullPolicy isn't correctly computed.

Looking at this code: https://github.com/buildkite/agent-stack-k8s/blob/107b1cfdbcae9f5a097e93468325f238aa31ddbc/internal/controller/scheduler/scheduler.go#L708-L721

When no DefaultImageCheckPullPolicy is set, correct defaults are applied based on the image ref (`Always` if it doesn't have a digest, otherwise `IfNotPresent`). It seems however that these defaults never get applied when the policy of the container is already set to something, here: https://github.com/buildkite/agent-stack-k8s/blob/107b1cfdbcae9f5a097e93468325f238aa31ddbc/internal/controller/scheduler/scheduler.go#L410-L412

So if there is a DefaultImagePullPolicy set or the container spec has an explicit policy, that will always take precedence over the CheckPolicy, either set by DefaultImageCheckPullPolicy or calculated by the logic mentioned above.

When now using the default config https://github.com/buildkite/agent-stack-k8s/blob/107b1cfdbcae9f5a097e93468325f238aa31ddbc/cmd/controller/controller.go#L118-L127
that sets `default-image-pull-policy: "IfNotPresent"` and `"default-image-check-pull-policy": ""`
we never get the desired behavior of always pulling images with no digest.

It seems like the only way to get the pull logic based on the digest being present is to unset the defaultImagePullPolicy. I assume then we'd fall back to the defaults the K8s sets: https://github.com/kubernetes/api/blob/b849e76aab64dcc5d2d4947cccccebdf0d8ec3c0/core/v1/types.go#L2912

This would be reasonable in a standalone scenario, but would pull twice again with your check logic. 

It seems like there's no combination of settings that would achieve correctness and pull deduplication as originally indented by https://github.com/buildkite/agent-stack-k8s/pull/459

DrJosh9000:
Fixes #486

### DIFF
--- a/charts/agent-stack-k8s/values.schema.json
+++ b/charts/agent-stack-k8s/values.schema.json
@@ -520,7 +520,7 @@
         "default-image-pull-policy": {
           "type": "string",
           "description": "Configures a default image pull policy for containers that do not specify a pull policy, or containers created by the stack itself",
-          "default": "IfNotPresent",
+          "default": "",
           "examples": ["Always", "IfNotPresent", "Never", ""]
         },
         "default-image-check-pull-policy": {
--- a/cmd/controller/controller.go
+++ b/cmd/controller/controller.go
@@ -129,7 +129,7 @@ func AddConfigFlags(cmd *cobra.Command) {
 	)
 	cmd.Flags().String(
 		"default-image-pull-policy",
-		"IfNotPresent",
+		"",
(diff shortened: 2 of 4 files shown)


petetomasik (approved):
✅

Issue #92, opened 2023-01-25
Use pagination with polling for jobs


benmoss:
Right now we only look at the first 100 scheduled builds, which is a pretty good simple approach, since as the builds are started we should see new ones appear on the next time we poll. It seems like it might be more performant though to iterate through the pages instead of waiting for the next loop in order to schedule all possible builds.

PR #515, opened 2025-02-17
SUP-3347: Enable Pagination of Scheduled Jobs

tomowatt:
By enabling Pagination of the Scheduled Jobs, more Jobs can be retrieved and scheduled at once on Kubernetes.
This reduced the need for the _controller_ to poll multiple times and potentially getting duplicate Jobs that are being scheduled already, but does not result in the less amount of Queries.
Set a same maximum depth of 20, in combination of a maximum of 500 Jobs per query, can result in trying to schedule 10k jobs on a kubernetes cluster

TODO:
- [x] update GetScheduledJobs usage to also paginate
- [ ] refactor pagination loops for retrieving both Unclustered and Clusterd Jobs (maybe?)
- [x] update example config
- [x] update test 
- [x] update README

Fixes https://github.com/buildkite/agent-stack-k8s/issues/92


tomowatt:
Notes on the changes to the `order` used into order to support pagination:

We use `RECENTLY_ASSIGNED` to get the oldest Job first but `RECENTLY_CREATED` will return the newest Job first. The former is FIFO, the latter is LIFO but changes the shape of GraphQL response.

using `RECENTLY_ASSIGNED` will never work with pagination as we would be trying to paginate up a stack, which explains the error received when using the `endCursor` with `RECENTLY_ASSIGNED`:
```
cursor is invalid. Ensure the cursor you passed came from a `cursor` field sourced from an `edge` node in the Buildkite GraphQL API
```

Why it doesn't work is because we would never know what's higher up the stack, so the cursor values are different and cannot be used to move the next position and check what's higher up e.g., if it's the end of the stack of Jobs; else could endlessly paginate. 

The change of ordering also affects the `startCursor` value as well as `endCursor. Thought I was going crazy overlooking something in the changes but when checking the GraphQL Console and checked the cursor values for the same Job using both `RECENTLY_CREATED` and `RECENTLY_ASSIGNED` we can see see the different values e.g.,
```
RECENTLY_CREATED = ktkhMjAyNS0wMi0yNyAwODoxMTo0NC45NDc2MzQwMDAgVVRD2SQwMTk1NDY3Ni0yZGM4LTQ5NzctOTdjZS05NGQ1Nzc2Nzk3MmI=
RECENTLY_ASSIGNED = ksDZJDAxOTU0Njc2LTJkYzgtNDk3Ny05N2NlLTk0ZDU3NzY3OTcyYg==
```

tomowatt:
@petetomasik should we replace use of `RECENTLY_ASSIGNED` for `RECENTLY_CREATED` and simplify this change? It will bring in a significant change overall but might not be as noticable depending on the scale of Jobs

petetomasik:
> @petetomasik should we replace use of `RECENTLY_ASSIGNED` for `RECENTLY_CREATED` and simplify this change? It will bring in a significant change overall but might not be as noticable depending on the scale of Jobs

Switching to `RECENTLY_CREATED` would greatly simply things. It would mean that pagination is always on, no matter the combination of `graphql-results-limit` and `pagination-depth-limit` config values. When `HasNextPage` is `nil`, the loop can break. This might happen after a single call or `20` (max limit of `pagination-depth-limit`). It would also mean processing a maximum of `10,000` jobs with `SCHEDULED` state in 20 API calls in ~20 seconds (default `poll-interval` is `1s`).

Additionally, we could revise our strategy and GQL queries to work from the oldest jobs to the newest jobs by using:
* `last` instead of `first` and `before` instead of `after` in the `jobs` field of the `organization()` queries
* `startCursor` instead of `endCursor` and `hasPreviousPage` instead of `hasNextPage` in the `PageInfo` object
This would ensure jobs are processed in order of creation and allow your logic with `pagination-depth-limit` to continue working to limit the total number of results returned.

petetomasik (approved):
🚀

Issue #166, opened 2023-05-02
Configurable job/pod name


calvinbui:
All job and pod names are configured to be `buildkite-%s`

https://github.com/buildkite/agent-stack-k8s/blob/676327839cb9f5bab6d4c1838fe0cd44d18333ac/internal/controller/config/config.go#L37)

https://github.com/buildkite/agent-stack-k8s/blob/676327839cb9f5bab6d4c1838fe0cd44d18333ac/internal/controller/scheduler/scheduler.go#L459

We have multiple environments, so it'll be easier to identify where jobs are run if this value was configurable.

![image](https://user-images.githubusercontent.com/3604363/235558380-3bd4a199-f576-46ca-8a8f-b2aef906f054.png)

triarius:
Hi @calvinbui, thanks for using agent-stack-k8s! This is certainly something we could implement, but there are a few choices for what to put there.

Some that I can think of are:
- The helm release name (`agent-stack-k8s` in the examples we give)
- The k8s namespace (`buildkite` in the examples we give)
- Some new config value in the agent-stack-k8s config, exposed in the helm chart (not as big a fan of this, by I could be convinced)
- the step label or key normalised in some way
- the hostname of the node that the pod ran on

It also should be noted that knowing where the jobs are run is a few clicks away in the job's timeline tab. There you can find a link to the agent, who's tags will make it clear it was run on Kubernetes (and some of the option for the choice above). But I completely agree that having it in the place in your screenshot is a lot more user-friendly.

Let me know what you think.

calvinbui:
> Hi @calvinbui, thanks for using agent-stack-k8s! This is certainly something we could implement, but there are a few choices for what to put there.
> 
> Some that I can think of are:
> 
> * The helm release name (`agent-stack-k8s` in the examples we give)
> * The k8s namespace (`buildkite` in the examples we give)
> * Some new config value in the agent-stack-k8s config, exposed in the helm chart (not as big a fan of this, by I could be convinced)
> * the step label or key normalised in some way
> * the hostname of the node that the pod ran on
> 
> It also should be noted that knowing where the jobs are run is a few clicks away in the job's timeline tab. There you can find a link to the agent, who's tags will make it clear it was run on Kubernetes (and some of the option for the choice above). But I completely agree that having it in the place in your screenshot is a lot more user-friendly.
> 
> Let me know what you think.

[...]

PR #526, opened 2025-02-25
Allow configurable job name prefix

petetomasik:
Allows configuration of the prefix used when generating Kubernetes job names. Previously, this prefix was hardcoded to `buildkite-`. The existing structure for generating Kubernetes job names remains the same, at `<job-prefix><buildkite-job-uuid>`. If not provided, configuration will default to `buildkite-`. This prefix will also be reflected in the UI as used for the agent's `Name` and `Hostname`.

Fixes https://github.com/buildkite/agent-stack-k8s/issues/166


tomowatt (approved):

nsuma8989 (approved):
Looks fine to me

PR #530, opened 2025-02-27
make it clear, clusters are required

pbuckley:


README.md
```diff
@@ -103,7 +103,7 @@ helm upgrade --install agent-stack-k8s oci://ghcr.io/buildkite/helm/agent-stack-
     --set graphqlToken=<your Buildkite GraphQL-enabled API token>
 ```
 
-If you are using [Buildkite Clusters](https://buildkite.com/docs/agent/clusters) to isolate sets of pipelines from each other, you will need to specify the cluster's UUID in the configuration for the controller. This may be done using a flag on the `helm` command like so: `--set config.cluster-uuid=<your cluster's UUID>`, or an entry in a values file.
+Since [Buildkite Clusters](https://buildkite.com/docs/agent/clusters) was released as generally available in 2024, it's required to specify your cluster's UUID in the configuration for the controller when you deploy with helm. This is done by using a flag on the `helm` command like so: `--set config.cluster-uuid=<your cluster's UUID>`, or an entry in a values file.
```
cnunciato, line 106:
```suggestion
With the introduction of [Buildkite Clusters](https://buildkite.com/docs/agent/clusters) 2024, it's now required to specify your cluster's UUID in the configuration for the controller when you deploy with Helm. You can do this by setting a flag on the `helm` command like so: `--set config.cluster-uuid=<your cluster's UUID>`, or an entry in a values file.
```

cnunciato: ☝🏻 Teeny tweak that makes this slightly clearer (for me at least). 😄

karensawrey (approved):

petetomasik:
Closing in favor of https://github.com/buildkite/agent-stack-k8s/pull/531 that has been merged in.

PR #533, opened 2025-03-04
SUP-3316: Add debug logging for stale jobs due to `max-in-flight` capacity

petetomasik:
* Replace `"job data stale"` error with `"stale-job-data-timeout"` to better indicate why jobs aren't able to be created when the number of running jobs is at `max-in-flight` capacity
* Add debug logging for when `staleCtx` channel is closed, signaling `stale-job-data-timeout` has elapsed
* Simplify testing for `jobActiveDeadlineSeconds`


tomowatt (approved):

nsuma8989 (approved):
Looks good to me

PR #534, opened 2025-03-05
SUP-3228: Allow `extraVolumeMounts` for default params

petetomasik:
Allows for `extraVolumeMounts` to be configured for all checkout/command/sidecar containers.
Example:
```yaml
config:
  default-checkout-params:
    gitCredentialsSecret:
      secretName: my-git-credentials
    extraVolumeMounts:
      - name: checkout-extra-dir
        mountPath: /extra-checkout
  default-command-params:
    extraVolumeMounts:
      - name: command-extra-dir
        mountPath: /extra-command
  default-sidecar-params:
    extraVolumeMounts:
      - name: sidecar-extra-dir
        mountPath: /extra-sidecar
  pod-spec-patch:
    containers:
      - name: checkout
        image: "buildkite/agent:latest"
    volumes:
      - name: checkout-extra-dir
        hostPath:
          path: /my/extra/dir/checkout
          type: DirectoryOrCreate
      - name: command-extra-dir
        hostPath:
          path: /my/extra/dir/command
          type: DirectoryOrCreate
      - name: sidecar-extra-dir
        hostPath:
          path: /my/extra/dir/sidecar
          type: DirectoryOrCreate
```
The `extraVolumeMounts` field of the `kubernetes` plugin is still available for use, which applies to _all_ containers. The addition of `extraVolumeMounts` under `default-checkout-params`, `default-command-params` and `default-sidecar-params` allows for a more scoped approach to Volumes.

Also included is a small bug fix for ensuring purposely failed jobs are no longer tracked as "in-flight" by the controller.

Fixes https://github.com/buildkite/agent-stack-k8s/issues/327

PR #535, opened 2025-03-06
SUP-3496: Ensure token is returned to bucket when BK jobs are acquired and failed

petetomasik:
While performing testing for https://github.com/buildkite/agent-stack-k8s/pull/534, I uncovered a scenario where `tryReturnToken()` was not being called when `failJob()` was invoked to acquire and fail a job. This created a situation where there were no more `available-tokens` and `num-in-flight` would be capped out at `max-in-flight`. No Kubernetes jobs would be running and the controller would be unable to create any new Kubernetes jobs. Restarting the controller was the only way to resolve this situation, as the token bucket is initialized (filled) with `max-in-flight` tokens at `Start()`.

Fixes https://github.com/buildkite/agent-stack-k8s/issues/302


karensawrey (approved):
LGTM!

PR #544, opened 2025-03-17
Dedupe extraVolumeMounts

bpoland:
- if a volumeMount is specified in the pipeline yaml already, don't add a duplicate which would cause a validation error
- this allows easier migration to use the functionality added in https://github.com/buildkite/agent-stack-k8s/pull/534

Note that I am not much of a go developer so please review carefully! Also not sure if a test should be added for this case?


petetomasik:
👋  Hi @bpoland. This makes sense. Let me revise this PR with some tests and get the changes merged in.

bpoland:
Replaced by https://github.com/buildkite/agent-stack-k8s/pull/545

Thanks for the quick look! 🎉

PR #545, opened 2025-03-17
Dedupe extraVolumeMounts for checkout, command, sidecar params

petetomasik:
Inspired by https://github.com/buildkite/agent-stack-k8s/pull/544, dedupes `extraVolumeMounts` from both controller config and `kubernetes` plugin against the `VolumeMounts` of the `corev1.Container` being created for each of the `checkout`, `command` and `sidecar` containers.


nsuma8989 (approved):
Looks good to me

PR #556, opened 2025-03-31
fix: migrate golangci-lint to v2

scadu:
Migrate golangci-lint over to v2 configuration.
Remove `ObjectMeta` as `Annotations` and `Labels` are directly available as `jobs.Items[0].Annotations[k]` now.


mcncl:
Can we get a summary of the changes? I notice a couple of deletes, so I assume the new version assumes children at a location and you don't need to declare the Object?

scadu:
> Can we get a summary of the changes? I notice a couple of deletes, so I assume the new version assumes children at a location and you don't need to declare the Object?

`Annotations` and `Labels` are directly available as `jobs.Items[0].Annotations[k]` now. 
Updated PR description.

petetomasik (approved):

PR #559, opened 2025-04-10
Fix the max-in-flight default value in schema

nsuma8989:
Incorrect default value is defined in values.schema as 0. Fixed it to reflect the correct value which is 25


petetomasik (approved):
🚀

PR #560, opened 2025-04-11
Skip jobs that are missing `queue` tag, enhance `info` logs

petetomasik:
If a controller is configured with a tag for `queue=default`, our API will also return jobs that are missing a `queue` tag (or missing tags altogether). This results in missing Labels on the created Kubernetes Jobs and will prevent the controller from being notified of Job/Pod events and subsequently cause `available-tokens` to eventually drop to `0` when `max-in-flight > 0`. When this happens no new jobs are able to be processed by the controller until it is restarted.

Example of Job with correct Labels:
```
Labels:           batch.kubernetes.io/controller-uid=bfbcbc64-d681-494f-b52f-70b498ceaf8d
                  batch.kubernetes.io/job-name=buildkite-019615d0-0af4-4719-99d3-eeb49ca5f4e4
                  buildkite.com/job-uuid=019615d0-0af4-4719-99d3-eeb49ca5f4e4
                  controller-uid=bfbcbc64-d681-494f-b52f-70b498ceaf8d
                  job-name=buildkite-019615d0-0af4-4719-99d3-eeb49ca5f4e4
                  tag.buildkite.com/queue=kubernetes
```
And when job wasn't correctly skipped:
```
Labels:                      buildkite.com/job-uuid=019620cd-3bcb-4fef-9618-b1f6c8b5e9db
```

The practice of skipping jobs should have already been occurring, but the condition of a `queue` tag missing from a job (or no job tags) allowed jobs to be processed even though the controller tags didn't match. This has been corrected. When a job is missing a `queue` tag, the following `info` log will be emitted:
```
job missing 'queue' tag, skipping...    {"org": "my-org", "job-uuid": "01962542-68b8-4970-bd9c-c17b1e23c7f3", "buildkite-job-tags": {"foo":"bar","hello":"world"}}
```

Additionally, some previous `debug` logs have been changed to `info` to provide better feedback around when/if jobs are being processed via GraphQL queries and when jobs are skipped due to job tags not matching:
```
retrieving scheduled jobs via GraphQL API...    {"poll-interval": "1s", "queue-tag": "kubernetes"}
job processing of GraphQL API results completed...      {"graphql-jobs-processed": 2, "queue-tag": "kubernetes"}
```

### DIFF
--- a/internal/controller/agenttags/tags.go
+++ b/internal/controller/agenttags/tags.go
@@ -75,6 +75,7 @@ func LabelsFromTags(tags []string) (map[string]string, []error) {
 // set of strings defined by the tag value in `jobTags` (eg a glob or regex)
 // See https://buildkite.com/docs/agent/v3/cli-start#agent-targeting
 func JobTagsMatchAgentTags(jobTags iter.Seq2[string, string], agentTags map[string]string) bool {
+	queueTagFound := false
 	for k, v := range jobTags {
 		agentTagValue, exists := agentTags[k]
 		if !exists {
@@ -83,8 +84,11 @@ func JobTagsMatchAgentTags(jobTags iter.Seq2[string, string], agentTags map[stri
 		if v != "*" && v != agentTagValue {
 			return false
 		}
+		if k == "queue" {
+			queueTagFound = true
+		}
 	}
-	return true
+	return queueTagFound
 }
 
 // ScanLabels returns an iterator over all labels that are tags.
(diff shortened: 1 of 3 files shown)


petetomasik:
> Your explanation 👏 LGTM
> 
> For my own curiosity, what's the logic behind the change from `Debug` to `Info`?

It's been mentioned that without enabling `debug` mode, the logs of the controller look like no work is being performed and users question if they properly configured their deployment. Moving a few of these logs to `info` now provides some details from the controller that jobs are being processed or, in the case of hitting `max-in-flight`, a very common reason that jobs are not being processed.

PR #562, opened 2025-04-14
feat: New APIs used to fetch jobs

DrJosh9000:


CerealBoy (approved):
LGTM 🚀 

I love the breakdown in the description, it really helps to frame the various changes coming in. There will be some potential teething hiccups as upgrades happen, nothing that isn't explained here though and we can help to transition with. I think this is great!

mcncl (approved):

examples/config.yaml
```diff
@@ -17,9 +16,10 @@ namespace: my-buildkite-ns
 org: my-buildkite-org
 default-image-pull-policy: Never
 default-image-check-pull-policy: IfNotPresent
-graphql-results-limit: 200
 enable-queue-pause: true
+pagination-page-size: 1000
 pagination-depth-limit: 5
+query-reset-interval: 10s
```
dabarrell, line 22: @DrJosh9000 @CerealBoy  Noting that the default value for this is 5s - should we update this example config to reflect the default?

DrJosh9000: The example config is a bit more of a "config zoo" used for the config parser test than a proper example that customers should follow. Maybe there should be a more realistic example config file.

PR #570, opened 2025-04-23
Update config defaults + docs

swaller-bk:
Update configuration defaults after switching out graphQL for REST for some actions. 

Also, a slight reorder of the flags table in docs as the order didn't match the output of our help command 🙃 
help output now looks like this:

```
Flags:
      --agent-token-secret string                          name of the Buildkite agent token secret (default "buildkite-agent-token")
      --allow-pod-spec-patch-unsafe-command-modification   Permits PodSpecPatch to modify the command or args fields of stack-provided containers. See the warning in the README before enabling this option
      --buildkite-token string                             Buildkite API token with GraphQL scopes
      --cluster-uuid string                                UUID of the Buildkite Cluster. The agent token must be for the Buildkite Cluster.
  -f, --config string                                      config file path
      --debug                                              debug logs
      --default-image-check-pull-policy string             Sets a default PullPolicy for image-check init containers, used if an image pull policy is not set for the corresponding container in a podSpec or podSpecPatch
      --default-image-pull-policy string                   Configures a default image pull policy for containers that do not specify a pull policy and non-init containers created by the stack itself
      --empty-job-grace-period duration                    Duration after starting a Kubernetes job that the controller will wait before considering failing the job due to a missing pod (e.g. when the podSpec specifies a missing service account) (default 30s)
      --enable-queue-pause                                 Allow controller to pause processing the jobs when queue is paused on Buildkite
      --graphql-endpoint string                            Buildkite GraphQL endpoint URL
  -h, --help                                               help for agent-stack-k8s
      --image string                                       The image to use for the Buildkite agent (default "ghcr.io/buildkite/agent:3.96.0")
      --image-pull-backoff-grace-period duration           Duration after starting a pod that the controller will wait before considering cancelling a job due to ImagePullBackOff (e.g. when the podSpec specifies container images that cannot be pulled) (default 30s)
      --job-active-deadline-seconds int                    maximum number of seconds a kubernetes job is allowed to run before terminating all pods and failing job (default 21600)
      --job-cancel-checker-poll-interval duration          Controls the interval between job state queries while a pod is still Pending (default 5s)
      --job-creation-concurrency int                       Number of concurrent goroutines to run for converting Buildkite jobs into Kubernetes jobs (default 25)
      --job-prefix string                                  The prefix to use when creating Kubernetes job names (default "buildkite-")
      --job-ttl duration                                   time to retain kubernetes jobs after completion (default 10m0s)
      --k8s-client-rate-limiter-burst int                  The burst value of the K8s client rate limiter. (default 20)
      --k8s-client-rate-limiter-qps int                    The QPS value of the K8s client rate limiter. (default 10)
      --max-in-flight int                                  max jobs in flight, 0 means no max (default 25)
      --namespace string                                   kubernetes namespace to create resources in (default "default")
      --org string                                         Buildkite organization name to watch
      --pagination-depth-limit int                         Sets the maximum number of pages when retreiving Buildkite Jobs to be Scheduled. Increasing this value will increase the number of requests made to the Buildkite API and number of Jobs to be scheduled on the Kubernetes Cluster. (default 2)
      --pagination-page-size int                           Sets the maximum number of Jobs per page when retreiving Buildkite Jobs to be Scheduled. (default 1000)
      --poll-interval duration                             time to wait between polling for new jobs (minimum 1s); note that increasing this causes jobs to be slower to start (default 1s)
      --profiler-address string                            Bind address to expose the pprof profiler (e.g. localhost:6060)
      --prohibit-kubernetes-plugin                         Causes the controller to prohibit the kubernetes plugin specified within jobs (pipeline YAML) - enabling this causes jobs with a kubernetes plugin to fail, preventing the pipeline YAML from having any influence over the podSpec
      --prometheus-port uint16                             Bind port to expose Prometheus /metrics; 0 disables it
      --query-reset-interval duration                      Controls the interval between pagination cursor resets. Increasing this value will increase the number of jobs to be scheduled but also delay picking up any jobs that were missed from the start of the query. (default 10s)
      --tags strings                                       A comma-separated list of agent tags. The "queue" tag must be unique (e.g. "queue=kubernetes,os=linux") (default [queue=kubernetes])
```

https://linear.app/buildkite/issue/PIPE-960/set-sensible-defaults-for-new-params


docs/controller_configuration.md
```diff
@@ -17,40 +17,41 @@
...
 | `--org string`                                 | Buildkite organization name to watch                                                                                                                                                                                                                                                                                          |
-| `--pagination-depth-limit int`                 | Sets the maximum depth of pagination when retrieving Buildkite Jobs to be Scheduled. Increasing this value will increase the number of requests made to the Buildkite GraphQL API and number of Jobs to be scheduled on the Kubernetes Cluster. (default 1)                                                                   |
+| `--pagination-depth-limit int`                 | Sets the maximum number of pages when retreiving Buildkite Jobs to be Scheduled. Increasing this value will increase the number of requests made to the Buildkite API and number of Jobs to be scheduled on the Kubernetes Cluster. (default 2)                                                                               |
```
dabarrell, line 43:
```suggestion
| `--pagination-depth-limit int`                 | Sets the maximum number of pages when retrieving Buildkite Jobs to be Scheduled. Increasing this value will increase the number of requests made to the Buildkite API and number of Jobs to be scheduled on the Kubernetes Cluster. (default 2)                                                                               |
```

CerealBoy (approved):
LGTM 🚀

docs/controller_configuration.md
swaller-bk: ahh damn, I think that is also wrong in source, let me check...

dabarrell (approved):

docs/controller_configuration.md
swaller-bk: aaaaand I also missed out on defaults in the helm `values.schema.json` 🤦‍♀️🤦‍♀️🤦‍♀️

PR #572, opened 2025-04-28
cleanup: Deprecate GraphQL token

DrJosh9000:
### What

1. Remove job cancellation from the pod watcher.
2. Mark the GraphQL token and endpoint options as deprecated.
3. Hide the remaining GraphQL bits inside the integration tests package.
4. Delete unused queries and mutations, and regenerate the stubs.

### Why

Job cancellation is probably safe to remove. There were two cases where the controller tried to cancel a job:

- An agent already started the job (the controller learns this by querying the job state)
- As a fallback when directly acquiring the job in order to fail it didn't succeed

Both of these can only happen if the pod got past initialisation, when the image-check containers run, but for some reason can't re-pull the same image again. So it's already a fairly rare case.

In the first case, we can now rely on the agent's "client lost" timeout to fail the job if one of the expected containers didn't pull.

In the second case, if the controller can't reach the agent API, then now it won't even know about any jobs to start, let alone be able to fail them directly. (If the API became unavailable in between, then consider that the GraphQL probably had a similar fate.) If instead the controller failed to acquire the job because the job's already acquired, that's equivalent to the first case.

The rest is fairly straightforward cleanup.


wolfeidau (commented):
@DrJosh9000 would it be a good idea to make integration tests a seperate module? 

This would isolate all these graphql deps to this module. 

Just a thought BTW

DrJosh9000:
@wolfeidau That sounds like a good idea, but I'm not sure it's worth untangling right now. If `internal/integration` were made a nested module then `github.com/buildkite/agent-k8s-stack/v2/internal/integration/api` becomes an ambiguous import, and if the `api` subpackage were made a nested module as-is I'd need to copy a bit more of the outer `api` package to avoid an outward dependency. The GraphQL client itself is responsible for only a handful of transitive dependencies.

CerealBoy (approved):
LGTM! 🚀 

Would definitely be a great idea to take graph out completely, certainly not a blocker for the time being though.

wolfeidau:
@DrJosh9000 yeah this makes it tough, was just a thought 🤔

wolfeidau (approved):
👍🏻 🚀

PR #577, opened 2025-04-29
Respect Retry-After header

DrJosh9000:
### What

Handle `Retry-After: <seconds>` as we do in some places in the agent.

### Why

So that there's another mechanism available for reducing load.

### How

Aside from some refactoring in the client to read and return the retry-after value, there are two infinite query loops, and two roko retriers. 

For the infinite loops, use `time.After` to create a channel to wait on. If the header is not present or 0, then the channel is receivable effectively immediately. The only slightly odd-looking thing here is to initialise it with `time.After(0)`, but it should be easy to understand.

For the roko retriers, set the next interval to be the bigger of the retry-after duration and the current next interval. If the header is not present or 0, the retrier continues as it was.

### DIFF
--- a/api/agent_client.go
+++ b/api/agent_client.go
@@ -99,7 +99,7 @@ type AgentScheduledJob struct {
 }
 
 // GetScheduledJobs gets a page of jobs that could be run.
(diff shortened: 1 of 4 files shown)


swaller-bk (approved):
great use of generics to respect the retry-after header agnostic of the specific typed api request 👍

Issue #503, opened 2025-02-10
Allow configuration of image check container resources


bpoland:
We have a LimitRange configuring default resources, but this means that all the image check containers created by the stack inherit this when they really don't need much at all. Could you add the ability to specify these in the controller config/helm chart? Thanks!

PR #580, opened 2025-04-30
Implement configurable resource limits for `imagecheck` containers

petetomasik:
Allows configurable CPU and memory resource limits on `imagecheck-*` containers via new configuration values:
* `image-check-container-cpu-limit`
* `image-check-container-memory-limit`

These configuration values are set to [CPU units](https://kubernetes.io/docs/tasks/configure-pod-container/assign-cpu-resource/#cpu-units) and [memory bytes](https://kubernetes.io/docs/tasks/configure-pod-container/assign-memory-resource/#memory-units). Invalid values for these resources will prevent the controller from starting. Default limits are set to:
```
DefaultImageCheckContainerCPULimit    = "200m"
DefaultImageCheckContainerMemoryLimit = "128Mi"
```

Fixes https://github.com/buildkite/agent-stack-k8s/issues/503


internal/controller/config/config.go
```diff
@@ -11,20 +11,22 @@ import (
...
+	DefaultImagePullBackOffGracePeriod    = 30 * time.Second
+	DefaultJobCancelCheckerPollInterval   = 5 * time.Second
+	DefaultEmptyJobGracePeriod            = 30 * time.Second
+	DefaultJobCreationConcurrency         = 25
+	DefaultK8sClientRateLimiterQPS        = 10
+	DefaultK8sClientRateLimiterBurst      = 20
+	DefaultPaginationPageSize             = 1000
+	DefaultPaginationDepthLimit           = 2
+	DefaultQueryResetInterval             = 10 * time.Second
+	DefaultImageCheckContainerCpuLimit    = "200m"
```
mcncl, line 28: `Cpu` should be `CPU`; it's a pattern we already use with `UUID` and `URL`

internal/controller/scheduler/scheduler.go
```diff
@@ -816,6 +829,10 @@ func (w *worker) Build(podSpec *corev1.PodSpec, skipCheckout bool, inputs buildI
 					corev1.ResourceCPU:    resource.MustParse("100m"),
 					corev1.ResourceMemory: resource.MustParse("64Mi"),
 				},
+				Limits: corev1.ResourceList{
+					corev1.ResourceCPU:    imageCheckContainerCpuLimit,
```
mcncl, line 833: Note here it is referenced as `CPU` in the imported package.

mcncl (changes requested):

internal/controller/config/config.go
petetomasik: All instances of `Cpu` have been replaced with `CPU` in the var names.

internal/controller/scheduler/scheduler.go
petetomasik: All instances of `Cpu` have been replaced with `CPU`.

mcncl (approved):

PR #581, opened 2025-05-05
Add Prometheus metrics docs page + generator

DrJosh9000:
Document the Prometheus metrics and label values.

I started by extracting the existing help strings with a quick and dirty Go program (eww, reflection to access un-exported fields), then added some extra words at the start.


wolfeidau (approved):
Great work, should this be run prior to each release to ensure it is up to date?

DrJosh9000:
I've tweaked the script to write a new file, and a `go:generate` directive so that it gets picked up by the existing step.

PR #582, opened 2025-05-05
feat: pave road to custom annotations

scadu:
The `annotations` field can be used to define custom annotations that will be applied to the `agent-stack-k8s` controller pod:

```yaml
# values.yml
...
annotations:
  kubernetes.io/description: "Agent Stack K8s Controller"
  prometheus.io/scrape: "true"
  prometheus.io/port: "8080"
config:
...
```


ivannalisetska (approved):
Looks good !

PR #585, opened 2025-05-06
Add prometheus-port to schema (it was missing)

DrJosh9000:
In #419 `prometheus-port` was added as a controller flag and used by the Helm chart to configure a container port, but I clearly forgot to put it in the schema or example.


wolfeidau (approved):
🎉 🔍

PR #590, opened 2025-05-08
Bump sleeps for flakey tests

CerealBoy:
This is a bandaid on the integration tests until we can spend some proper time to validate them in a more robust manner. Assuming this continues to pass 😅


catkins (approved):
:+1:

catkins:
https://github.com/buildkite/agent-stack-k8s/blob/7aec6c7b8ca06a80cf83a60ccab4bd309ac4b355/internal/integration/testcase_test.go#L246-L251

I think we could make these tests a bit less brittle, whilst being faster in the typical case if we updated the log assertion to accept a timeout for the logs to appear? (or just hardcode a default timeout I guess), then in the `AssertLogsContain` helper we we could poll the logs every 500ms or whatever until the timeout before we fail the test.

Issue #586, opened 2025-05-07
[BUG] Default Active Deadline Seconds of 60s marks jobs as failed


marc-barry:
## Describe the bug

I have observed that when jobs are running they have the correct default `Active Deadline Seconds: 21600s`. But when the job is complete it changes to `Active Deadline Seconds: 60s` and the jobs are all marked as failed. Here's an example of catching jobs in different states which are running and marked as failed:

```
kubectl get jobs -n buildkite
NAME                                             STATUS          COMPLETIONS   DURATION   AGE
buildkite-0196ab14-908f-4b45-adc9-4f3298ab3db7   Failed          0/1           11m        11m
buildkite-0196ab1b-18d8-41cb-9749-ab46b89c091f   Failed          0/1           4m         4m
buildkite-0196ab1b-18df-4760-acdc-b1b7052a948c   Failed          0/1           4m         4m
buildkite-0196ab1b-18e0-46c4-85e1-de7119930f4d   Failed          0/1           2m15s      2m15s
buildkite-0196ab1b-18e1-4c5d-adf1-8decb3d5770e   Failed          0/1           2m15s      2m15s
buildkite-0196ab1b-18e2-4d3f-97b6-4d63dd49c3c7   FailureTarget   0/1           2m15s      2m15s
buildkite-0196ab1b-18e3-48df-8c6d-cf6556d3d8af   Failed          0/1           2m15s      2m15s
buildkite-0196ab1b-18e4-48e3-b313-df4029180837   Failed          0/1           2m15s      2m15s
buildkite-0196ab1b-18e4-4e62-ad9a-7c8539fbb7af   Failed          0/1           2m15s      2m15s
buildkite-0196ab1b-18e5-4f0b-a6f7-710f026c4a22   Running         0/1           2m15s      2m15s
```

This is the final state:
```
kubectl get jobs -n buildkite
NAME                                             STATUS   COMPLETIONS   DURATION   AGE
buildkite-0196ab1b-18d8-41cb-9749-ab46b89c091f   Failed   0/1           7m56s      7m56s
buildkite-0196ab1b-18df-4760-acdc-b1b7052a948c   Failed   0/1           7m56s      7m56s
buildkite-0196ab1b-18e0-46c4-85e1-de7119930f4d   Failed   0/1           6m11s      6m11s
buildkite-0196ab1b-18e1-4c5d-adf1-8decb3d5770e   Failed   0/1           6m11s      6m11s
[...]

PR #591, opened 2025-05-09
Send `SIGTERM` to any application (non-init) sidecars for cleanup

petetomasik:
Attempts to initiate graceful shutdown of application containers running as sidecars by using `remotecommand.NewSPDYExecutor` to execute a `kill 1` inside each of the containers. This logic does *not* apply to official Kubernetes [Sidecar](https://kubernetes.io/docs/concepts/workloads/pods/sidecar-containers/) containers (which run as init containers), only those specified in the `kubernetes` plugin. Example using sidecar for Docker in Docker:
```
# pipeline.yml
- steps:
    - label: ":k8s:"
      plugins:
        - kubernetes:
          sidecars:
            - name: dind
              image: docker:dind
              command: [dockerd-entrypoint.sh]
              volumeMounts:
                - name: docker-socket
                  mountPath: /var/run/
              securityContext:
                privileged: true
...
```
Removes the update that set the Job's `ActiveDeadlineSeconds` to `60` to cleanup any running sidecars. The default value of `ActiveDeadlineSeconds` for all Jobs created by the controller [defaults](https://github.com/buildkite/agent-stack-k8s/releases/tag/v0.24.0) to `21600` (6 hours), unless overridden in the controller's config or by the `kubernetes` plugin.

Fixes https://github.com/buildkite/agent-stack-k8s/issues/586

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -85,6 +85,7 @@ require (
 	github.com/google/gnostic-models v0.6.8 // indirect
 	github.com/google/s2a-go v0.1.9 // indirect
 	github.com/google/shlex v0.0.0-20191202100458-e7afc7fbc510 // indirect
+	github.com/gorilla/websocket v1.5.0 // indirect
 	github.com/gowebpki/jcs v1.0.1 // indirect
 	github.com/hashicorp/go-secure-stdlib/parseutil v0.1.7 // indirect
 	github.com/hashicorp/go-secure-stdlib/strutil v0.1.2 // indirect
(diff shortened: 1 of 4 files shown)


marc-barry:
@petetomasik this is looking awesome. Thanks for doing this. We have a number of sidecars such as databases and Docker-in-Docker containers as you have in your example.

mcncl (approved):
🚀

PR #596, opened 2025-05-13
Ensure controller agent tags match job tags 100%

petetomasik:
The shared informer watches Jobs matching [Labels created from tags](https://github.com/buildkite/agent-stack-k8s/blob/d36d4288f17bd1e2dcdec4302ec15b7335d3904b/internal/controller/controller.go#L230) in the controller's config. When a scheduled job was retrieved from the API the controller was previously checking if all job tags (`agentQueryRules`) were present as agent tags (and not the other way around). This was allowing scheduled jobs with only a matching `queue` tag to be created as a K8s Job. This then prevented the informer from matching the Labels of these Jobs and the controller was left with an infinitely increasing number of running jobs. The only remediation was to set `max-in-flight: 0` or periodically restart the controller.

Now the controller will only create K8s Jobs when tags match 100%. If tags do not match, the following `INFO` log is emitted:
```
2025-05-13T15:47:38.375-0400    INFO    monitor monitor/monitor.go:197  job tags do not match expected tags in configuration, skipping...       {"org": "my-org", "job-uuid": "0196cb30-235f-4ded-926a-d5a3aaad04a4", "controller-tags": {"foo":"bar","hello":"world","queue":"kubernetes"}, "buildkite-job-tags": {"queue":"kubernetes"}}
```

### DIFF
--- a/internal/controller/agenttags/tags.go
+++ b/internal/controller/agenttags/tags.go
@@ -82,7 +82,7 @@ func LabelsFromTags(tags []string) (map[string]string, []error) {
 	return labels, append(errs1, errs2...)
 }
 
-// AgentTagsMatchJobTags reports whether each tag key in `agentTags` is also
+// MatchJobTags reports whether each tag key in `agentTags` is also
 // present in `jobTags`, and the tag value in `jobTags` is either "*" or the
 // same as the tag value in `agentTags`.
 //
--- a/internal/controller/agenttags/tags.go
+++ b/internal/controller/agenttags/tags.go
@@ -82,20 +82,24 @@ func LabelsFromTags(tags []string) (map[string]string, []error) {
 	return labels, append(errs1, errs2...)
 }
 
-// JobTagsMatchAgentTags reports whether each tag key in `jobTags` is also
-// present in `agentTags`, and the tag value in `jobTags` is either "*" or the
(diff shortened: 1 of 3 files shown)


mcncl (approved):

internal/controller/agenttags/tags.go
```diff
@@ -82,20 +82,24 @@ func LabelsFromTags(tags []string) (map[string]string, []error) {
...
 // same as the tag value in `agentTags`.
 //
 // In the future, this may be expanded to: if the tag value `agentTags` is in some
 // set of strings defined by the tag value in `jobTags` (eg a glob or regex)
 // See https://buildkite.com/docs/agent/v3/cli-start#agent-targeting
-func JobTagsMatchAgentTags(jobTags, agentTags map[string]string) bool {
-	for k, v := range jobTags {
-		agentTagValue, exists := agentTags[k]
+
+func AgentTagsMatchJobTags(agentTags, jobTags map[string]string) bool {
```
DrJosh9000, line 93: Given that this is in the `agenttags` package, `AgentTags` could be dropped from the start of the name so that calling it reads like `agenttags.MatchJobTags(...)`:

```suggestion
func MatchJobTags(agentTags, jobTags map[string]string) bool {
```

DrJosh9000 (approved):
Good catch!

PR #601, opened 2025-05-28
PG-148: ensure checkout-params is properly passed to checkout container

zhming0:
This fixes a regression in beta-5 where some necessary checkout params coming from helm or bk yaml were not correctly to passed to checkout parameters. 

This PR sadly isn't categorically addressing this, but rather just ensure checkout params works. It should be a good net positive change. 

More discussion will happen to discuss how to keep this complexity within our control.


internal/controller/config/checkout_params.go
```diff
@@ -22,8 +22,15 @@ type CheckoutParams struct {
 	ExtraVolumeMounts    []corev1.VolumeMount       `json:"extraVolumeMounts,omitempty"`
 }
 
-func (co *CheckoutParams) ApplyToAgentStart(podSpec *corev1.PodSpec, ctr *corev1.Container) {
-	if co == nil || podSpec == nil || ctr == nil {
+// ApplyToAgentStart send checkout params's env variables to Agent container
+// Agent container will propogate these env variables to command and other containers when they do
+// `bootstrap-kubernetes`.
```
DrJosh9000, line 27:
```suggestion
// `kubernetes-bootstrap`.
```

internal/controller/config/checkout_params.go
```diff
@@ -34,6 +41,16 @@ func (co *CheckoutParams) ApplyToAgentStart(podSpec *corev1.PodSpec, ctr *corev1
...
+
+// Any k8s related config things need to be passed to checkout container directly.
+// "kubernetes-bootstrap" won't work for those for obvious reason: they are passed k8s pod lifecycle.
+//
+// NOTE: despite this is called ApplyToCheckout, it mutate not only the container spec but also the pod spec.
+func (co *CheckoutParams) ApplyToCheckout(podSpec *corev1.PodSpec, ctr *corev1.Container) {
+	if co == nil || podSpec == nil || ctr == nil {
+		return
+	}
 	co.GitMirrors.ApplyTo(podSpec, ctr)
```
DrJosh9000, line 54: For consistency, I think `GitMirrorsParams.ApplyTo` should be split the same way. `GitMirrorsParams.ApplyToAgentStart` to set the env vars, `ApplyToCheckout` to add the volume mount. Maybe the manipulation of `podSpec.Volumes` should be a standalone method too.

DrJosh9000 (commented):
Good stuff!

internal/controller/config/checkout_params.go
zhming0: I am a little bit unsure about this. Because my impression is that the main goal of these env var was to reducing env var numbers on k8s pod spec, and that really works when env var number is variadic. 

For the git mirror case, when the number of config fields is static. I wonder if passing them directly to pod spec is better? 🤔

DrJosh9000: I'm not sure I explained very clearly. I'm envisioning something like:

```go
func (co *CheckoutParams) ApplyToCheckout(podSpec *corev1.PodSpec, ctr *corev1.Container) {
	// ...
	co.GitMirrors.ApplyToCheckout(ctr)
	ctr.EnvFrom = append(ctr.EnvFrom, co.EnvFrom...)
	ctr.VolumeMounts = append(ctr.VolumeMounts, co.ExtraVolumeMounts...)
}

// ...

func (gm *GitMirrorsParams) ApplyToAgentStart(ctr *corev1.Container) {
	if gm == nil || ctr == nil {
		return
	}
	if gm.Volume != nil {
		path := "/buildkite/git-mirrors"
		if gm.Path == nil {
			gm.Path = &path
		}
	}
	appendToEnvOpt(ctr, "BUILDKITE_GIT_MIRRORS_PATH", gm.Path)
	appendToEnvOpt(ctr, "BUILDKITE_GIT_CLONE_MIRROR_FLAGS", gm.CloneFlags)
	if gm.LockTimeout > 0 {
		appendToEnv(ctr, "BUILDKITE_GIT_MIRRORS_LOCK_TIMEOUT", strconv.Itoa(gm.LockTimeout))
	}
	appendBoolToEnvOpt(ctr, "BUILDKITE_GIT_MIRRORS_SKIP_UPDATE", gm.SkipUpdate)
}

func (gm *GitMirrorsParams) ApplyToCheckout(ctr *corev1.Container) {
	if gm.Volume != nil {
		path := "/buildkite/git-mirrors"
		if gm.Path == nil {
			gm.Path = &path
		}
		ctr.VolumeMounts = append(ctr.VolumeMounts, corev1.VolumeMount{
			Name:      gm.Volume.Name,
			MountPath: *gm.Path,
		})
	}
	appendToEnvOpt(ctr, "BUILDKITE_GIT_MIRRORS_PATH", gm.Path)
	appendToEnvOpt(ctr, "BUILDKITE_GIT_CLONE_MIRROR_FLAGS", gm.CloneFlags)
	if gm.LockTimeout > 0 {
		appendToEnv(ctr, "BUILDKITE_GIT_MIRRORS_LOCK_TIMEOUT", strconv.Itoa(gm.LockTimeout))
	}
	appendBoolToEnvOpt(ctr, "BUILDKITE_GIT_MIRRORS_SKIP_UPDATE", gm.SkipUpdate)
}

func (gm *GitMirrorsParams) ApplyToPod(podSpec *corev1.PodSpec) {
	if gm == nil || gm.Volume == nil {
		return
	}
	podSpec.Volumes = append(podSpec.Volumes, *gm.Volume)
}
```

There's a small amount of repetition in there, but probably fine?

zhming0: I currently hold the view that we won't be able to move "ALL" env vars. therefore the consistency will always to be a bit confusing. 

If achieving consistency is not quite realistic, I do lean towards less indirection approach, wdyt?

(lastly, I wonder if you are happy for me to address this in a separated PR so we can have enough time to discuss this properly?)

DrJosh9000: It's more about the flow. setting the config env vars directly on `checkout` works, but overrides whatever the agent tries to set via the socket. Setting the config env vars on `agent` lets the agent filter or transform them as it normally would when forking the bootstrap.

zhming0: > more about the flow

It makes sense. done.

internal/controller/config/checkout_params.go
```diff
@@ -34,7 +41,20 @@ func (co *CheckoutParams) ApplyToAgentStart(podSpec *corev1.PodSpec, ctr *corev1
...
 	appendCommaSepToEnv(ctr, "BUILDKITE_GIT_SUBMODULE_CLONE_CONFIG", co.SubmoduleCloneConfig)
-	co.GitMirrors.ApplyTo(podSpec, ctr)
+
+	co.GitMirrors.ApplyToAgentStart(ctr)
+}
+
+// Any k8s related config things need to be passed to checkout container directly.
+// "kubernetes-bootstrap" won't work for those for obvious reason: they are passed k8s pod lifecycle.
+//
+// NOTE: despite this is called ApplyToCheckout, it mutate not only the container spec but also the pod spec.
```
DrJosh9000, line 51:
```suggestion
// NOTE: despite this is called ApplyToCheckout, it mutate not only the container spec but also the pod spec.
```

Maybe we should do the `GitMirrors.ApplyToPod` call from the scheduler instead

DrJosh9000 (approved):
Nice work, thanks!

PR #602, opened 2025-05-28
Revert PR-591: fixing dangling sidecar container issue

zhming0:
Revert #591. We have reason to believe 591 introduced an issue that if pod restart policy is set to `OnFailure`, sidecar containers will never be terminated.


internal/controller/scheduler/scheduler.go
```diff
@@ -461,8 +461,7 @@ func (w *worker) Build(podSpec *corev1.PodSpec, skipCheckout bool, inputs buildI
 			c.EnvFrom = append(c.EnvFrom, inputs.k8sPlugin.GitEnvFrom...)
 
 			// Add annotations to the Job indicating which of the user-defined
```
DrJosh9000, line 463: Noting that this code to add new annotations was added in #591, but they're not likely to be a problem.

PR #603, opened 2025-05-29
Fix TestCancelCheckerDeletePod flakiness

zhming0:
Use roko to curb flakiness of TestCancelCheckerDeletePod.

### DIFF
--- a/internal/integration/integration_test.go
+++ b/internal/integration/integration_test.go
@@ -2,14 +2,15 @@ package integration_test
 
 import (
 	"context"
+	"fmt"
 	"strconv"
 	"testing"
 	"time"
 
 	"github.com/buildkite/agent-stack-k8s/v2/internal/controller/scheduler"
 	"github.com/buildkite/agent-stack-k8s/v2/internal/integration/api"
+	"github.com/buildkite/roko"
 	"github.com/google/uuid"
-	"github.com/kylelemons/godebug/pretty"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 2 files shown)


internal/integration/integration_test.go
```diff
@@ -619,24 +620,46 @@ func TestCancelCheckerDeletePod(t *testing.T) {
...
 	if err != nil {
-		t.Fatalf("kubernetes.CoreV1().Pods(%q).List(ctx, %v) error = %v", cfg.Namespace, opts, err)
+		return false, err
 	}
 	if len(pods.Items) > 0 {
-		t.Fatalf("kubernetes.CoreV1().Pods(%q).List(ctx, %v): found %d pods, there should be none (they should have been deleted). Pods:\n%s",
-			cfg.Namespace, opts, len(pods.Items), pretty.Sprint(pods.Items))
+		return true, nil
 	}
+	return false, nil
```
DrJosh9000, line 662: Another thought I had reading this (take it or leave it):

```suggestion
return len(pods.Items) > 0, nil
```

internal/integration/integration_test.go
```diff
@@ -619,24 +620,46 @@ func TestCancelCheckerDeletePod(t *testing.T) {
...
+			return fmt.Errorf("found at least one pods, waiting for deletion")
+		}
+		return nil
+	}); err != nil {
+		t.Fatalf("kubernetes.CoreV1().Pods(%q).List(ctx, %v): pods were not deleted after retries. Final error: %v",
+			cfg.Namespace, opts, err)
+	}
+}
+
+func hasJobPod(tc testcase, ctx context.Context, job *api.JobJobTypeCommand) (bool, error) {
```
DrJosh9000, line 650: The logic to construct `opts` is now repeated. Perhaps make `opts` an argument instead of `job`?

But that would leave only the `List` query and some checks. So is the separation into another func worthwhile? Or if it is worth keeping, could it take on more of the responsibility currently in the closure passed to roko? e.g.

```go
func shouldHaveZeroPods(tc testcase, ctx context.Context, opts metav1.ListOptions) error {
	pods, err := tc.Kubernetes.CoreV1().Pods(cfg.Namespace).List(ctx, opts)
	if err != nil {
		return err
	}
	if len(pods.Items) > 0 {
		return fmt.Errorf("found at least one pod, waiting for deletion")
	}
	return nil
}
```

DrJosh9000 (commented):
Looks pretty good!

DrJosh9000 (approved):
LGTM 🚢

PR #605, opened 2025-05-29
Add clean-up-job.yaml for Buildkite kubernetes job clean up 

Damilola-obasa:
Adding the clean up script for users to be able to clean up their old Buildkite kubernetes jobs.


utils/clean-up-job.yaml
zhming0, line 1: [non blocking] I think if we are going to offer this, it would be awesome if offer this as part of our helm installation and: 

* Conditionally provision this when k8s api version <= 1.21.
* Make sure key parameters like k8s namespace follows helm config.

zhming0: But, it's also okay to keep it in utils as a handy tool for customer to manage by themselves, in that case, we should offer ample documents about the context of this cronjob.

utils/clean-up-job.yaml
```diff
@@ -0,0 +1,54 @@
+apiVersion: batch/v1
```
zhming0, line 15:
```suggestion
# This cronjob is a utility useful for k8s version below 1.21, because ...
# You need to install this script in the same namespace with k8s controller.
apiVersion: batch/v1
```

utils/clean-up-job.yaml
```diff
@@ -0,0 +1,54 @@
...
+  failedJobsHistoryLimit: 1
+  jobTemplate:
+    spec:
+      backoffLimit: 1
+      template:
+        spec:
+          serviceAccountName: agent-stack-k8s-controller
+          containers:
+          - name: buildkite-agent-stack-k8s-job-cleanup
+            image: bitnami/kubectl:latest
```
zhming0: This will result in using newer version of kubectl talking to old API, it's a risk. Since we know the version of k8s server, it might be wise to pin the kubectl version down too.

utils/clean-up-job.yaml
```diff
@@ -0,0 +1,54 @@
+apiVersion: batch/v1
+kind: CronJob
+metadata:
+  name: buildkite-agent-stack-k8s-job-cleanup
+  namespace: buildkite
```
DrJosh9000: Further to @zhming0's comment above, given that we make users specify a namespace on install, we should remove the namespace here and instead instruct them to "pass `-n buildkite` (or `-n your-namespace`) to `kubectl apply`" (or similar)

```suggestion
```

utils/clean-up-job.yaml
```diff
@@ -0,0 +1,54 @@
...
+            - name: JOB_AGE_SECONDS
+              value: "600" # Delete jobs that completed over 10 minutes ago
+            command:
+            - /bin/bash
+            - -c
+            - |
+              #!/bin/bash
+              set -euo pipefail
+
+              current_time=$(date +%s)
```
DrJosh9000:
```suggestion
              current_time="$(date +%s)"
```

utils/clean-up-job.yaml
```diff
@@ -0,0 +1,54 @@
...
+              echo "Starting Buildkite job cleanup at $(date -d @$current_time '+%F %H:%M:%S %Z')"
+
+              # Find and clean up old Buildkite jobs using Buildkite labels
+              echo "Finding Buildkite jobs that have been completed for over ${JOB_AGE_SECONDS} seconds..."
+              kubectl get jobs -n "$CLEANUP_K8S_NAMESPACE" \
```
DrJosh9000:
```suggestion
              kubectl get jobs -n "${CLEANUP_K8S_NAMESPACE}" \
```

utils/clean-up-job.yaml
```diff
@@ -0,0 +1,54 @@
...
+      template:
+        spec:
+          serviceAccountName: agent-stack-k8s-controller
+          containers:
+          - name: buildkite-agent-stack-k8s-job-cleanup
+            image: bitnami/kubectl:latest
+            imagePullPolicy: IfNotPresent
+            env:
+            - name: CLEANUP_K8S_NAMESPACE
+              value: "buildkite" # Default agent-stack-k8s namespace
```
DrJosh9000: This is also tricky if the user is using a different namespace. Is there a way to infer this?

utils/clean-up-job.yaml
```diff
@@ -0,0 +1,54 @@
...
+              echo "Starting Buildkite job cleanup at $(date -d @$current_time '+%F %H:%M:%S %Z')"
+
+              # Find and clean up old Buildkite jobs using Buildkite labels
+              echo "Finding Buildkite jobs that have been completed for over ${JOB_AGE_SECONDS} seconds..."
+              kubectl get jobs -n "$CLEANUP_K8S_NAMESPACE" \
+                -l buildkite.com/job-uuid \
+                -o json | \
+                jq -r '.items[] | select(.status.completionTime != null) | "\(.metadata.name) \(.metadata.labels["buildkite.com/job-uuid"]) \(.status.completionTime)"' | \
+                while read -r job_name job_uuid completion_time; do
+                  if [ -z "$job_name" ]; then continue; fi
```
DrJosh9000:
```suggestion
                  if [[ -z "${job_name}" ]]; then continue; fi
```

utils/clean-up-job.yaml
```diff
@@ -0,0 +1,54 @@
...
+
+                  # Check job completion age
+                  completion_timestamp=$(date -d "$completion_time" +%s)
```
DrJosh9000:
```suggestion
                  completion_timestamp="$(date -d "${completion_time}" +%s)"
```

utils/clean-up-job.yaml
```diff
@@ -0,0 +1,54 @@
...
+                  age_seconds=$(( current_time - completion_timestamp ))
```
DrJosh9000:
```suggestion
                  age_seconds="$(( current_time - completion_timestamp ))"
```

utils/clean-up-job.yaml
```diff
@@ -0,0 +1,54 @@
...
+
+                  if [ "$age_seconds" -ge "$JOB_AGE_SECONDS" ]; then
```
DrJosh9000:
```suggestion
                  if (( age_seconds >= JOB_AGE_SECONDS )); then
```

utils/clean-up-job.yaml
```diff
@@ -0,0 +1,54 @@
...
+                    echo "Deleting completed Buildkite job: $job_name (UUID: $job_uuid, completed ${age_seconds} seconds ago)"
+                    kubectl delete job "$job_name" -n "$CLEANUP_K8S_NAMESPACE"
```
DrJosh9000:
```suggestion
                    kubectl delete job "${job_name}" -n "${CLEANUP_K8S_NAMESPACE}"
```

DrJosh9000 (commented):
Here are some shell style suggestions, be sure to check they work before applying as I am only human

utils/clean-up-job.yaml
petetomasik: Update to docs will accompany this PR. This is very much an "extra" tool we are providing as an option for customers to use should they be confined to an older version of K8s that does not support the `.spec.ttlSecondsAfterFinished` Job field. Some customers are already running their own custom cleanup Cronjobs alongside the controller for both older and newer versions of K8s to ensure controller-created Jobs are always cleaned up.

DrJosh9000 (approved):

sj26:
Interesting!

I would expect the controller to notice and clean up orphaned Kubernetes jobs. Does it not do that?

If not, this seems like a good pragmatic add. But I think this should ultimately be handled by the controller.

DrJosh9000:
> I would expect the controller to notice and clean up orphaned Kubernetes jobs. Does it not do that?

It does, but using a mechanism introduced in Kubernetes 1.22. (Which is getting pretty long in the tooth now.)

sj26:
Can you say more - is a critical function of the controller relying on removed features?

PR #609, opened 2025-06-04
PIPE-1084: remove org as stack parameter

zhming0:
`Org` has been an unused parameter in the stack but it's marked as required.

This PR remove this params for good. There are various places in test setup needing it, a `AgentTokenClient` has been introduced so we can derive the Org from agent token.

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -68,7 +68,7 @@ To get the integration test running locally, you will need:
 
 1. A valid Buildkite API token with GraphQL enabled.
 2. A valid Buildkite Agent Token in your target Buildkite Cluster.
-3. The name of your Buildkite organization (slug) and your target Buildkite Cluster UUID.
+3. Your target Buildkite Cluster UUID.
 4. Depending on test cases, you may also need SSH keys, please keep reading.
 5. Your shell environment will need CLI write access to a Kubernetes cluster such as the one provided by https://orbstack.dev/.
 
@@ -78,7 +78,6 @@ It's generally convenient to supply the API token, your Buildkite organization n
 
 ```bash
 export BUILDKITE_TOKEN="bkua_**************"
-export ORG="your-cool-org-slug"
 export CLUSTER_UUID="UUID-UUID-UUID-UUID"
 ```
 
@@ -87,7 +86,7 @@ export CLUSTER_UUID="UUID-UUID-UUID-UUID"
 To run the controller locally, with the environment variables, run the following example. Note that in this example the queue is overridden to ensure jobs from the default queue, which is "", are picked up by the Buildkite agent.
 
 ```bash
-just run --org $ORG --buildkite-token $BUILDKITE_TOKEN --debug --tags 'queue=,os=linux'
+just run --buildkite-token $BUILDKITE_TOKEN --debug --tags 'queue=,os=linux'
 ```
 
 ## Testing locally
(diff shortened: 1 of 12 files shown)


DEVELOPMENT.md
```diff
@@ -68,7 +68,7 @@ To get the integration test running locally, you will need:
 
 1. A valid Buildkite API token with GraphQL enabled.
 2. A valid Buildkite Agent Token in your target Buildkite Cluster.
-3. The name of your Buildkite organization (slug) and your target Buildkite Cluster UUID.
+3. Your target Buildkite Cluster UUID.
```
moskyb, line 71: given that the `/token` endpoint includes the cluster uuid, can we get rid of this from the config as well?

DEVELOPMENT.md
```diff
@@ -68,7 +68,7 @@ To get the integration test running locally, you will need:
 
 1. A valid Buildkite API token with GraphQL enabled.
```
moskyb, line 69: do we still need an API token?

cmd/controller/controller.go
```diff
@@ -396,3 +396,10 @@ func New() *cobra.Command {
 
 	return cmd
 }
+
+func GetTokenIdentity(cfg *config.Config) (result *api.AgentTokenIdentity) {
+	// TODO
+	// client := api.NewAgentTokenClient(config)
+
+	return nil
+}
```
DrJosh9000:
```suggestion
```

DEVELOPMENT.md
zhming0: you read my mind 🙂. Check out this PR too:buildkite/agent-stack-k8s#610

DEVELOPMENT.md
zhming0: Yes, integration test still rely on graphql to create/delete pipeline, queues etc.

DEVELOPMENT.md
moskyb: oooh he's too good

DEVELOPMENT.md
moskyb: ah, good call!

api/agent_token_client.go
```diff
@@ -0,0 +1,58 @@
...
+import (
+	"context"
+	"net/http"
+	"net/url"
+	"time"
+)
+
+// This is a special agent client: it can only be used to query the GET /token API.
+// Meaning for this client to function, there isn't a need for a cluster id nor queue id.
+func NewAgentTokenClient(token, endpoint string) (*AgentTokenClient, error) {
```
DrJosh9000, line 12: Should we combine this with `AgentClient` later on?

zhming0: That was my plan but `AgentClient` needs org, cluster to function. On a quick glance it doesn't seem quite obvious to me on how to combine these (nor on the benefit), so I kept them separate for now.

DrJosh9000: Well, you showed that `org` is unnecessary. I'm thinking that the work `AgentTokenClient` does could be part of "AgentClient connection setup", so that `NewAgentClient` no longer needs the cluster ID as an arg. That leaves the queue parameter, which isn't needed for `AgentTokenClient`, but the arg could be passed an empty string for clients that don't need to get scheduled jobs, etc.

zhming0: I see what you mean now. I agree.

PR #613, opened 2025-06-05
Increase test queue name uniqueness and improve test cleanup robustness

zhming0:
This will resolve our long standing queue key conflict during test run. 

Currently when two builds running at the same time, it's very likely they will have conflict test queue keys because some test names are getting long. `<test name> + <first 3 character of uuid v7>` isn't unique at all.

### DIFF
--- a/internal/integration/testcase_test.go
+++ b/internal/integration/testcase_test.go
@@ -3,6 +3,7 @@ package integration_test
 import (
 	"bytes"
 	"context"
+	"crypto/sha256"
 	"fmt"
 	"os"
 	"slices"
@@ -17,6 +18,7 @@ import (
 	"github.com/buildkite/agent-stack-k8s/v2/internal/controller/config"
 	"github.com/buildkite/agent-stack-k8s/v2/internal/integration/api"
 	"github.com/buildkite/go-buildkite/v3/buildkite"
+	"github.com/buildkite/roko"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"go.uber.org/zap"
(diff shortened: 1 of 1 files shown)


internal/integration/testcase_test.go
```diff
@@ -110,8 +128,17 @@ func (t testcase) createClusterQueueWithCleanup() *buildkite.ClusterQueue {
 			return
 		}
 
-		_, err := t.Buildkite.ClusterQueues.Delete(cfg.Org, cfg.ClusterUUID, *queue.ID)
-		if err != nil {
+		if err := roko.NewRetrier(
+			roko.WithMaxAttempts(5),
+			roko.WithStrategy(roko.Exponential(time.Second, 5*time.Second)),
```
DrJosh9000: I think this is equivalent?
```suggestion
			roko.WithStrategy(roko.Constant(5*time.Second)),
			roko.WithJitter(),
```

internal/integration/testcase_test.go
```diff
@@ -110,8 +128,17 @@ func (t testcase) createClusterQueueWithCleanup() *buildkite.ClusterQueue {
...
+		).DoWithContext(context.Background(), func(r *roko.Retrier) error {
+			_, err := t.Buildkite.ClusterQueues.Delete(cfg.Org, cfg.ClusterUUID, *queue.ID)
+			if err != nil {
+				// There is a small chance that we are deleting queue too soon before queue realize agent has disconnected.
+				return err
+			}
+			return nil
```
DrJosh9000: What you have here is totally fine, but I'm very tempted to suggest
```suggestion
			// There is a small chance that we are deleting queue too soon before queue realize agent has disconnected.
			_, err := t.Buildkite.ClusterQueues.Delete(cfg.Org, cfg.ClusterUUID, *queue.ID)
			return err
```

DrJosh9000 (approved):
Nice

PR #615, opened 2025-06-10
[BREAKING]: listen to default queue when no queue is specified

zhming0:
Context: PIPE-1072

## Background

In the past when we don't supply a queue tag to k8s stack, it will by default listen to a `kubernetes` queue. User would have to manually provision this queue before stack can run. 

This is unnecessary friction (does user even need to know about queue before getting this to run?) and it's inconsistent with agent's behavior where in the case of unspecified queue, it will listen to the "default" queue. 

## Change

This PR changes the k8s stack behavior to align with agent.

After this change, when a user starts with buildkite k8s stack, all they need to supply is the cluster queue token, everything works out of box.

### DIFF
--- a/api/agent_client.go
+++ b/api/agent_client.go
@@ -12,6 +12,9 @@ import (
 	"time"
 )
 
+// This is a special keyword supported on backend for polling from the current default queue in the cluster.
+const defaultQueueKey = "_default"
+
 func NewAgentClient(token, endpoint, clusterID, queue string, agentQueryRules []string) (*AgentClient, error) {
 	if endpoint == "" {
 		endpoint = "https://agent.buildkite.com/v3"
@@ -22,6 +25,9 @@ func NewAgentClient(token, endpoint, clusterID, queue string, agentQueryRules []
 	endpointURL, err := url.Parse(endpoint)
 	if err != nil {
 		return nil, err
+	}
+	if queue == "" {
+		queue = defaultQueueKey
 	}
 	return &AgentClient{
 		endpoint: endpointURL,
--- a/charts/agent-stack-k8s/values.schema.json
+++ b/charts/agent-stack-k8s/values.schema.json
@@ -265,7 +265,7 @@
         "tags": {
           "type": "array",
           "default": [],
-          "title": "Buildkite agent tags used for acquiring jobs - 'queue' is required",
+          "title": "Buildkite agent tags used for acquiring jobs. If you don't specify a queue tag, it will listen to the default queue.",
           "items": {
             "type": "string"
           },
--- a/internal/controller/config/config.go
+++ b/internal/controller/config/config.go
@@ -46,7 +46,7 @@ type Config struct {
 	JobPrefix                string        `json:"job-prefix"               validate:"required"`
 	MaxInFlight              int           `json:"max-in-flight"            validate:"min=0"`
 	Namespace                string        `json:"namespace"                validate:"required"`
-	Tags                     stringSlice   `json:"tags"                     validate:"min=1"`
+	Tags                     stringSlice   `json:"tags"`
 	PrometheusPort           uint16        `json:"prometheus-port"          validate:"omitempty"`
 	ProfilerAddress          string        `json:"profiler-address"         validate:"omitempty,hostname_port"`
 	PaginationPageSize       int           `json:"pagination-page-size"     validate:"min=1,max=1000"`
--- a/internal/controller/controller.go
+++ b/internal/controller/controller.go
@@ -88,8 +88,7 @@ func Run(
 
 	queue := agentTags["queue"]
 	if queue == "" {
-		logger.Error("The 'queue' tag is missing, but it is required", zap.Error(err))
-		return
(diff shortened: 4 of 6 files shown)


charts/agent-stack-k8s/values.schema.json
```diff
@@ -265,7 +265,7 @@
         "tags": {
           "type": "array",
           "default": [],
-          "title": "Buildkite agent tags used for acquiring jobs - 'queue' is required",
+          "title": "Buildkite agent tags used for acquiring jobs. If you don't specify a queue tag, it will listen to default queue.",
```
DrJosh9000:
```suggestion
          "title": "Buildkite agent tags used for acquiring jobs. If you don't specify a queue tag, it will listen to the default queue.",
```

internal/controller/controller.go
```diff
@@ -88,8 +88,7 @@ func Run(
 
 	queue := agentTags["queue"]
 	if queue == "" {
-		logger.Error("The 'queue' tag is missing, but it is required", zap.Error(err))
-		return
+		logger.Info("Listening to the default job queue, to listen to a specific queue, please pass in 'queue' tag")
```
DrJosh9000:
```suggestion
		logger.Info("Listening to the default job queue, to listen to a specific queue, please pass in 'queue' as a tag, e.g. --tags='queue=kubernetes'")
```

api/agent_client.go
```diff
@@ -23,6 +23,10 @@ func NewAgentClient(token, endpoint, clusterID, queue string, agentQueryRules []
 	if err != nil {
 		return nil, err
 	}
+	if queue == "" {
+		// This is a special keyword on backend.
+		queue = "_default"
```
moskyb: perhaps it's `const`ifying the `"_default"` string, if only just to indicate that it's Important

internal/controller/controller.go
moskyb:
```suggestion
		logger.Info("Listening to the default queue for the given cluster. To listen to a specific queue, please pass in 'queue' tag")
```

also, might it be worth doing the defaulting here (`queue = "_default"`) here to keep all of the queue logic in the same place?

api/agent_client.go
sj26: Oh interesting, where do we treat "_default" specially?

zhming0: @sj26 it's implemented in https://github.com/buildkite/buildkite/pull/23261 

@moskyb yep I constified this string.

internal/controller/controller.go
zhming0: Updated the log line. 

> might it be worth doing the defaulting here

hmm, I don't think we should make the `_default` concept visible beyond transport layer, since it's really a transport layer thing. 

For most code paths in agent stack k8s, it should probably operate as: "I didn't set a queue tag, I don't know about the _default string".

api/agent_client.go
```diff
@@ -12,6 +12,9 @@ import (
 	"time"
 )
 
+// This is a special keyword supported on backend for polling from the current default queue in the cluster.
+const DEFAULT_QUEUE_KEY = "_default"
```
moskyb:
```suggestion
const DefaultQueueKey = "_default"
```
:gopher:

zhming0: 👍🏿 Golang memory is slowly coming back!

api/agent_client.go
sj26: Interesting!

Why a special value, why not treat empty/missing as "default"?

sj26: (Not a blocking question, feel free to ship and get this working!)

zhming0: @sj26 I would've asked the same question. My first thought was to use empty to represent "default" but it has been tricky given the url structure is: `/cluster/~/queue/~/scheduled_jobs`, allowing empty queue bit will confuse the router. 

My second thought was to pull default queue key before making this request, but it will suffer issue when default queue changes on backend. 

So, with the understanding that empty value is a subset of "special values", we end up using special keyword. 

The special keyword is safe because a real queue key has to start with alphanumeric characters.

api/agent_client.go
```diff
@@ -12,6 +12,9 @@ import (
 	"time"
 )
 
+// This is a special keyword supported on backend for polling from the current default queue in the cluster.
+const DefaultQueueKey = "_default"
```
DrJosh9000: It probably doesn't need to be an exported name, especially if it is an internal protocol detail.

```suggestion
const defaultQueueKey = "_default"
```

internal/controller/config/config.go
```diff
@@ -46,7 +46,7 @@ type Config struct {
 	JobPrefix                string        `json:"job-prefix"               validate:"required"`
 	MaxInFlight              int           `json:"max-in-flight"            validate:"min=0"`
 	Namespace                string        `json:"namespace"                validate:"required"`
-	Tags                     stringSlice   `json:"tags"                     validate:"min=1"`
+	Tags                     stringSlice   `json:"tags"                     validate:"min=0"`
```
DrJosh9000: May as well drop the `validate` tag entirely.

```suggestion
	Tags                     stringSlice   `json:"tags"`
```

DrJosh9000 (changes requested):
LGTM with these changes.

PR #616, opened 2025-06-10
Add integration test for default queue handling

DrJosh9000:
To catch the regression fixed in #604.

### DIFF
--- a/internal/integration/integration_test.go
+++ b/internal/integration/integration_test.go
@@ -43,6 +43,10 @@ func TestWalkingSkeleton(t *testing.T) {
 }
 
 func TestDefaultQueue(t *testing.T) {
+	// Note: this test assumes the default queue is called "default".
+	// This happens to be the case for our CI setup. 
+	// TODO: generalise the test to work with any name default queue once the 
+	// controller can function without setting the queue explicitly.
 	tc := testcase{
 		T:           t,
 		Fixture:     "default-queue.yaml",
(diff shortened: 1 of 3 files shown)


internal/integration/integration_test.go
```diff
@@ -39,6 +39,24 @@ func TestWalkingSkeleton(t *testing.T) {
...
+		T:           t,
+		Fixture:     "default-queue.yaml",
+		Repo:        repoHTTP,
+		GraphQL:     api.NewGraphQLClient(cfg.BuildkiteToken, cfg.GraphQLEndpoint),
+		CustomQueue: "default",
+	}.Init()
+	ctx := context.Background()
+	// Note: this shouldn't interfere with the stack running the tests, because
+	// the stack uses the "kubernetes" queue.
+	pipeline := tc.createPipelineWithCleanup(ctx, "default")
```
zhming0: It won't impact the stack running the tests also because they use different clusters. 

But this will in effect make the concurrency of our test pipeline to be 1. It sounds like an unfortunate tradeoff that we have to take for now..

also I wonder if there is a way to prevent people from running this by default?

internal/integration/integration_test.go
```diff
@@ -39,6 +39,24 @@ func TestWalkingSkeleton(t *testing.T) {
 	)
 }
 
+func TestDefaultQueue(t *testing.T) {
+	tc := testcase{
+		T:           t,
+		Fixture:     "default-queue.yaml",
+		Repo:        repoHTTP,
+		GraphQL:     api.NewGraphQLClient(cfg.BuildkiteToken, cfg.GraphQLEndpoint),
+		CustomQueue: "default",
```
zhming0, line 55: It seems this test can only work in an org with the default queue having the key `default`. I think it might be something worth documenting.

internal/integration/integration_test.go
DrJosh9000: I just thought of a tweak to this approach that should avoid the concurrency problem - we could use another tag to target the job to the controller.

internal/integration/integration_test.go
DrJosh9000: Added as a comment for now.

PR #619, opened 2025-06-11
Fix missing hooks and plugins volume mounts

### LINKED ISSUE #617: [BUG] Hooks are not automatically mounted inside command and checkout containers
## Describe the bug
According to https://buildkite.com/docs/agent/v3/agent-stack-k8s/setting-up-hooks-and-plugins, setting up agent hooks should be as simple as setting up a `hooksVolume` in de `agent-config` of `values.yaml`. Sadly, this does not entirely work as advertised.

Specifically, the documentation mentions:
> If used, the corresponding volumes named buildkite-hooks and buildkite-plugins will be automatically mounted on checkout and command containers, with the Buildkite Agent configured to use them.

However, this does not appear to be the case. Once `hooksVolume` is set-up, it is in fact added as a volume to the `PodSpec` and a volume mount is created in the `agent` container, however, for the command and checkout container, no volume mount is created automatically and no environment variable `BUILDKITE_HOOKS_PATH` is set either, causing hooks not to run as advertised per documentation.

## To Reproduce
1. Deploy with a `hooksVolume`, e.g.:
```yaml
config:
  agent-config:
    hooksVolume:
      name: buildkite-hooks
      configMap:
        defaultMode: 493  # 755
        name: some-buildkite-hooks-configmap
```

2. Run a simple command inside a job on the Kubernetes cluster. No Kubernetes plug-in required:
```yaml
- name: ":buildkite:"
  command: bin/some_command
```

3. See that the scheduled job correctly creates the volume inside the PodSpec and mounts it in the `agent` container, but not the `checkout` or `container-0`.

## Expected behavior
Based on the documentation, I expect `volumeMount`s to be automatically created for the `checkout` and `container-0` containers. I furthermore expect the `BUILDKITE_HOOKS_PATH` environment variable to be set automatically and correctly.

## Environment
- agent-stack-k8s version: `0.28.0`
- Kubernetes version: `1.33.1`

## Logs
N/A

## Additional context
Based on the source code of the controller, I believe the hookVolume is only mounted here https://github.com/buildkite/agent-stack-k8s/blob/main/internal/controller/config/agent_config.go#L107-L117, but that is indeed only applied to the `agent` container. Something like `ApplyToCommand` does not contain this: https://github.com/buildkite/agent-stack-k8s/blob/main/internal/controller/config/agent_config.go#L174-L189

As a user, I can workaround this bug by manually providing `extraVolumeMounts` through the Kubernetes plugin. By using the volume already added to the podspec and setting the `BUILDKITE_HOOKS_PATH` environment variable, I can get hooks to work correctly:
```
- name: ":buildkite:"
  command: bin/some_command
  env:
    BUILDKITE_HOOKS_PATH: /buildkite/hooks
  kubernetes:
    extraVolumeMounts:
      - mountPath: /buildkite/hooks
        name: buildkite-hooks
```

Furthermore, based on the source code, I expect exactly the same problem to be present for the plugins volume, however, I have not tested that in detail, since I had no particular need for it in our use case.

DrJosh9000:
### What
Attach the hooks and plugins volumes to the other containers (as they were before #595).

Add an integration test to prevent regression.

### Why

Fixes #617 

### Notes

Hooks are easy to write in a configmap and then mount.

Plugins are less easy to write declaratively, since all plugins (except vendored plugins) are Git repos, and the agent expects to "check out" plugins that already exist in the local filesystem into the plugins path. So the plugins path has to be writable, so a config map can't be mounted there, requiring shenanigans (PodSpecPatch, ExtraVolumeMounts) just to get working, and because it must be a Git repo, it would contain ugly blobs of binary data. In lieu of that, the test specifies a new emptyDir for PluginsVolume and just uses metahook, which should be checked out into that directory.

### DIFF
--- a/internal/controller/config/agent_config.go
+++ b/internal/controller/config/agent_config.go
@@ -104,28 +104,10 @@ func (a *AgentConfig) ApplyToAgentStart(ctr *corev1.Container) {
 	appendCommaSepToEnv(ctr, "BUILDKITE_AGENT_DISABLE_WARNINGS_FOR", a.DisableWarningsFor)
 	appendBoolToEnvOpt(ctr, "BUILDKITE_AGENT_DEBUG_SIGNING", a.DebugSigning)
 
-	if a.HooksVolume != nil {
-		hooksPath := "/buildkite/hooks"
-		if a.HooksPath == nil {
-			a.HooksPath = &hooksPath
-		}
-		ctr.VolumeMounts = append(ctr.VolumeMounts, corev1.VolumeMount{
-			Name:      a.HooksVolume.Name,
-			MountPath: *a.HooksPath,
-		})
-	}
+	a.applyHooksVolumeTo(ctr)
 	appendToEnvOpt(ctr, "BUILDKITE_HOOKS_PATH", a.HooksPath)
 
-	if a.PluginsVolume != nil {
-		pluginsPath := "/buildkite/plugins"
-		if a.PluginsPath == nil {
-			a.PluginsPath = &pluginsPath
-		}
-		ctr.VolumeMounts = append(ctr.VolumeMounts, corev1.VolumeMount{
-			Name:      a.PluginsVolume.Name,
-			MountPath: *a.PluginsPath,
-		})
-	}
+	a.applyPluginsVolumeTo(ctr)
 	appendToEnvOpt(ctr, "BUILDKITE_PLUGINS_PATH", a.PluginsPath)
 
 	// The agent transforms these into the corresponding negated versions when
@@ -175,6 +157,8 @@ func (a *AgentConfig) ApplyToCommand(ctr *corev1.Container) {
 	if a == nil || ctr == nil {
 		return
 	}
+	a.applyHooksVolumeTo(ctr)
+	a.applyPluginsVolumeTo(ctr)
 	// Signing happens either with "pipeline upload" or "tool sign", so the
 	// signing side of any key needs to be attached to the command container,
 	// not the agent start container or checkout containers.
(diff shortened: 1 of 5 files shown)


internal/controller/config/agent_config.go
```diff
@@ -188,6 +172,40 @@ func (a *AgentConfig) ApplyToCommand(ctr *corev1.Container) {
 	})
 }
 
+func (a *AgentConfig) ApplyToCheckout(ctr *corev1.Container) {
+	if a == nil || ctr == nil {
+		return
+	}
```
zhming0, line 178: Non-blocking but I feel many of these guard clause can throw fatal error (panic) instead.

internal/integration/integration_test.go
```diff
@@ -713,3 +716,49 @@ func TestJobActiveDeadlineSeconds(t *testing.T) {
...
+				},
+			},
+		},
+		PluginsVolume: &corev1.Volume{
+			Name: "plugins",
+			VolumeSource: corev1.VolumeSource{
+				EmptyDir: &corev1.EmptyDirVolumeSource{},
+			},
+		},
+	}
```
zhming0: [not blocking] I didn't realize this is the hurdles that people need to jump over in order to use plugins etc.

internal/controller/config/agent_config.go
DrJosh9000: If it were to panic, it would almost be better to let it panic on the line where a nil pointer is used.

In this case, I have been introducing these guard clauses here as a way to cut down on the alternative, which is a bit more typing in `scheduler.go`, e.g.:

```go
if config.AgentConfig != nil {
	config.AgentConfig.ApplyToCommand(c)
}
```

Normally I don't mind the typing, but `scheduler.go` is too big and complicated.

internal/integration/integration_test.go
RemcodM: Plugins seem to also work without a valid `PluginsVolume`, I guess that in that case they are kept local to the pod. However, that means that every job should checkout the plugins again. I did not test this, but I suspect that settings a `PluginsVolume` helps keeping the plugins cached.

DrJosh9000: Without `PluginsVolume`, plugins would be checked out to a directory within the `BUILDKITE_PLUGINS_PATH` value set over in the scheduler, which is `/workspace/plugins`. So they end up in the same emptyDir volume (or whatever is configured for `WorkspaceVolume`) as the rest of the `/workspace`.

Unlike `HooksVolume`, which makes it easier to supply a set of local hooks for the agent to run from a volume, `PluginsVolume` is really just providing a separate space for the agent to write downloaded plugins. 

`PluginsVolume` probably won't even help plugin caching, since the agent goes to the trouble of making separate subdirectories per agent, and the agent's identity within each job is different. (The volume would have to implement content-based deduplication.)

In terms of changes going forward, there's an agent improvement here: add an easier way to supply a set of plugins to the agent through the local filesystem, and not require them to be structured as Git repos or require a writable filesystem to use them.

PR #620, opened 2025-06-11
PIPE-1139: fix checkout skip crash + container-0 double handling

### LINKED ISSUE #618: [BUG] Running command without explicit containers and skipping checkout crashed agent
## Describe the bug
When running a job within the Kubernetes cluster that does not set containers through the Kubernetes plugin, while setting `checkout: { skip: true }`, causes the agent in `container-0` to crash silently.

## To Reproduce
1. Deploy with any minimal configuration
2. Run a job that does not use the Kubernetes plugin other than skipping checkout:
```yaml
- name: ":buildkite:"
  command: bin/some_command
  kubernetes:
    checkout:
      skip: true
```
3. See that job in Buildkite keeps to be running forever (possibly up to the job timeout, did not test) without any output.

## Expected behavior
I expect the command to run and be executed as part of the job, or at least for failure to be represented through the Buildkite job... Now it fails 'silently'.

## Environment
- agent-stack-k8s version: `0.28.0`
- Kubernetes version: `1.33.1`

## Logs
```
2025-06-10 17:49:16.129Z copy-agent + mkdir /workspace/sockets
2025-06-10 17:49:16.131Z copy-agent + chmod 777 /workspace/sockets
2025-06-10 17:49:16.132Z copy-agent + cp /usr/local/bin/buildkite-agent /sbin/tini-static /workspace
2025-06-10 17:49:25.335Z agent 
2025-06-10 17:49:25.335Z agent    _           _ _     _ _    _ _                                _
2025-06-10 17:49:25.335Z agent   | |         (_) |   | | |  (_) |                              | |
2025-06-10 17:49:25.335Z agent   | |__  _   _ _| | __| | | ___| |_ ___    __ _  __ _  ___ _ __ | |_
2025-06-10 17:49:25.335Z agent   | '_ \| | | | | |/ _` | |/ / | __/ _ \  / _` |/ _` |/ _ \ '_ \| __|
2025-06-10 17:49:25.335Z agent   | |_) | |_| | | | (_| |   <| | ||  __/ | (_| | (_| |  __/ | | | |_
2025-06-10 17:49:25.335Z agent   |_.__/ \__,_|_|_|\__,_|_|\_\_|\__\___|  \__,_|\__, |\___|_| |_|\__|
2025-06-10 17:49:25.335Z agent                                                  __/ |
2025-06-10 17:49:25.335Z agent  https://buildkite.com/agent                    |___/
2025-06-10 17:49:25.335Z agent 
2025-06-10 17:49:25.335Z agent 2025-06-10 17:49:25 NOTICE Starting buildkite-agent v3.98.1 with PID: 7
2025-06-10 17:49:25.335Z agent 2025-06-10 17:49:25 NOTICE The agent source code can be found here: https://github.com/buildkite/agent
2025-06-10 17:49:25.335Z agent 2025-06-10 17:49:25 NOTICE For questions and support, email us at: hello@buildkite.com
2025-06-10 17:49:25.335Z agent 2025-06-10 17:49:25 INFO   Configuration loaded path=/buildkite/buildkite-agent.cfg
2025-06-10 17:49:25.335Z agent 2025-06-10 17:49:25 INFO   Build Path doesn't exist, creating it (/workspace/build)
2025-06-10 17:49:25.335Z agent 2025-06-10 17:49:25 INFO   Registering agent with Buildkite...
2025-06-10 17:49:25.335Z agent 2025-06-10 17:49:25 WARN   Failed to find unique machine-id: machineid: machineid: open /etc/machine-id: no such file or directory
2025-06-10 17:49:25.553Z agent 2025-06-10 17:49:25 INFO   Successfully registered agent "buildkite-01975af4-8207-4fa9-beaa-5522eb570e36-c2qn6-1" with tags [k8s:agent-stack-version=v0.28.0, queue=kubernetes, k8s:service-account=default, k8s:node=pi2, k8s:namespace=buildkite]
2025-06-10 17:49:25.554Z agent 2025-06-10 17:49:25 INFO   Starting 1 Agent(s)
2025-06-10 17:49:25.554Z agent 2025-06-10 17:49:25 INFO   You can press Ctrl-C to stop the agents
2025-06-10 17:49:25.554Z agent 2025-06-10 17:49:25 INFO   buildkite-01975af4-8207-4fa9-beaa-5522eb570e36-c2qn6-1 Connecting to Buildkite...
2025-06-10 17:49:25.719Z agent 2025-06-10 17:49:25 INFO   buildkite-01975af4-8207-4fa9-beaa-5522eb570e36-c2qn6-1 Attempting to acquire job 01975af4-8207-4fa9-beaa-5522eb570e36...
2025-06-10 17:49:26.087Z agent 2025-06-10 17:49:26 INFO   buildkite-01975af4-8207-4fa9-beaa-5522eb570e36-c2qn6-1 Starting job 01975af4-8207-4fa9-beaa-5522eb570e36 for build at https://buildkite.com/moneybird/moneybird-app-2-dot-0-ci-on-premise/builds/30
2025-06-10 17:50:30.286Z container-0 buildkite-agent: fatal: error connecting to kubernetes runner: unrecognized client id: 0
```

## Additional context
The error thrown by the Buildkite Agent in the `container-0` container seems relevant:
> `buildkite-agent: fatal: error connecting to kubernetes runner: unrecognized client id: 0`

Based on my inspection of the source code, this seems to originate from the following line in the Buildkite Agent itself: https://github.com/buildkite/agent/blob/main/kubernetes/runner.go#L253.
This check suggests that ID 0 was unexpected. The ID is checked based on the length of the slice of `clientResult`s, which appears to be empty in this case. Based on https://github.com/buildkite/agent/blob/main/kubernetes/runner.go#L41, I believe the length of the slice to be based on the `ClientCount` from the config, which is influenced by the `BUILDKITE_CONTAINER_COUNT` environment variable (https://github.com/buildkite/agent/blob/main/agent/job_runner.go#L366-L373), which is indeed set to `0` for the failing job.

The reason that `BUILDKITE_CONTAINER_COUNT` appears to be `0` is because the `container-0` seems to never be counted within this container count. This happens here: https://github.com/buildkite/agent-stack-k8s/blob/main/internal/controller/scheduler/scheduler.go#L414. In this case, because we skip checkout, `systemContainerCount` was set `0` at https://github.com/buildkite/agent-stack-k8s/blob/main/internal/controller/scheduler/scheduler.go#L334-L337. Because there are not explicitly defined containers as part of the job configuration, `len(podSpec.Containers)` will still be `0`. Only directly after line 414, at https://github.com/buildkite/agent-stack-k8s/blob/main/internal/controller/scheduler/scheduler.go#L416, the `container-0` will be added to run the specified command. This used `BUILDKITE_CONTAINER_ID`, but `containerCount` is never increased, causing the `agent` container to start with `BUILDKITE_CONTAINER_COUNT` set to `0`.

Based on this explanation, I also suspect something fishy going on with the container IDs when checkout is not skipped. In that case, `systemContainerCount` is set to `1`, causing the `container-0` to have `BUILDKITE_CONTAINER_ID` equal to `1`, but `BUILDKITE_CONTAINER_COUNT` is also still `1`, meaning one `clientResult` will be missing in the `agent`. Somehow, this has not really caused issues yet, but as I write this we are still testing serveral configurations, so there might be something broken there as well.

The simple workaround would be to add explicitly defined containers through the Kubernetes plugin:
```yaml
- name: ":buildkite:"
  kubernetes:
    checkout:
      skip: true
    podSpec:
      containers:
        - image: alpine:latest
          command: sleep 10
```
In that case there are already containers present in the `podSpec`, which will cause the calculations to be correct.

zhming0:
Fixes #618. 

This PR fixes an intricate bug. When `podSpec` is not nil, but does not contain any containers, and at the same time when `checkout.skip = true`, it will cause an incorrectly configured `container-0` to be spawned. 

The obvious reason is that the `containerCount` calculation is wrong. But if this is the case, this issue is supposed to be pretty obvious and discoverable. 

The only reason that we only discover this know is that, we have (seemly incorrect) double handled the default `container-0` logic. Removing that makes this defect easily reproducible. 

This fix could mean a breaking change for a subset of users, I will use inline comment to explain this.

### DIFF
--- a/internal/controller/scheduler/scheduler.go
+++ b/internal/controller/scheduler/scheduler.go
@@ -146,15 +146,7 @@ func (w *worker) Handle(ctx context.Context, job *api.AgentScheduledJob) error {
 		return w.failJob(ctx, inputs, fmt.Sprintf("agent-stack-k8s failed to parse the job: %v", err))
 	}
 
-	// Default command container using default image.
-	podSpec := &corev1.PodSpec{
-		Containers: []corev1.Container{
-			{
-				Image:   w.cfg.Image,
-				Command: []string{jobToRun.Command},
-			},
-		},
-	}
+	podSpec := &corev1.PodSpec{}
 	// Use the podSpec provided by the plugin, if allowed.
 	if inputs.k8sPlugin != nil && inputs.k8sPlugin.PodSpec != nil {
 		podSpec = inputs.k8sPlugin.PodSpec
@@ -411,8 +403,6 @@ func (w *worker) Build(podSpec *corev1.PodSpec, skipCheckout bool, inputs buildI
 		podSpec.Containers[i] = c
 	}
 
-	containerCount := len(podSpec.Containers) + systemContainerCount
-
 	if len(podSpec.Containers) == 0 {
 		// Create a default command container named "container-0".
 		c := corev1.Container{
@@ -450,6 +440,7 @@ func (w *worker) Build(podSpec *corev1.PodSpec, skipCheckout bool, inputs buildI
 		podSpec.Containers = append(podSpec.Containers, c)
 	}
 
+	sidecarCounterCount := 0
 	if inputs.k8sPlugin != nil {
 		for i, c := range inputs.k8sPlugin.Sidecars {
 			if c.Name == "" {
@@ -468,6 +459,8 @@ func (w *worker) Build(podSpec *corev1.PodSpec, skipCheckout bool, inputs buildI
 			maps.Copy(kjob.Spec.Template.Annotations, sidecarAnnotation)
 
 			podSpec.Containers = append(podSpec.Containers, c)
+
+			sidecarCounterCount += 1
 		}
 	}
 
@@ -489,6 +482,12 @@ func (w *worker) Build(podSpec *corev1.PodSpec, skipCheckout bool, inputs buildI
 	redactedVars := slices.Clone(clicommand.RedactedVars.Value.Value())
 	redactedVars = append(redactedVars, w.cfg.AdditionalRedactedVars...)
 
+	// Managed containers are those containers that runs agent as parent process.
+	// The coordinating agent container are expecting those managed containers to connect.
+	//
+	// Calculating this imperatively is risky given we lack control over the context, this is subject to refactor.
+	managedContainerCount := len(podSpec.Containers) + systemContainerCount - sidecarCounterCount
+
 	agentContainer := corev1.Container{
 		Name:         AgentContainerName,
 		Args:         []string{"start"},
@@ -502,7 +501,7 @@ func (w *worker) Build(podSpec *corev1.PodSpec, skipCheckout bool, inputs buildI
 			},
 			{
 				Name:  "BUILDKITE_CONTAINER_COUNT",
-				Value: strconv.Itoa(containerCount),
+				Value: strconv.Itoa(managedContainerCount),
 			},
 			{
 				Name:  "BUILDKITE_AGENT_TAGS",
--- a/internal/integration/fixtures/skip-checkout.yaml
+++ b/internal/integration/fixtures/skip-checkout.yaml
@@ -1,0 +1,11 @@
+
+steps:
+  - label: "No checkout"
+    agents:
+      queue: "{{.queue}}"
+    command: echo hello-skip-checkout
+    plugins:
+      - kubernetes:
+          checkout:
+            skip: true
+          podSpec: {}
--- a/internal/integration/integration_test.go
+++ b/internal/integration/integration_test.go
@@ -44,8 +44,8 @@ func TestWalkingSkeleton(t *testing.T) {
 
 func TestDefaultQueue(t *testing.T) {
 	// Note: this test assumes the default queue is called "default".
-	// This happens to be the case for our CI setup. 
-	// TODO: generalise the test to work with any name default queue once the 
+	// This happens to be the case for our CI setup.
+	// TODO: generalise the test to work with any name default queue once the
 	// controller can function without setting the queue explicitly.
 	tc := testcase{
 		T:           t,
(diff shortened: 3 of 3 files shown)


internal/controller/scheduler/scheduler.go
```diff
@@ -146,15 +146,7 @@ func (w *worker) Handle(ctx context.Context, job *api.AgentScheduledJob) error {
...
 
-	// Default command container using default image.
-	podSpec := &corev1.PodSpec{
-		Containers: []corev1.Container{
-			{
-				Image:   w.cfg.Image,
-				Command: []string{jobToRun.Command},
-			},
-		},
-	}
```
zhming0, line 157: I believe this change is reasonable as we are doubling down agent as a parent process strategy. 

However, I can imagine that it will impact some users who do not want agent to be the parent process. I think as we are recently about to release a number of breaking change, this is a good moment to draw a line.

internal/controller/scheduler/scheduler.go
```diff
@@ -412,6 +404,10 @@ func (w *worker) Build(podSpec *corev1.PodSpec, skipCheckout bool, inputs buildI
 	}
 
 	containerCount := len(podSpec.Containers) + systemContainerCount
+	if len(podSpec.Containers) == 0 {
+		// We will at least have a container 0 even if no containers are specified in pod spec
+		containerCount = 1 + systemContainerCount
+	}
```
DrJosh9000: The condition in this `if` is the same as the condition immediately below. Rather than adding another `if`, or even conditionally computing the value at all, maybe the `len(podSpec.Containers) + systemContainerCount` computation could be done when setting `BUILDKITE_CONTAINER_COUNT` on the agent container.

internal/controller/scheduler/scheduler.go
DrJosh9000: Due to the processing within `Build`, the agent should get added as a parent process anyway. But I'm in favour of removing this code block and letting the no-containers logic in `Build` handle creating the default.

internal/controller/scheduler/scheduler.go
zhming0: Make sense. Updated 👍🏿 .

internal/controller/scheduler/scheduler.go
zhming0: > Due to the processing within Build, the agent should get added as a parent process anyway

Good to know this isn't a breaking change!

DrJosh9000 (approved):
LGTM!

internal/controller/scheduler/scheduler.go
zhming0: as a result, because `podSpec.Containers` is mutable, its scope changes so this calculation is wrong. I made an bandaid looking update.

PR #622, opened 2025-06-12
PIPE-1072-2: fix explicitly setting job default queue causing controller to miss the job

zhming0:
When bk yaml looks like: 

```
agent: 
  queue: default # -> happen to be the default queue
```

And k8s controller starts without a queue tag, the controller will miss the job because controller wants a job that has `queue=""` instead. 

----

This PR fixed it on transport layer so when we poll jobs, its queue tag will always match our expectation.

### DIFF
--- a/internal/controller/agenttags/tags.go
+++ b/internal/controller/agenttags/tags.go
@@ -42,6 +42,12 @@ func SetTag(tags []string, key, val string) []string {
 	return tags
 }
 
+func RemoveTag(tags []string, key string) []string {
+	return slices.DeleteFunc(tags, func(tag string) bool {
+		return strings.HasPrefix(tag, key+"=")
+	})
+}
+
 // labelsFromTagMap converts map[key->value] to map[tag.buildkite.com/key->value],
 // with k8s compatibility checks
 func labelsFromTagMap(m map[string]string) (map[string]string, []error) {
--- a/internal/integration/integration_test.go
+++ b/internal/integration/integration_test.go
@@ -71,6 +71,35 @@ func TestDefaultQueue(t *testing.T) {
 	tc.AssertLogsContain(build, "Hi there")
 }
 
+func TestExplicitDefaultQueue(t *testing.T) {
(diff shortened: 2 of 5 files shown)


internal/controller/agenttags/tags.go
```diff
@@ -42,6 +42,16 @@ func SetTag(tags []string, key, val string) []string {
...
 }
 
+func RemoveTag(tags []string, key string) []string {
+	i := slices.IndexFunc(tags, func(tag string) bool {
+		return strings.HasPrefix(tag, key+"=")
+	})
+	if i < 0 {
+		return tags
+	}
+	return slices.Delete(tags, i, i+1)
```
DrJosh9000: Avoids dealing with an index `i`:

```suggestion
	return slices.DeleteFunc(tags, func(tag string) bool {
		return strings.HasPrefix(tag, key+"=")
	})
```

internal/integration/integration_test.go
```diff
@@ -71,6 +71,37 @@ func TestDefaultQueue(t *testing.T) {
 	tc.AssertLogsContain(build, "Hi there")
 }
 
+// This is test case for an interesting edge case!
+// Read more from: https://buildkite-corp.slack.com/archives/C08N9Q51CTC/p1749704326621419?thread_ts=1749704031.661719&cid=C08N9Q51CTC
```
DrJosh9000: Rather than a link to Slack, a couple of words explaining it here would be better both for others (that don't have access to our Slack instance) and us (so we don't have to read through the thread, some of which isn't relevant to  this bug).

Or maybe the comment in `GetScheduledJobs` is enough?

DrJosh9000 (approved):

internal/integration/integration_test.go
zhming0: ah yeah this is a public repo! 🤦🏿 

I removed it. I agree git blame and other comments should offer enough breadcrumbs.

PR #624, opened 2025-06-13
Output bad yaml when kube api server rejects the request

nsuma8989:
Output bad yaml when kube api server rejects the request.

### DIFF
--- a/internal/controller/scheduler/scheduler.go
+++ b/internal/controller/scheduler/scheduler.go
@@ -34,6 +34,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/strategicpatch"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/utils/ptr"
+	"sigs.k8s.io/yaml"
 )
 
 const (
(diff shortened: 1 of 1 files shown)


internal/controller/scheduler/scheduler.go
```diff
@@ -168,8 +169,15 @@ func (w *worker) Handle(ctx context.Context, job *api.AgentScheduledJob) error {
 			return fmt.Errorf("%w: %w", model.ErrDuplicateJob, err)
 
 		case kerrors.IsInvalid(err):
+			out := ""
+			outBytes, errYaml := yaml.Marshal(kjob)
```
DrJosh9000: Some notes on variable naming. More of an art than a science.

"Bytes" is clear from the type, and it seems like that's only present to deconflict with `out`. `errYaml` could be more specific to the function/action that produced the error (_marshaling_ YAML). If `YAML` needs to be in the name then it would be `errYAML`.

So I would tweak the variable names like so: 

```suggestion
			y, marshalErr := yaml.Marshal(kjob)
```

Single-letter var names are usually fine when the number of lines the variable is used across is small, as it is in this case, but if `y` feels too short, then consider:

```suggestion
			yamlOut, marshalErr := yaml.Marshal(kjob)
```

internal/controller/scheduler/scheduler.go
```diff
@@ -168,8 +169,15 @@ func (w *worker) Handle(ctx context.Context, job *api.AgentScheduledJob) error {
...
+			if errYaml != nil {
+				out = fmt.Sprintf("%v", errYaml)
```
DrJosh9000: Error messages start lower-case to make them easier to incorporate into longer sentences. This would be the place to provide that context for human consumption:

```suggestion
				out = fmt.Sprintf("Couldn't marshal the Job into YAML: %v", marshalErr)
```

DrJosh9000 (approved):

PR #626, opened 2025-06-16
PIPE-1171: fix all jobs must match agent tags bug + introduce controller ID

### LINKED ISSUE #625: [BUG] All job tags must match agent tags
## Describe the bug

This PR (https://github.com/buildkite/agent-stack-k8s/pull/596) changed how agent targeting works. Now, a job's tags must match ALL the agent's tags. 

This behaviour is not the same as how [buildkite/agent](https://github.com/buildkite/agent) works, where it only has to match at least 1 of the agent's tags for it to be picked up.

## To Reproduce

Appears on version >0.28.0 from PR https://github.com/buildkite/agent-stack-k8s/pull/596

## Expected behavior

Jobs request a list of tags, and it will run on agents matching them, but necessarily all of them.

## Environment
- agent-stack-k8s version: [e.g., v1.0.0] 0.28.1
- Kubernetes version: [e.g., v1.24.0] 1.33.0
- Deployment method: [e.g., Helm chart, manual manifest] helm

## Logs
```
job tags do not match expected tags in configuration, skipping...
```

## Additional context

N/A

zhming0:
Solves #625

This PR first of all fixes #625, effectively reverting #596, as the fix was addressing one issue, introducing another.

As a replacement, this PR changed controller so it stops using tags to find correct pods to monitor. Using tags to match pods has a few issue: 
- first and foremost: tags can changes between restart.
- tags matching logic is much more difficult to express using k8s. 
  - Previously the controller is finding pods that has identical tags with controller tags. 
  - But not all pods will have the identical tags -> so this is clearly wrong. 

This PR proposed a new `ID` config variable as an stable identifier for controller, the `ID` will be set by helm during install, otherwise it's optional.

### DIFF
--- a/internal/controller/agenttags/tags.go
+++ b/internal/controller/agenttags/tags.go
@@ -88,8 +88,8 @@ func LabelsFromTags(tags []string) (map[string]string, []error) {
 	return labels, append(errs1, errs2...)
 }
 
-// MatchJobTags reports whether each tag key in `agentTags` is also
-// present in `jobTags`, and the tag value in `jobTags` is either "*" or the
+// MatchJobTags reports whether each tag key in `jobTags` is also
+// present in `agentTags`, and the tag value in `jobTags` is either "*" or the
 // same as the tag value in `agentTags`.
 //
 // In the future, this may be expanded to: if the tag value `agentTags` is in some
@@ -97,15 +97,12 @@ func LabelsFromTags(tags []string) (map[string]string, []error) {
 // See https://buildkite.com/docs/agent/v3/cli-start#agent-targeting
 
 func MatchJobTags(agentTags, jobTags map[string]string) bool {
-	if len(agentTags) != len(jobTags) {
-		return false
-	}
-	for k, v := range agentTags {
-		jobTagValue, exists := jobTags[k]
+	for k, jobTagValue := range jobTags {
+		agentTagValue, exists := agentTags[k]
 		if !exists {
 			return false
 		}
-		if jobTagValue != "*" && jobTagValue != v {
+		if jobTagValue != "*" && jobTagValue != agentTagValue {
 			return false
 		}
 	}
--- a/internal/controller/config/config.go
+++ b/internal/controller/config/config.go
@@ -12,6 +12,7 @@ import (
 
 const (
 	UUIDLabel                             = "buildkite.com/job-uuid"
+	ControllerIDLabel                     = "buildkite.com/controller-id"
 	BuildURLAnnotation                    = "buildkite.com/build-url"
 	JobURLAnnotation                      = "buildkite.com/job-url"
 	PriorityAnnotation                    = "buildkite.com/job-priority"
@@ -55,6 +56,12 @@ type Config struct {
 	EnableQueuePause         bool          `json:"enable-queue-pause"       validate:"omitempty"`
 	WorkQueueLimit           int           `json:"work-queue-limit"         validate:"omitempty"`
 	// Agent endpoint is set in agent-config.
+
+	// ID is an optional uniquely ID string for the controller.
+	// This is useful when running multiple bk k8s controllers within the same k8s namespace.
+	// So the controller can target the correct pods.
+	// By default, if helm is used to install, this will be set as helm release full name.
+	ID string `json:"id" validate:"omitempty"`
 
 	K8sClientRateLimiterQPS   int `json:"k8s-client-rate-limiter-qps" validate:"omitempty"`
 	K8sClientRateLimiterBurst int `json:"k8s-client-rate-limiter-burst" validate:"omitempty"`
@@ -123,6 +130,7 @@ func (c Config) MarshalLogObject(enc zapcore.ObjectEncoder) error {
 	enc.AddInt("job-creation-concurrency", c.JobCreationConcurrency)
 	enc.AddInt("max-in-flight", c.MaxInFlight)
 	enc.AddString("namespace", c.Namespace)
+	enc.AddString("id", c.ID)
 	if err := enc.AddArray("tags", c.Tags); err != nil {
 		return err
 	}
--- a/internal/controller/controller.go
+++ b/internal/controller/controller.go
@@ -144,6 +144,7 @@ func Run(
 	// a pod to run that job. It talks to the k8s API to create pods.
 	sched := scheduler.New(logger.Named("scheduler"), k8sClient, agentClient, scheduler.Config{
 		Namespace:                      cfg.Namespace,
+		ID:                             cfg.ID,
 		Image:                          cfg.Image,
 		AgentTokenSecretName:           cfg.AgentTokenSecret,
 		JobTTL:                         cfg.JobTTL,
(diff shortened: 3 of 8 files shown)


internal/controller/agenttags/tags.go
```diff
@@ -88,24 +88,21 @@ func LabelsFromTags(tags []string) (map[string]string, []error) {
...
 // See https://buildkite.com/docs/agent/v3/cli-start#agent-targeting
 
 func MatchJobTags(agentTags, jobTags map[string]string) bool {
-	if len(agentTags) != len(jobTags) {
-		return false
-	}
-	for k, v := range agentTags {
-		jobTagValue, exists := jobTags[k]
+	for k, jobTagValue := range jobTags {
+		agentTagValue, exists := agentTags[k]
```
DrJosh9000, line 101: This looks like a rollback of #596 - is the bug described in #596 fixed some other way? 

> This was allowing scheduled jobs with only a matching queue tag to be created as a K8s Job. This then prevented the informer from matching the Labels of these Jobs and the controller was left with an infinitely increasing number of running jobs.

Perhaps we should:

* make the Informer listen for objects matching fewer labels 
* put the necessary tag/label matching in our code

DrJosh9000 (changes requested):

internal/controller/agenttags/tags.go
zhming0: @DrJosh9000 I've re-implemented this PR. Another look please 🙏🏿

internal/controller/controller.go
```diff
@@ -238,26 +239,28 @@ func Run(
...
-	for l, v := range labelsFromTags {
-		hasLabel, err := labels.NewRequirement(l, selection.Equals, []string{v})
+	if id != "" {
+		hasRightId, err := labels.NewRequirement(config.ControllerIDLabel, selection.Equals, []string{id})
+		if err != nil {
+			return nil, fmt.Errorf("failed to build controller id label selector for job manager: %w", err)
+		}
+		requirements = append(requirements, *hasRightId)
+	} else {
+		// In the case when Id isn't specified, we shouldn't try to monitor those jobs that has id label.
```
DrJosh9000: Consider adding a similar comment about how the Helm chart set it here:
 ```suggestion
		// In the case when ID isn't specified, we shouldn't try to monitor those jobs that has id label.
		// Note that ID is set by the Helm chart, so this should not happen in usual installations.
```

File: internal/controller/scheduler/scheduler.go [hunk 3]
```diff
@@ -53,6 +53,7 @@ var (
 
 type Config struct {
 	Namespace                      string
+	Id                             string
```
DrJosh9000: As for config/config.go
```suggestion
	ID                             string
```

internal/controller/config/config.go
```diff
@@ -56,6 +57,12 @@ type Config struct {
 	WorkQueueLimit           int           `json:"work-queue-limit"         validate:"omitempty"`
 	// Agent endpoint is set in agent-config.
 
+	// Id is an optional uniquely ID string for the controller.
+	// This is useful when running multiple bk k8s controllers within the same k8s namespace.
+	// So the controller can target the correct pods.
+	// By default, if helm is used to install, this will be set as helm release full name.
+	Id string `json:"id" validate:"omitempty"`
```
DrJosh9000: I typically spell it `ID` rather than `Id` (see also https://google.github.io/styleguide/go/decisions#initialisms)
```suggestion
	ID string `json:"id" validate:"omitempty"`
```

internal/controller/controller.go
```diff
@@ -238,26 +239,29 @@ func Run(
...
 	hasUUID, err := labels.NewRequirement(config.UUIDLabel, selection.Exists, nil)
 	if err != nil {
 		return nil, fmt.Errorf("failed to build uuid label selector for job manager: %w", err)
 	}
 	requirements = append(requirements, *hasUUID)
 
-	for l, v := range labelsFromTags {
-		hasLabel, err := labels.NewRequirement(l, selection.Equals, []string{v})
+	if id != "" {
+		hasRightId, err := labels.NewRequirement(config.ControllerIDLabel, selection.Equals, []string{id})
```
DrJosh9000:
```suggestion
		hasRightID, err := labels.NewRequirement(config.ControllerIDLabel, selection.Equals, []string{id})
```

internal/controller/controller.go
```diff
@@ -238,26 +239,29 @@ func Run(
...
+		if err != nil {
+			return nil, fmt.Errorf("failed to build controller id label selector for job manager: %w", err)
+		}
+		requirements = append(requirements, *hasRightId)
```
DrJosh9000:
```suggestion
		requirements = append(requirements, *hasRightID)
```

internal/controller/controller.go
```diff
@@ -238,26 +239,29 @@ func Run(
...
+	} else {
+		// In the case when ID isn't specified, we shouldn't try to monitor those jobs that has ID label.
+		// Note that ID is set by the Helm chart, so this should not happen in usual installations.
+		noId, err := labels.NewRequirement(config.ControllerIDLabel, selection.DoesNotExist, nil)
```
DrJosh9000:
```suggestion
		noID, err := labels.NewRequirement(config.ControllerIDLabel, selection.DoesNotExist, nil)
```

internal/controller/controller.go
```diff
@@ -238,26 +239,29 @@ func Run(
...
 		if err != nil {
-			return nil, fmt.Errorf("failed create label selector agent tag: %w", err)
+			return nil, fmt.Errorf("failed to build controller id label selector for job manager: %w", err)
 		}
-		requirements = append(requirements, *hasLabel)
+		requirements = append(requirements, *noId)
```
DrJosh9000:
```suggestion
		requirements = append(requirements, *noID)
```

internal/controller/scheduler/scheduler_test.go
```diff
@@ -539,6 +540,9 @@ func TestBuild(t *testing.T) {
 
 	require.Len(t, kjob.Spec.Template.Spec.Containers, 3)
 
+	controllerIdLabel := kjob.Labels["buildkite.com/controller-id"]
+	assert.Equal(t, controllerIdLabel, "controller-1")
```
DrJosh9000:
```suggestion
	controllerIDLabel := kjob.Labels["buildkite.com/controller-id"]
	assert.Equal(t, controllerIDLabel, "controller-1")
```

DrJosh9000 (approved):
LGTM!

My remaining suggestions are just quick follow-ons about `ID`.

---

Do you have any ideas for how we could write a test to cover the issue originally fixed by #596?

zhming0:
@DrJosh9000 that's good question. I could have a misunderstood the symptom of #596, I thought our `TestMaxInFlightLimited` should catch the issue but apparently `TestMaxInFlightLimited` was added long before the issue exist. 

Do you have more contexts? (happy to move this chat to slack if you prefer too)

As for this PR, I think the change is sound, I will merge this first and maybe come back with a test later.

PR #627, opened 2025-06-17
Stack error signal reason

moskyb:
THis PR makes it so that when the k8s stack fails a job (outside of the normal flow of something in the job exiting with a non-zero status), eg when a job can't be launched due to the image not being pullable, the job will complete with a `stack_error` signal reason.

This will allow users to define specific retry policies for build failures caused by underlying infrastructure issues in the k8s stack.


zhming0 (approved):
I can't think of another case where controller will fail a job for non-stack error, so I think this a 👍🏿

Issue #575, opened 2025-04-28
[BUG] Buildkite not passing through error codes from Pods


Blaze2305:
## Describe the bug
Error codes from pods aren't being passed through the agent , preventing the use of exit code based retries.

When we see that pods fail to get scheduled, we see

```
The following init containers failed:
 
 CONTAINER   EXIT CODE  SIGNAL  REASON                  MESSAGE                                                        
 copy-agent        137       0  ContainerStatusUnknown  The container could not be located when the pod was terminated
```

in the buildkite UI. The exit code emitted however was 1. We have retries configured for such scenarios where we get exit code 137, but due to the improper exit code we cannot auto retry which causes the build to fail entirely.

## To Reproduce

One way would be to induce memory pressure on the node and have the queue pick up a job, when the container fails and exits with code 137, BK will exit with code 1

## Expected behavior
If the container fails with exit code X, the agent should also emit the same code X without changes

## Environment
- agent-stack-k8s version: v0.25.0
- Kubernetes version: v1.29

## Logs
```
The following init containers failed:
 
 CONTAINER   EXIT CODE  SIGNAL  REASON                  MESSAGE                                                        
 copy-agent        137       0  ContainerStatusUnknown  The container could not be located when the pod was terminated
```

rohitraut3366:
Additionally, the Buildkite job details are displaying the controller hostname instead of the pod or job name.
```
buildkite-0196947a-6c78-45c2-94e9-8960495bcbfe  
Accepted Job  
Today at 10:21:06.594 (+32ms)  
Host name: bk-rippling-main-master-3-agent-stack-k8s-8f4597db9-659wm  
Operating System: Wolfi 20230201  
Agent Version: 3.95.1  
```

rohitraut3366:
Additionally, the Buildkite job details are displaying the controller hostname instead of the pod . make it hard to troubleshoot
```
Accepted Job  
Today at 10:21:06.594 (+32ms)  
Host name: bk-controller  
Operating System: Wolfi 20230201  
Agent Version: 3.95.1  
```

sj26:
I think we'll do a mixture of things:

1. We'll add a signal reason to identify stack-level errors (before the agent or its job can be started) to enable broad stack-level retry rules
2. When a pod is created and containers start running but the agent and its job do not start then we'll propagate the first non-zero container exit code (with that stack error signal reason), allowing more targeted retry rules for certain stack errors

sj26:
[...]

PR #628, opened 2025-06-18
PB-5: improve exit code propogation for non-command containers

zhming0:
This PR slightly change how we propagate non-command containers error to bk api. 

* In initContainer failure, we will pass the first non-0 exit code from init containers back to backend. 
* When scheduler yaml failure, we will raise a `AgentRefused` error reason -> this can be helpful to indicate that this is not an retryable error. 

solves #575 
solves PB-7 and PB-5

### DIFF
--- a/internal/controller/scheduler/fail_job.go
+++ b/internal/controller/scheduler/fail_job.go
@@ -1,6 +1,7 @@
 package scheduler
 
 import (
+	"cmp"
 	"context"
 	"errors"
 	"fmt"
@@ -19,6 +20,13 @@ import (
 	"k8s.io/client-go/kubernetes"
 )
 
+// FailureInfo contains data about a job failure for reporting to Buildkite.
+type FailureInfo struct {
+	Message  string
+	ExitCode int32
+	Reason   string
+}
+
 // acquireAndFailForObject figures out how to fail the BK job corresponding to
 // the k8s object (a pod or job) by inspecting the object's labels.
 func acquireAndFailForObject(
@@ -27,7 +35,7 @@ func acquireAndFailForObject(
 	k8sClient kubernetes.Interface,
 	cfg *config.Config,
 	obj metav1.Object,
-	message string,
+	failureInfo FailureInfo,
 ) error {
 	agentToken, err := fetchAgentToken(ctx, logger, k8sClient, obj.GetNamespace(), cfg.AgentTokenSecret)
 	if err != nil {
@@ -45,7 +53,7 @@ func acquireAndFailForObject(
 	tags := agenttags.TagsFromLabels(labels)
 	opts := cfg.AgentConfig.ControllerOptions()
 
-	if err := acquireAndFail(ctx, logger, agentToken, cfg.JobPrefix, jobUUID, tags, message, opts...); err != nil {
+	if err := acquireAndFail(ctx, logger, agentToken, cfg.JobPrefix, jobUUID, tags, failureInfo, opts...); err != nil {
 		logger.Error("failed to acquire and fail the job on Buildkite", zap.Error(err))
 		return err
 	}
@@ -61,7 +69,7 @@ func acquireAndFail(
 	jobPrefix string,
 	jobUUID string,
 	tags []string,
-	message string,
+	failureInfo FailureInfo,
 	options ...agentcore.ControllerOption,
 ) error {
 	opts := append([]agentcore.ControllerOption{
@@ -89,13 +97,19 @@ func acquireAndFail(
 		return fmt.Errorf("starting job: %w", err)
 	}
 
-	if err := jctr.WriteLog(ctx, message); err != nil {
+	if err := jctr.WriteLog(ctx, failureInfo.Message); err != nil {
 		zapLogger.Error("writing log", zap.Error(err))
 		return fmt.Errorf("writing log: %w", err)
 	}
 
 	var ignoreAgentInDispatches *bool
-	if err := jctr.Finish(ctx, agentcore.ProcessExit{Status: 1, SignalReason: agent.SignalReasonStackError}, ignoreAgentInDispatches); err != nil {
+
+	failureInfo.ExitCode = cmp.Or(failureInfo.ExitCode, 1)
+	// By default, we consider all failure triggered by the stack controller are stack errors
+	failureInfo.Reason = cmp.Or(failureInfo.Reason, string(agent.SignalReasonStackError))
+
+	processExit := agentcore.ProcessExit{Status: int(failureInfo.ExitCode), SignalReason: failureInfo.Reason}
+	if err := jctr.Finish(ctx, processExit, ignoreAgentInDispatches); err != nil {
 		zapLogger.Error("finishing job", zap.Error(err))
 		return fmt.Errorf("finishing job: %w", err)
 	}
--- a/internal/controller/scheduler/job_watcher.go
+++ b/internal/controller/scheduler/job_watcher.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/buildkite/agent-stack-k8s/v2/internal/controller/config"
 	"github.com/buildkite/agent-stack-k8s/v2/internal/controller/model"
+	"github.com/buildkite/agent/v3/agent"
 
 	"github.com/google/uuid"
 	"github.com/jedib0t/go-pretty/v6/table"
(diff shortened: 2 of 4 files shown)


internal/controller/scheduler/fail_job.go
```diff
@@ -19,15 +19,22 @@ import (
 	"k8s.io/client-go/kubernetes"
 )
 
+type FailureContext struct {
```
DrJosh9000: This doesn't seem like a _context_ - message, exit code, and reason seem like data that directly concerns a failure, unlike contextual data. `FailureInfo`?

```suggestion
// FailureInfo contains data about a job failure for reporting to Buildkite.
type FailureInfo struct {
```

internal/controller/scheduler/fail_job.go
```diff
@@ -19,15 +19,22 @@ import (
...
+	Message  string
+	ExitCode int32
+	Reason   string
+}
+
 // acquireAndFailForObject figures out how to fail the BK job corresponding to
 // the k8s object (a pod or job) by inspecting the object's labels.
+// failureInfo can be a string or a FailureContext.
```
DrJosh9000: This line doesn't make sense to me. 

```suggestion
```

internal/controller/scheduler/fail_job.go
```diff
@@ -19,15 +19,22 @@ import (
...
 func acquireAndFailForObject(
 	ctx context.Context,
 	logger *zap.Logger,
 	k8sClient kubernetes.Interface,
 	cfg *config.Config,
 	obj metav1.Object,
-	message string,
+	failureCtx *FailureContext,
```
DrJosh9000: It's necessary, and the struct size is not very "big". So probably better to not pass as a pointer?

DrJosh9000 (commented):
Some initial thoughts...

internal/controller/scheduler/fail_job.go
```diff
@@ -27,7 +34,7 @@ func acquireAndFailForObject(
 	k8sClient kubernetes.Interface,
 	cfg *config.Config,
 	obj metav1.Object,
-	message string,
+	failureCtx FailureInfo,
```
DrJosh9000: If it's not a "context", then it ought not to have `Ctx` in the name...
```suggestion
	info FailureInfo,
```

internal/controller/scheduler/fail_job.go
```diff
@@ -45,7 +52,7 @@ func acquireAndFailForObject(
 	tags := agenttags.TagsFromLabels(labels)
 	opts := cfg.AgentConfig.ControllerOptions()
 
-	if err := acquireAndFail(ctx, logger, agentToken, cfg.JobPrefix, jobUUID, tags, message, opts...); err != nil {
+	if err := acquireAndFail(ctx, logger, agentToken, cfg.JobPrefix, jobUUID, tags, failureCtx, opts...); err != nil {
```
DrJosh9000:
```suggestion
	if err := acquireAndFail(ctx, logger, agentToken, cfg.JobPrefix, jobUUID, tags, info, opts...); err != nil {
```

internal/controller/scheduler/fail_job.go
```diff
@@ -61,7 +68,7 @@ func acquireAndFail(
 	jobPrefix string,
 	jobUUID string,
 	tags []string,
-	message string,
+	failureCtx FailureInfo,
```
DrJosh9000:
```suggestion
	info FailureInfo,
```

internal/controller/scheduler/fail_job.go
```diff
@@ -89,13 +96,24 @@ func acquireAndFail(
...
+
+	status := failureCtx.ExitCode
+	if status == 0 {
+		status = 1
+	}
+	reason := failureCtx.Reason
+	if reason == "" {
+		// By default, we consider all failure triggered by the stack controller are stack errors
+		reason = agent.SignalReasonStackError
+	}
```
DrJosh9000: [totally optional suggestion] Now that it's not a pointer, there's no need to avoid mutating the input:

```suggestion
	if info.ExitCode == 0 {
		info.ExitCode = 1
	}
	if info.Reason == "" {
		// By default, we consider all failure triggered by the stack controller are stack errors
		info.Reason = agent.SignalReasonStackError
	}
```

I've been toying with the idea of using `cmp.Or` more often, but I'm undecided on readability:

```suggestion
	info.ExitCode = cmp.Or(info.ExitCode, 1)
	// By default, we consider all failure triggered by the stack controller are stack errors
	info.Reason = cmp.Or(info.Reason, agent.SignalReasonStackError)
```

internal/controller/scheduler/job_watcher.go
```diff
@@ -231,7 +232,12 @@ func (w *jobWatcher) fetchEvents(ctx context.Context, log *zap.Logger, kjob *bat
 }
 
 func (w *jobWatcher) failJob(ctx context.Context, log *zap.Logger, kjob *batchv1.Job, message string) {
-	if err := acquireAndFailForObject(ctx, log, w.k8s, w.cfg, kjob, message); err != nil {
+	failureCtx := FailureInfo{
```
DrJosh9000:
```suggestion
	info := FailureInfo{
```

internal/controller/scheduler/pod_watcher.go
```diff
@@ -322,7 +325,12 @@ func (w *podWatcher) failOnInitContainerFailure(ctx context.Context, log *zap.Lo
 	// probably shouldn't interfere.
 	log.Info("One or more init containers failed. Failing.")
 	message := w.formatInitContainerFails(containerFails)
-	if err := acquireAndFailForObject(ctx, log, w.k8s, w.cfg, pod, message); err != nil {
+	failureCtx := FailureInfo{
```
DrJosh9000:
```suggestion
	info := FailureInfo{
```

internal/controller/scheduler/pod_watcher.go
```diff
@@ -477,7 +485,11 @@ func (w *podWatcher) failForImageFailure(ctx context.Context, log *zap.Logger, f
 		// We can acquire it and fail it ourselves.
 		log.Info("One or more job containers are waiting too long for images. Failing.")
 		message := w.formatImagePullFailureMessage(statuses)
-		if err := acquireAndFailForObject(ctx, log, w.k8s, w.cfg, pod, message); err != nil {
+		failureCtx := FailureInfo{
```
DrJosh9000:
```suggestion
		info := FailureInfo{
```

internal/controller/scheduler/scheduler.go
```diff
@@ -1272,7 +1273,14 @@ func (w *worker) failJob(ctx context.Context, inputs buildInputs, message string
 	}
 
 	opts := w.cfg.AgentConfig.ControllerOptions()
-	if err := acquireAndFail(ctx, w.logger, agentToken, w.cfg.JobPrefix, inputs.uuid, inputs.agentQueryRules, message, opts...); err != nil {
+	failureCtx := FailureInfo{
```
DrJosh9000:
```suggestion
	info := FailureInfo{
```

DrJosh9000 (commented):
This is pretty good

internal/controller/scheduler/fail_job.go
zhming0: I like `cmp.Or`, it feels more statement-like and declarative so it's good.

internal/controller/scheduler/scheduler.go
```diff
@@ -1272,7 +1273,14 @@ func (w *worker) failJob(ctx context.Context, inputs buildInputs, message string
 	}
 
 	opts := w.cfg.AgentConfig.ControllerOptions()
-	if err := acquireAndFail(ctx, w.logger, agentToken, w.cfg.JobPrefix, inputs.uuid, inputs.agentQueryRules, message, opts...); err != nil {
+	failureInfo := FailureInfo{
+		Message: message,
+		// In scheduler worker, often these failure are technically users error in YAML.
+		// Using agent refuse seems to be reasonable as that represent that agent won't execute these yaml.
+		// Deserve more discussion though..
+		Reason: agent.SignalReasonAgentRefused,
```
zhming0: @DrJosh9000 I think this change is debatable, subject to our confidence level. Just wanting to highlight this in case in you have opinion here

DrJosh9000: Hmm. I think the cases here where `failJob` is called are all "pre-agent", so using "agent refused" feels wrong for that reason. We also can't rule out the Kubernetes cluster having a bad day and rejecting a valid job. "Stack error" seems like the best choice for now.

Maybe the agent change should have included a "stack rejected" reason, as well as "stack error"?

zhming0: Ah true, the controller isn't an agent. I changed it to stack error but with a bit comments explaining the caveats. 🙏🏿

DrJosh9000 (approved):
LGTM!

PR #636, opened 2025-06-24
go.mod: remove mapstructure replace

DrJosh9000:
This replace directive was originally added in https://github.com/buildkite/agent-stack-k8s/commit/8f253f55b807acf309c112157a9e6df235c62552 to address a podSpecPatch implementation issue caused by unmarshaling Kubernetes structures in the config via mapstructure, needed by Viper. 

Viper has since adopted mapstructure, so Viper is no longer using the mitchellh original. So the replace should have no effect. I believe the squash tag patch made it into Viper's fork - if it didn't, I presume we would have heard about it by now.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,8 +4,6 @@ go 1.24.0
 
 toolchain go1.24.1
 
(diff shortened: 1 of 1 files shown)


zhming0 (approved):
🚀

PR #641, opened 2025-07-01
PIPE-1263: set user-agent header

zhming0:
This PR ensures that we set a consistent user-agent header for all outbound http requests.


DrJosh9000 (approved):
I'm a little sad it won't have a version number, but you're right, we can kick it down the road.

PR #642, opened 2025-07-07
PB-164 part 1: relocating imagecheck related logic to another file

zhming0:
This PR does a small scale refactor, moving image check related logics from `scheduler.go` to `imagecheck.go`. It aligns with our general direction of improving `scheduler` maintainability.

No logic change.

### DIFF
--- a/internal/controller/scheduler/imagecheck.go
+++ b/internal/controller/scheduler/imagecheck.go
@@ -1,0 +1,233 @@
+package scheduler
+
+import (
+	"cmp"
+	"strconv"
+
+	"github.com/buildkite/agent-stack-k8s/v2/internal/controller/config"
+	"github.com/distribution/reference"
+	"go.uber.org/zap"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+)
+
+// Rank pull policy by preference. If two containers specify different pull
+// policies for the same image, the most preferred policy is used for the
+// image check.
+// Always is most preferred, Never is less preferred.
+var pullPolicyPreference = map[corev1.PullPolicy]int{
+	corev1.PullNever:        1, // least preferred
+	corev1.PullIfNotPresent: 2,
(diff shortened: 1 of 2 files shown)


DrJosh9000 (commented):
Good idea, there's a bunch of fiddly logic here that gets in the way of the overall scheduler.

internal/controller/scheduler/imagecheck.go
```diff
@@ -0,0 +1,234 @@
...
+)
+
+// Use init containers to check that images can be used or pulled before
+// any other containers run. These are added _after_ podSpecPatch is applied
+// since podSpecPatch may freely modify each container's image ref or add
+// more containers.
+// This process also checks that image refs are well-formed so we can fail
+// the job early.
+// Init containers run before the agent, so we can acquire and fail the job
+// if an init container stays in ImagePullBackOff for too long.
```
DrJosh9000: This section of comments is the rationale for using init containers for checking image pull and an overview of how it works and why they are added when they are added. So it would be more useful to someone reading scheduler.go.

internal/controller/scheduler/imagecheck.go
```diff
@@ -0,0 +1,234 @@
...
+				policy: nominalPolicy,
+				ref:    ref,
+			}
+		}
+	}
+
+	return preflightImageChecks, invalidRefs
+}
+
+func cullCheckForExisting(
```
DrJosh9000, line 143: The reason for `cullCheckForExisting` seems to be missing. The existing comment wasn't very strong, so here's my attempt:

```suggestion
// cullCheckForExisting removes image-check containers with images that are
// already present among existing init containers.
// Such image-check containers don't need to be added, since the image pull
// failure checking happens across all init containers. Adding another init
// container specifically for checking image pull would be redundant.
// However, if a stronger pull policy is specified for a non-init container
// than for an existing init container, then an image check may still be
// needed.
func cullCheckForExisting(
```

PR #643, opened 2025-07-07
PB-164 part 2: add opt-out flag for imagecheck containers

zhming0:
Allow image check containers to be opted out.

based on top of #642.

### DIFF
--- a/charts/agent-stack-k8s/values.schema.json
+++ b/charts/agent-stack-k8s/values.schema.json
@@ -589,6 +589,12 @@
             }
           }
         },
+        "skip-image-check-containers": {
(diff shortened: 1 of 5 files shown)


charts/agent-stack-k8s/values.schema.json
```diff
@@ -589,6 +589,12 @@
             }
           }
         },
+        "skip-image-check-containers": {
+          "type": "string",
```
DrJosh9000:
```suggestion
          "type": "boolean",
```

DrJosh9000 (approved):

PR #650, opened 2025-07-15
PB-140 part 1: prototype Reserver job handler

zhming0:
This PR does not change any stack behavior, it already prepares a directional structure for our upcoming Rerserved state support. 

It's part of PB-140, subsequent PRs will fill gap and bring test cases when the backend is ready.


api/agent_client.go
```diff
@@ -199,6 +199,11 @@ func (c *AgentClient) GetJobState(ctx context.Context, id string) (result *Agent
 	return decodeResponse[AgentJobState](resp)
 }
 
+func (c *AgentClient) ReserverJob(ctx context.Context, id string) error {
```
DrJosh9000: "reserve" as a verb, here

```suggestion
func (c *AgentClient) ReserveJob(ctx context.Context, id string) error {
```

We should also strongly consider reserving multiple jobs in batches.

api/agent_client.go
```diff
@@ -199,6 +199,11 @@ func (c *AgentClient) GetJobState(ctx context.Context, id string) (result *Agent
...
+	// TODO: implement me as subsequent PRs in PB-140
+	return nil
```
DrJosh9000: Worth doing this?

```suggestion
	return errors.New("unimplemented")
```

internal/controller/model/model.go
```diff
@@ -18,6 +18,9 @@ var ErrDuplicateJob = errors.New("job already scheduled")
 // begin scheduling.
 var ErrStaleJob = errors.New("stale-job-data-timeout")
 
+// ErrUnableToReserverJob is a sentinel error returned when job has already been reserved or assigned.
+var ErrUnableToReserverJob = errors.New("unable-to-reserve-job")
```
DrJosh9000:
```suggestion
// ErrUnableToReserveJob is a sentinel error returned when job has already been reserved or assigned.
var ErrUnableToReserveJob = errors.New("unable to reserve job")
```

I'm not sure why ErrStaleJob's value has dashes in it.

internal/controller/controller.go
```diff
@@ -119,10 +120,10 @@ func Run(
 		return
 	}
 
-	// **************************************************************
-	// ***                        JOB FLOW                        ***
-	// ***       Monitor -> Limiter -> Deduper -> Scheduler       ***
-	// **************************************************************
+	// **************************************************************************
+	// ***                        JOB FLOW                                    ***
+	// ***       Monitor -> Limiter -> Deduper -> Reserver -> Scheduler       ***
```
DrJosh9000: I guess I have a more fundamental question: why put the Reserver in the chain near the end, rather than closer to the start (monitor -> reserver -> limiter)? It would be cool to see "job has been acknowledged by the controller, but waiting for jobs to finish because of max-in-flight" in the UI, but that implies reserving the job straight away.

zhming0: I put a fair bit of thoughts into this too. I left some comments trying to explain why: 

```
        // There is an argument to make this run straight after the bk job monitor, before limiter and deduper
	// but the queue setup makes it a bit tricky -> limiter will drop message after work queue reach max.
	// We should only reserve things when we are relatively confident that we can schedule it.
```

TLDR, the work queue implementation is in the way. But, we could rely on the backend expirer to make this work, is that ideal though? not sure.

I think there is also an argument to support that reservation should come right before scheduling, because:

-  if the controller crashed after reservation, any jobs that aren't scheduled will essentially be facing reservation timeout. 
- Once the job are handled to limiter, there is no guarantee on latency, so it could be that the reservation has expired before hitting scheduler -> this introduce a new error surface. 


My stance on this issue isn't very strong, open to your opinion. 

(This might be something that's worth a sync chat?)

zhming0: (One benefit of bringing it to before limiter is that it can be a ManyJobHandler, which makes the batching slightly easier to implement. )

zhming0: As I was thinking more while replying to you, maybe we should consider reservation expiration a normal thing, in that case I'd be in support of putting `Reserver` before `Limiter`

DrJosh9000: If the controller crashes, we can probably release the reserved jobs on the backend if we can detect a new controller has started for the queue.

Regarding the reservation expiry, wouldn't that also apply once the job has been scheduled and is spinning up? We should consider implementing a way to regularly post reservation updates for a job, wherever it is in the pipeline.

zhming0: > we can probably release the reserved jobs on the backend 

Hmm, backend wouldn't have very good visibility on controller's state yet. 

But I see where you are going, we can use short expiration and regular update, this will justify putting Reserver in the beginning. And I agree with this: 

> wouldn't that also apply once the job has been scheduled and is spinning up

So I guess, if the expiration extension mechanism is necessary, I support putting the Reserver in the beginning. 

---- 

The job dropping mechanism is still in the way though -- because subsequent handler can drop jobs, we would have to carefully design the state flow, so we can deterministically know if a job is in pipeline or not. 

One option might be introducing a new `ReservationUpdater` interface for each stateful `JobHandler/MultiJobHandler/Watcher` (scheduler, deduper, limiter, job/pod watcher), so we can probably have a central goroutine to periodically invoke each handler to report back internal pending jobs. 

~I will give it a go, but let me know what you think 🙏🏿~

Still thinking...

zhming0: anyway the update mechanism is probably out of scope for PB-140, good to have these thinking though, I will focus on moving Reserver to the beginning of the chain first per our agreement here 🚂

api/agent_client.go
zhming0: I've updated it be `ReserveJobs` per @moskyb's incoming change.

api/agent_client.go
zhming0: I added a dummy implementation instead, I don't think it matter very much, since I am going to replace this bit in the next PR 🚀

internal/controller/controller.go
zhming0: ready for another look 🙏🏿

api/agent_client.go
sj26: Yeah batch reserve needs to exist, but can be another slice

sj26: (e.g. our partners will likely want batch reserve)

internal/controller/controller.go
sj26: Love the thinking here you two

api/agent_client.go
zhming0: It's batch reserve already 🙂 😁

sj26: Oh man I can't keep up 😅 👏🏻

api/agent_client.go
```diff
@@ -199,6 +199,18 @@ func (c *AgentClient) GetJobState(ctx context.Context, id string) (result *Agent
 	return decodeResponse[AgentJobState](resp)
 }
 
+type MassReserveJobsResult struct {
```
DrJosh9000: We've tended to call these things "Batch" rather than "Mass"

Compare with e.g. https://github.com/buildkite/agent/blob/dcd1e6af5381427bc4627a3615cafb6a1e0b74be/api/artifacts.go#L53

Also, please add doc comments:

```suggestion
// ReserveJobBatchResult describes the result of a batch job reservation.
type ReserveJobBatchResult struct {
```

api/agent_client.go
```diff
@@ -199,6 +199,18 @@ func (c *AgentClient) GetJobState(ctx context.Context, id string) (result *Agent
...
+	ReservedJobUUIDs    []string `json:"reserved"`
+	NotReservedJobUUIDs []string `json:"not_reserved"`
+}
+
+func (c *AgentClient) ReserveJobs(ctx context.Context, ids []string) (result *MassReserveJobsResult, err error) {
```
```suggestion
// ReserveJobs reserves a batch of jobs.
func (c *AgentClient) ReserveJobs(ctx context.Context, ids []string) (result *MassReserveJobsResult, err error) {
```

DrJosh9000 (commented):
Looking pretty good! Some suggestions

api/agent_client.go
zhming0: ✅

internal/controller/reserver/reserver.go
```diff
@@ -0,0 +1,91 @@
...
+		enabled:     enabled,
+	}
+
+	return r
+}
+
+// Pause pauses (or un-pauses) the queue.
+func (r *Reserver) Pause(pause bool) {
+	r.paused = pause
+}
```
DrJosh9000: I think we should pass the call onto the next handler too:

```suggestion
func (r *Reserver) Pause(pause bool) {
	r.handler.Pause(pause)
	r.paused = pause
}
```

Because the Limiter has a queue in it, and when paused should stop sending jobs down the pipeline.

internal/controller/reserver/reserver.go
```diff
@@ -0,0 +1,91 @@
...
+	agentClient *api.AgentClient
+
+	// Logs goes here
+	logger *zap.Logger
+
+	enabled bool
+	paused  bool
+}
+
+func New(logger *zap.Logger, agentClient *api.AgentClient, nextHanler model.ManyJobHandler, enabled bool) *Reserver {
```
DrJosh9000:
```suggestion
func New(logger *zap.Logger, agentClient *api.AgentClient, nextHandler model.ManyJobHandler, enabled bool) *Reserver {
```

DrJosh9000 (approved):
LGTM with just two more changes

internal/controller/reserver/reserver.go
zhming0: Good catch 👍🏿

PR #651, opened 2025-07-17
PB-140 part 2: job reservation support

zhming0:
This is based on top of #650, can only be merged after that. 

In addition, this shall not be merged before the backend counter part been deployed https://github.com/buildkite/buildkite/pull/24130. 

This PR also enables the experiment for test env and local dev env.

-----------

This PR filled the gap in #650, making actual reserve api call to backend.

### DIFF
--- a/api/agent_client.go
+++ b/api/agent_client.go
@@ -17,7 +17,15 @@ import (
 // This is a special keyword supported on backend for polling from the current default queue in the cluster.
 const defaultQueueKey = "_default"
 
-func NewAgentClient(token, endpoint, clusterID, queue string, agentQueryRules []string) (*AgentClient, error) {
+type AgentClientOption func(*AgentClient)
+
+func WithReservation(enable bool) AgentClientOption {
+	return func(c *AgentClient) {
+		c.reservation = enable
+	}
+}
+
+func NewAgentClient(token, endpoint, clusterID, queue string, agentQueryRules []string, opts ...AgentClientOption) (*AgentClient, error) {
 	if endpoint == "" {
 		endpoint = "https://agent.buildkite.com/v3"
 	}
@@ -31,7 +39,7 @@ func NewAgentClient(token, endpoint, clusterID, queue string, agentQueryRules []
 	if queue == "" {
 		queue = defaultQueueKey
 	}
-	return &AgentClient{
+	client := &AgentClient{
 		endpoint: endpointURL,
 		httpClient: &http.Client{
 			Timeout:   60 * time.Second,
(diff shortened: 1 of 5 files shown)


justfile
```diff
@@ -28,6 +28,9 @@ test *FLAGS:
 
   GIT_BRANCH=$(git rev-parse --abbrev-ref HEAD)
 
+  # FIXME Enable me once the backend counterpart is ready
+  # export EXPERIMENTAL_JOB_RESERVATION_SUPPORT=true
```
zhming0: I will uncomment this before merge.

api/agent_client.go
```diff
@@ -206,11 +214,38 @@ type BatchReserveJobsResult struct {
 }
 
 // ReserveJobs reserves a batch of jobs.
-func (c *AgentClient) ReserveJobs(ctx context.Context, ids []string) (result *BatchReserveJobsResult, err error) {
-	// TODO: implement me as subsequent PRs in PB-140
-	return &BatchReserveJobsResult{
-		ReservedJobUUIDs: ids,
-	}, nil
+func (c *AgentClient) ReserveJobs(ctx context.Context, ids []string) (result *BatchReserveJobsResult, retryAfter time.Duration, err error) {
+	u := c.endpoint.JoinPath(
```
DrJosh9000, line 233: Should we defensively guard against not having reservation enabled in this method?

e.g.

```suggestion
	if !c.reservationSupport {
		return nil, 0, errors.New("reservation not enabled")
	}

	u := c.endpoint.JoinPath(
```

api/agent_client.go
```diff
@@ -17,7 +17,7 @@ import (
 // This is a special keyword supported on backend for polling from the current default queue in the cluster.
 const defaultQueueKey = "_default"
 
-func NewAgentClient(token, endpoint, clusterID, queue string, agentQueryRules []string) (*AgentClient, error) {
+func NewAgentClient(token, endpoint, clusterID, queue string, agentQueryRules []string, reseverationSupport bool) (*AgentClient, error) {
```
```suggestion
func NewAgentClient(token, endpoint, clusterID, queue string, agentQueryRules []string, reservationSupport bool) (*AgentClient, error) {
```

This could be the right time to introduce functional options, e.g.

```go
type ClientOption func(*AgentClient)

func WithReservation(enable bool) ClientOption {
	return func(c *AgentClient) {
		c.reservation = enable
	}
}

func NewAgentClient(token, endpoint, clusterID, queue string, agentQueryRules []string, opts ...ClientOption) (*AgentClient, error) {

```

(Endpoint would be another good one to make optional)

api/agent_client.go
```diff
@@ -48,6 +49,8 @@ type AgentClient struct {
 	httpClient *http.Client
 	clusterID  string // or "unclustered"
 	queue      string
+	// This impacts a number of endpoints' query parameters
+	reservationSupport bool
```
DrJosh9000: Maybe this should be `reservation` or `enableReservation` or similar? To me it feels like "support" for reservation is this code change that adds, whether to use it or not is an option.

```suggestion
	reservation bool
```

DrJosh9000 (approved):
LGTM!

PR #653, opened 2025-07-17
PS-906: allow configuring agent bootstap timeout

zhming0:
This PR is relying on https://github.com/buildkite/agent/pull/3398, can only be merged after that. 

it allows a `bootstrap-timeout` config under agent config, this config can be used to let children agent containers wait longer for agent "server" container (the one that talks to agent job api) to get ready. 

close PS-906

(TODO: I intend to test this after https://github.com/buildkite/agent/pull/3398 being deployed, but ready for accepting reviews)


zhming0:
with some new discovery, I think this makes sense. I will put it back to draft.

PR #658, opened 2025-07-23
Upgrade agent version to 3.103.0, change code owners

zhming0:
Upgrade agent so we can configure bootstrap connection timeout and get the latest if-changed change. 

Change code owner.


.github/CODEOWNERS
DrJosh9000, line 1: Thanks for doing this - my only issue is that fixing this is unrelated to updating the agent version.

zhming0: I know. 😬 

Would you prefer a standalone PR?

DrJosh9000: Probably should?

PR #659, opened 2025-07-28
PIPE-1664: job level image attribute take precedence over controller podSpecPatch

zhming0:
We want precedence order goes 

1. Job level plugin spec overrides 👇🏿 
2. Job level individual attribute, like `image`, overrides 👇🏿 
3. Controller level config.  

This PR fixes a situation where controller level config will override job level `image` attribute.

### DIFF
--- a/internal/controller/scheduler/build_inputs.go
+++ b/internal/controller/scheduler/build_inputs.go
@@ -1,0 +1,22 @@
+package scheduler
+
+import v1 "k8s.io/api/core/v1"
+
+func applyCustomImageIfPresent(podSpec *v1.PodSpec, inputs *buildInputs) *v1.PodSpec {
+
+	customImage := inputs.envMap["BUILDKITE_IMAGE"]
+	if customImage == "" {
+		return podSpec
+	}
+
+	for i := range podSpec.Containers {
+		c := &podSpec.Containers[i]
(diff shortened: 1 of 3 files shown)


internal/controller/scheduler/build_inputs.go
```diff
@@ -0,0 +1,22 @@
...
+		return podSpec
+	}
+
+	for i := range podSpec.Containers {
+		c := &podSpec.Containers[i]
+		if c.Name != CommandCommanderName {
+			continue
+		}
+
+		c.Image = inputs.envMap["BUILDKITE_IMAGE"]
```
matthewborden:
```suggestion
		c.Image = customImage
```

internal/controller/scheduler/scheduler.go
```diff
@@ -43,6 +43,7 @@ const (
 	CopyAgentContainerName        = "copy-agent"
 	ImageCheckContainerNamePrefix = "imagecheck-"
 	CheckoutContainerName         = "checkout"
+	CommandCommanderName          = "container-0"
```
matthewborden:
```suggestion
	CommandContainerName          = "container-0"
```

zhming0: 🤦🏿 Good catch!

PR #663, opened 2025-07-30
Return token on context cancellation in worker

JoeColeman95:
Adds a call to l.tryReturnToken("Handle") when the worker exits due to context cancellation, ensuring tokens are properly returned and resource accounting remains accurate


DrJosh9000 (approved):
Thanks!

PR #668, opened 2025-08-07
Introduce resource_class

zhming0:
# Resource Classes

Resource Classes allow you to define reusable resource configurations that can be applied to CI workloads based on agent tags.

## Configuration

Resource classes are defined in the controller configuration under the `resource-classes` key:

```yaml
resource-classes:
  class-name:
    resource:      # Optional: Kubernetes resource requirements
      requests:
        cpu: "100m"
        memory: "128Mi"
      limits:
        cpu: "200m"
        memory: "256Mi"
    nodeSelector:  # Optional: Kubernetes node selector
      instance-type: "small"
      zone: "us-west-2a"
```

### Fields

- **resource** (optional): Kubernetes `ResourceRequirements` object that will be applied to the command container
  - **requests**: Resource requests (CPU, memory, etc.)
  - **limits**: Resource limits (CPU, memory, etc.)
- **nodeSelector** (optional): Key-value pairs for Kubernetes node selection

## Usage

To use a resource class in your CI pipeline, specify the `resource_class` agent tag:

```yaml
steps:
  - label: "Build"
    command: "make build"
    agents:
      resource_class: "medium"
      queue: "default"
```

## Behavior

When a job specifies a `resource_class` agent tag:

1. The scheduler looks up the resource class in the controller configuration
2. If found, it applies the resource requirements and node selector to the pod spec
3. The `resource_class` tag is added to the spawned agent process
4. If the resource class is not found, a warning is logged and the job continues with default settings

### Small Resource Class
```yaml
resource-classes:
  xs:
    resource:
      requests:
        cpu: "100m"
        memory: "128Mi"
      limits:
        cpu: "200m"
        memory: "256Mi"
```

### GPU Resource Class
```yaml
resource-classes:
  gpu:
    resource:
      requests:
        nvidia.com/gpu: "1"
        cpu: "1000m"
        memory: "2Gi"
      limits:
        nvidia.com/gpu: "1"
        cpu: "2000m"
        memory: "4Gi"
    nodeSelector:
      accelerator: "nvidia-tesla-k80"
```

### Node Selector Only
```yaml
resource-classes:
  spot:
    nodeSelector:
      node-type: "spot"
      instance-type: "large"
```

### DIFF
--- a/internal/controller/agenttags/tags.go
+++ b/internal/controller/agenttags/tags.go
@@ -98,6 +98,10 @@ func LabelsFromTags(tags []string) (map[string]string, []error) {
 
 func MatchJobTags(agentTags, jobTags map[string]string) bool {
 	for k, jobTagValue := range jobTags {
+		// resource_class is a reserved keyword in k8s stack. It does not participate matching
+		if k == "resource_class" {
+			continue
+		}
 		agentTagValue, exists := agentTags[k]
 		if !exists {
 			return false
--- a/internal/controller/config/resource_class.go
+++ b/internal/controller/config/resource_class.go
@@ -1,0 +1,58 @@
+package config
+
+import (
+	"maps"
+
+	corev1 "k8s.io/api/core/v1"
+)
+
+const (
+	// This is duplicated from scheduler to avoid cyclic dependencies, it might be a good time to introduce a constant
+	// module now
+	commandContainerName = "container-0"
+)
+
+// ResourceClass represents a reusable resource configuration.
+// Affinity or Toleration/taint based configuration may come later.
+type ResourceClass struct {
+	Resource     *corev1.ResourceRequirements `json:"resource,omitempty"`
+	NodeSelector map[string]string            `json:"nodeSelector,omitempty"`
+}
+
+// Apply adds the resource class NodeSelector to the podSpec, and resource
+// requests and limits to the command container. It assumes there is only one
+// command container named "container-0".
+func (rc *ResourceClass) Apply(podSpec *corev1.PodSpec) {
+	if rc == nil || podSpec == nil {
+		return
+	}
+
+	if len(rc.NodeSelector) > 0 {
+		if podSpec.NodeSelector == nil {
+			podSpec.NodeSelector = make(map[string]string)
+		}
+		maps.Copy(podSpec.NodeSelector, rc.NodeSelector)
+	}
+
+	if rc.Resource != nil {
+		for i := range podSpec.Containers {
+			container := &podSpec.Containers[i]
+
+			// We only care about command container.
+			// checkout and other containers resources can be configured via controller setting.
+			if container.Name != commandContainerName {
+				continue
+			}
+
+			if container.Resources.Requests == nil {
+				container.Resources.Requests = make(corev1.ResourceList)
+			}
+			if container.Resources.Limits == nil {
+				container.Resources.Limits = make(corev1.ResourceList)
+			}
+
+			maps.Copy(container.Resources.Requests, rc.Resource.Requests)
+			maps.Copy(container.Resources.Limits, rc.Resource.Limits)
+		}
+	}
+}
--- a/internal/controller/config/resource_class_test.go
+++ b/internal/controller/config/resource_class_test.go
@@ -1,0 +1,141 @@
+package config
+
+import (
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+)
+
+func TestResourceClass_Apply(t *testing.T) {
+	tests := []struct {
+		name          string
+		resourceClass *ResourceClass
+		podSpec       *corev1.PodSpec
+		want          *corev1.PodSpec
+	}{
+		{
+			name: "applies resource requirements to command container",
+			resourceClass: &ResourceClass{
+				Resource: &corev1.ResourceRequirements{
+					Requests: corev1.ResourceList{
+						corev1.ResourceCPU:    resource.MustParse("100m"),
+						corev1.ResourceMemory: resource.MustParse("128Mi"),
+					},
+					Limits: corev1.ResourceList{
+						corev1.ResourceCPU:    resource.MustParse("200m"),
+						corev1.ResourceMemory: resource.MustParse("256Mi"),
+					},
+				},
+			},
+			podSpec: &corev1.PodSpec{
+				Containers: []corev1.Container{
+					{
+						Name:      commandContainerName,
+						Resources: corev1.ResourceRequirements{},
+					},
+				},
+			},
+			want: &corev1.PodSpec{
+				Containers: []corev1.Container{
+					{
+						Name: commandContainerName,
+						Resources: corev1.ResourceRequirements{
+							Requests: corev1.ResourceList{
+								corev1.ResourceCPU:    resource.MustParse("100m"),
+								corev1.ResourceMemory: resource.MustParse("128Mi"),
+							},
+							Limits: corev1.ResourceList{
+								corev1.ResourceCPU:    resource.MustParse("200m"),
+								corev1.ResourceMemory: resource.MustParse("256Mi"),
+							},
+						},
+					},
+				},
+			},
+		},
+		{
+			name: "applies node selector",
+			resourceClass: &ResourceClass{
+				NodeSelector: map[string]string{
+					"instance-type": "large",
+					"zone":          "us-west-2a",
+				},
+			},
+			podSpec: &corev1.PodSpec{
+				Containers: []corev1.Container{
+					{Name: commandContainerName},
+				},
+			},
+			want: &corev1.PodSpec{
+				NodeSelector: map[string]string{
+					"instance-type": "large",
+					"zone":          "us-west-2a",
+				},
+				Containers: []corev1.Container{
+					{Name: commandContainerName},
+				},
+			},
+		},
+		{
+			name: "applies both resource requirements and node selector",
+			resourceClass: &ResourceClass{
+				Resource: &corev1.ResourceRequirements{
+					Requests: corev1.ResourceList{
+						corev1.ResourceCPU: resource.MustParse("500m"),
+					},
+				},
+				NodeSelector: map[string]string{
+					"instance-type": "xlarge",
+				},
+			},
+			podSpec: &corev1.PodSpec{
+				Containers: []corev1.Container{
+					{Name: commandContainerName},
+				},
+			},
+			want: &corev1.PodSpec{
+				NodeSelector: map[string]string{
+					"instance-type": "xlarge",
+				},
+				Containers: []corev1.Container{
+					{
+						Name: commandContainerName,
+						Resources: corev1.ResourceRequirements{
+							Requests: corev1.ResourceList{
+								corev1.ResourceCPU: resource.MustParse("500m"),
+							},
+							Limits: corev1.ResourceList{},
+						},
+					},
+				},
+			},
+		},
+		{
+			name:          "handles nil resource class",
+			resourceClass: nil,
+			podSpec: &corev1.PodSpec{
+				Containers: []corev1.Container{
+					{Name: commandContainerName},
+				},
+			},
+			want: &corev1.PodSpec{
+				Containers: []corev1.Container{
+					{Name: commandContainerName},
+				},
+			},
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			tt.resourceClass.Apply(tt.podSpec)
+
+			// Compare the entire PodSpec using cmp.Diff
+			if diff := cmp.Diff(tt.want, tt.podSpec); diff != "" {
+				t.Errorf("PodSpec mismatch (-expected +actual):\n%s", diff)
+			}
+		})
+	}
+}
--- a/internal/controller/scheduler/scheduler.go
+++ b/internal/controller/scheduler/scheduler.go
@@ -77,6 +77,7 @@ type Config struct {
 	SkipImageCheckContainers       bool
 	ImageCheckContainerCPULimit    string
 	ImageCheckContainerMemoryLimit string
+	ResourceClasses                map[string]*config.ResourceClass
 }
 
 func New(logger *zap.Logger, client kubernetes.Interface, agentClient *api.AgentClient, cfg Config) *worker {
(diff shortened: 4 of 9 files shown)


internal/controller/config/resource_class.go
```diff
@@ -0,0 +1,56 @@
...
+						container.Resources.Requests[resourceName] = quantity
+					}
+				}
+
+				if rc.Resource.Limits != nil {
+					for resourceName, quantity := range rc.Resource.Limits {
+						container.Resources.Limits[resourceName] = quantity
+					}
+				}
+				break
```
moskyb: given all the nested contexts here, IMO the statement that this `break` applies to is kinda hard to parse out. maybe it'd be easier to read as a
```Go
if container.Name != CommandContainerName {
  continue
}
```

at the top of the block?

internal/controller/scheduler/scheduler.go
```diff
@@ -490,6 +491,22 @@ func (w *worker) Build(podSpec *corev1.PodSpec, skipCheckout bool, inputs buildI
...
+				// Add the resource_class tag to the agent tags
+				agentTags["resource_class"] = resourceClassName
+				w.logger.Debug("Applied resource class", zap.String("resource_class", resourceClassName), zap.String("job-uuid", inputs.uuid))
+			} else {
+				w.logger.Warn("Resource class not found in configuration", zap.String("resource_class", resourceClassName), zap.String("job-uuid", inputs.uuid))
+			}
+		} else {
+			w.logger.Warn("Resource classes not configured but resource_class tag specified", zap.String("resource_class", resourceClassName), zap.String("job-uuid", inputs.uuid))
+		}
+	}
```
moskyb, line 497: a couple of things here:

the nested if/elses here are a little hard to read ([the happy path is left-aligned](https://maelvls.dev/go-happy-line-of-sight/)) -- it might be easier to read as a method with early returns etc

```Go
if resourceClassName, exists := tags["resource_class"]; exists {
  w.applyResourceClass(resourceClassName, podSpec)
}

// ...

func (w *worker) applyResourceClass(podSpec *corev1.PodSpec, resourceClassName string) {
  if w.cfg.ResourceClasses == nil {
    w.logger.Warn("Resource classes not configured but resource_class tag specified", zap.String("resource_class", resourceClassName), zap.String("job-uuid", inputs.uuid))
    return
  }

  if resourceClass, found := w.cfg.ResourceClasses[resourceClassName]; !found {
    w.logger.Warn("Resource class not found in configuration", zap.String("resource_class", resourceClassName), zap.String("job-uuid", inputs.uuid))
    return
  }

  // ... etc
}
```

the other (more important?) thing to note here is that as far as i can tell, the behaviour when resource class allocation fails (the job uses a class not available on the stack, or vice-versa), the behaviour is to continue on and just use the default podspec, which i'm guessing could be counterintuitive to some users.

some things to think about on this front:
- Do we want to give users the option to (or enforce that) there be a default resource class?
- Should giving an invalid resource class fail the job? (this will be much easier when we get server-side fail and job stack events)

internal/controller/config/resource_class.go
```diff
@@ -0,0 +1,56 @@
...
+)
+
+const (
+	CommandContainerName = "container-0"
+)
+
+type ResourceClass struct {
+	Resource     *corev1.ResourceRequirements `json:"resource,omitempty"`
+	NodeSelector map[string]string            `json:"nodeSelector,omitempty"`
+}
```
moskyb, line 20: are there other things that customers might want to specify in a resource class that aren't these two? maybe labels/annotations? i suspect that the motivation here is to _not_ make this a full podspec patch, and to keep its use relatively constrained, but it might be worth thinking about if there are other features we might want to have in here, even if we don't add them right now.

internal/controller/agenttags/tags.go
```diff
@@ -98,6 +98,10 @@ func LabelsFromTags(tags []string) (map[string]string, []error) {
 
 func MatchJobTags(agentTags, jobTags map[string]string) bool {
 	for k, jobTagValue := range jobTags {
+		// resource_class is a reserved keyword in k8s stack. It does not participate matching
+		if k == "resource_class" {
+			continue
+		}
```
moskyb, line 104: is it possible that users are already using a tag called `resource_class` for job matching? will this break their workflow?

i suspect the answer to the second question is "yes" -- i don't think that this means we shouldn't add this feature, but i also wonder if there's space we could create in jobs for this behaviour, a `labels` field or similar. 

it feels like there's a small but notable semantic difference between agent tags ("when using traditional dispatch, only send this job to agents that look like this") and this behaviour ("when using inverted dispatch, create agents that look like this to service this job"). 

again, i don't think that we necessarily need to change anything here, but it's worth thinking about:
- Are there backend changes we could make to support this behaviour better
- _should we_ make those backend changes

internal/integration/fixtures/resource-class.yaml
```diff
@@ -0,0 +1,67 @@
...
+        CPU_PERIOD=$$(cat /sys/fs/cgroup/cpu/cpu.cfs_period_us)
+      elif [ -f /sys/fs/cgroup/cpu.max ]; then
+        # cgroup v2
+        CPU_MAX=$$(cat /sys/fs/cgroup/cpu.max)
+        CPU_QUOTA=$$(echo $$CPU_MAX | cut -d' ' -f1)
+        CPU_PERIOD=$$(echo $$CPU_MAX | cut -d' ' -f2)
+      else
+        echo "Could not find CPU limit files"
+        exit 1
+      fi
```
moskyb, line 50: this is gnarly, good work!

internal/controller/agenttags/tags.go
zhming0: Yeah breaking existing workflow is probably the biggest risk. I did a dig in our data lake, I will share this result in Slack. 

Re semantic change, to be precise:
 
* From pipeline yaml's perspective, the semantic does not change. -> which is the main thing that I care about. 
  * It's still "find me an agent that can meet this agent query rule"
  * It kinda mean that if there is a mismatch resource_class we should bail, I will make this change. 

* From k8s stack operator's perspective, `resource_class` become a reserved tag, which is relatively low impact given my data lake query result. The semantic is as you say: "when there is a demand for a `resource_class` tag, fulfill it". 

I will be against adding a "labels" field to our pipeline yaml for an use case like this, because that would be almost the same with `agents` field.

moskyb: yeah, i agree that creating a new construct in the pipeline yaml for labels probably isn't a good idea, especially because the semantic difference is so small.

framing the semantic difference from the pipeline yaml's perspective is a really good call -- i'm convinced on this, happy to go forward

internal/controller/config/resource_class.go
zhming0: There are mainly 3 configurations to impact how a pod should be a scheduled to a pod. 

- nodeSelector -> which is basically `label`, very common and simple, black and white.
- Affinity. 
- Taints and tolerations. 

The following two were designed for more different use cases. 

It would be trivial to support these two as well, but I reckon we should test the market first with the most common option. 

If this feature is truly helpful, I expect to receive contributions.

internal/controller/scheduler/scheduler.go
zhming0: I've refactored this bit and hopefully making it easier to read.

I also make invalid resource class fail the job. 

The default resource class sounds very desirable, let's let test the market and hopefully people will raise the same request. 😁

internal/controller/config/resource_class.go
zhming0: Done ✅

internal/controller/config/resource_class.go
```diff
@@ -0,0 +1,60 @@
...
+
+	if rc.Resource != nil {
+		for i := range podSpec.Containers {
+			container := &podSpec.Containers[i]
+
+			// We only care about command container.
+			// checkout and other containers resources can be configured via controller setting.
+			if container.Name != CommandContainerName {
+				continue
+			}
```
DrJosh9000: Could you explain a bit more how this interacts with podSpecs from either the controller config or the plugin, which provide containers (possibly more than one) that might not be named this way?

zhming0: podSpec, podSpec patch overrides the resource class setup. 

For job that customized their command container name and/or use more than one command containers, this change doesn't work. 

Given we are finding a golden path outside of podSpec, turning per step podSpec into an escape hatch, hopefully this PR can benefit majority of customers 🙏🏿 . 

I can document it when I add the doc. Is this a blocker?

DrJosh9000: I don't think it's a blocker necessarily, but yeah it should at least be documented.

internal/controller/config/resource_class.go
```diff
@@ -0,0 +1,60 @@
+package config
+
+import (
+	corev1 "k8s.io/api/core/v1"
+)
+
+const (
+	CommandContainerName = "container-0"
```
DrJosh9000: Does this need to be exported?

internal/controller/config/resource_class.go
```diff
@@ -0,0 +1,60 @@
...
+)
+
+type ResourceClass struct {
```
DrJosh9000, line 17: Please add doc comments to exported identifiers

```suggestion
// ResourceClass represents a reusable resource configuration.
type ResourceClass struct {
```

internal/controller/config/resource_class.go
```diff
@@ -0,0 +1,60 @@
...
+	Resource     *corev1.ResourceRequirements `json:"resource,omitempty"`
+	NodeSelector map[string]string            `json:"nodeSelector,omitempty"`
+}
+
+func (rc *ResourceClass) Apply(podSpec *corev1.PodSpec) {
```
DrJosh9000, line 25:
```suggestion
// Apply adds the resource class NodeSelector to the podSpec, and resource
// requests and limits to the command container. It assumes there is only one
// command container named "container-0".
func (rc *ResourceClass) Apply(podSpec *corev1.PodSpec) {
```

internal/controller/config/resource_class.go
```diff
@@ -0,0 +1,60 @@
...
+		return
+	}
+
+	if len(rc.NodeSelector) > 0 {
+		if podSpec.NodeSelector == nil {
+			podSpec.NodeSelector = make(map[string]string)
+		}
+		for key, value := range rc.NodeSelector {
+			podSpec.NodeSelector[key] = value
+		}
```
DrJosh9000: The `maps` package can help:

```suggestion
		maps.Copy(podSpec.NodeSelector, rc.NodeSelector)
```

Also it probably doesn't need the surrounding `len(rc.NodeSelector) > 0` check

internal/controller/config/resource_class.go
```diff
@@ -0,0 +1,67 @@
...
+				for resourceName, quantity := range rc.Resource.Requests {
+					container.Resources.Requests[resourceName] = quantity
+				}
+			}
+
+			if rc.Resource.Limits != nil {
+				for resourceName, quantity := range rc.Resource.Limits {
+					container.Resources.Limits[resourceName] = quantity
+				}
+			}
```
DrJosh9000: Same here - `maps.Copy`, which can handle nil and empty source maps:

```suggestion
			maps.Copy(container.Resources.Requests, rc.Resource.Requests)
			maps.Copy(container.Resources.Limits, rc.Resource.Limits)
```

internal/controller/config/resource_class_test.go
```diff
@@ -0,0 +1,174 @@
...
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+)
+
+func TestResourceClass_Apply(t *testing.T) {
+	tests := []struct {
+		name          string
+		resourceClass *ResourceClass
+		podSpec       *corev1.PodSpec
+		expected      *corev1.PodSpec
```
DrJosh9000: [Josh's favourite Go testing hobby-horse] I personally prefer `want` instead of `expected` https://google.github.io/styleguide/go/decisions#got-before-want

internal/controller/config/resource_class_test.go
```diff
@@ -0,0 +1,174 @@
...
+
+			// Compare node selector
+			if len(tt.expected.NodeSelector) != len(tt.podSpec.NodeSelector) {
+				t.Errorf("NodeSelector length mismatch: expected %d, got %d", len(tt.expected.NodeSelector), len(tt.podSpec.NodeSelector))
+			}
+			for k, v := range tt.expected.NodeSelector {
+				if tt.podSpec.NodeSelector[k] != v {
+					t.Errorf("NodeSelector[%s] = %s, expected %s", k, tt.podSpec.NodeSelector[k], v)
+				}
+			}
```
DrJosh9000: See above re `want` instead of `expected`.

You can check the whole `NodeSelector` structure pretty easily with [cmp.Diff](https://pkg.go.dev/github.com/google/go-cmp/cmp#Diff) though:

```suggestion
			if diff := cmp.Diff(tt.podSpec.NodeSelector, tt.want.NodeSelector); diff != "" {
					t.Errorf("NodeSelector diff (-got +want):\n%s", diff)
				}
			}
```

internal/controller/config/resource_class_test.go
```diff
@@ -0,0 +1,174 @@
...
+
+				// Compare resource limits
+				for resourceName, expectedQuantity := range expectedContainer.Resources.Limits {
+					actualQuantity := actualContainer.Resources.Limits[resourceName]
+					if !actualQuantity.Equal(expectedQuantity) {
+						t.Errorf("Container[%d].Resources.Limits[%s] = %s, expected %s", i, resourceName, actualQuantity.String(), expectedQuantity.String())
+					}
+				}
+			}
+		})
```
DrJosh9000: I think much of this could also use `cmp.Diff`.

zhming0: nice, much cleaner!

internal/controller/config/resource_class.go
zhming0: I kept the `len(rc.NodeSelector) > 0` check to avoid creating `podSpec.NodeSelector` unnecessarily. 

but `maps.Copy` is nice!

internal/controller/config/resource_class_test.go
zhming0: Not disagreeing, but it is interesting that the first sentence of the paragraph: 

> Test outputs should include the **actual** value that the function returned before printing the value that was **expected**.

😅

DrJosh9000 (approved):
LGTM

PR #669, opened 2025-08-12
Add Grafana dashboard (based on Prometheus metrics)

DrJosh9000:
Adds a Grafana dashboard JSON definition. 

Rather than simply provide the JSON, this PR adds the dashboard as a ConfigMap installed with the Helm chart, having the conventional kube-prometheus-stack label `grafana_dashboard: "1"`. The deployment can be prevented with `deployGrafanaDashboard: false` in the Helm `values.yaml`.

Image 13-8-2025 at 4 53 pm

### DIFF
--- a/charts/agent-stack-k8s/values.yaml
+++ b/charts/agent-stack-k8s/values.yaml
@@ -18,3 +18,5 @@ annotations: {}
 secretsMetadata: {}
 
 serviceAccountMetadata: {}
+
+deployGrafanaDashboard: true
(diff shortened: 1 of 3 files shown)


examples/grafana.json
zhming0, line 1: It looks good but I think since we are doing helm. We should deploy this dashboard as a configmap with label: `grafana_dashboard: "1"` so typical grafana deployment will automatically install this.

DrJosh9000: Good idea!

charts/agent-stack-k8s/values.yaml
```diff
@@ -18,3 +18,5 @@ annotations: {}
 secretsMetadata: {}
 
 serviceAccountMetadata: {}
+
+installGrafanaDashboard: true
```
zhming0: deploy or install? The values.schema.json uses `deploy` 😬

zhming0 (changes requested):

charts/agent-stack-k8s/values.yaml
DrJosh9000: Good catch, thanks!

zhming0 (approved):

PR #670, opened 2025-08-13
Fix monitor jobs handled metric

DrJosh9000:
### What
Fix the monitor metrics to do with jobs being passed to the next handler. 
Add new metrics that have the same values as the current metrics.

### Huh?
`buildkite_monitor_jobs_handled_total` is described as "Count of jobs that were passed to the next handler in the chain", and incremented for every call to the next handler. This was correct when each job was passed in an individual `Handle` call, but for a while now (0.28 betas) the limiter has received batches of jobs with `HandleMany`. There's a similar issue for errors counts.

### DIFF
--- a/internal/controller/monitor/monitor.go
+++ b/internal/controller/monitor/monitor.go
@@ -208,6 +208,7 @@ func (m *Monitor) passJobsToNextHandler(
 		zap.Int("job-count", len(filteredJobs)),
 	)
 	jobHandlerCallsCounter.Inc()
+	jobsHandledCounter.Add(float64(len(filteredJobs)))
 
 	if err := handler.HandleMany(ctx, filteredJobs); err != nil {
 		if ctx.Err() != nil {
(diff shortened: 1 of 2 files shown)


internal/controller/monitor/monitor.go
```diff
@@ -208,12 +208,14 @@ func (m *Monitor) passJobsToNextHandler(
...
 	jobHandlerCallsCounter.Inc()
+	jobsHandledCounter.Add(float64(len(filteredJobs)))
 
 	if err := handler.HandleMany(ctx, filteredJobs); err != nil {
 		if ctx.Err() != nil {
 			return
 		}
 		m.logger.Error("failed to create jobs", zap.Error(err))
 		jobHandlerErrorCounter.Inc()
+		jobsHandledErrorsCounter.Add(float64(len(filteredJobs)))
```
zhming0, line 219: I wonder if the semantic of `HandleMany` is atomic?

DrJosh9000: Our primary implementation (`Limiter`) wraps a mutex around its operations, and deals with all the jobs its given at once (adds them to its queue, does some queue maintenance, and then wakes up workers in a non-blocking fashion). So yes, in both senses.

PR #678, opened 2025-08-26
Stacks API client

moskyb:
This PR adds an (admittedly somewhat meaty) skeleton of an API client to interact with Buildkite's forthcoming Stacks API. The idea of the Stacks API is to have a dedicated and documented place for orchestrators like the k8s stack to interact with Buildkite.

At some point, we'll probably pull this package out of the k8s stack and into its own repo, but until we have a relatively stable API it'll probably remain in the internal folder of this repo.

This PR contains the meat and potatoes of the API client itself, including default retry policies etc, as well as a single endpoint method, which hits the `POST /stacks/register` endpoint.

### DIFF
--- a/internal/stacksapi/client.go
+++ b/internal/stacksapi/client.go
@@ -1,0 +1,354 @@
+package stacksapi
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io"
+	"log/slog"
+	"net/http"
+	"net/http/httputil"
+	"net/url"
+	"time"
+
+	"github.com/buildkite/agent-stack-k8s/v2/internal/version"
+	"github.com/buildkite/roko"
+)
+
+// Client is a Buildkite Stacks API client.
+type Client struct {
+	baseURL    *url.URL
+	userAgent  string
+	authHeader string
+
+	logger          *slog.Logger
+	logHTTPPayloads bool
+
+	httpClient  *http.Client
+	retrierOpts []roko.RetrierOpt
+}
+
+func urlMustParse(u string) *url.URL {
+	parsed, err := url.Parse(u)
+	if err != nil {
+		panic(err)
+	}
+	return parsed
+}
+
+type ClientOpt func(*Client) error
+
+const (
+	DefaultBaseURL = "https://agent.buildkite.com/v3/"
+	DefaultQueue   = "_default"
+)
+
+var (
+	DefaultRetrierOptions = []roko.RetrierOpt{
+		roko.WithMaxAttempts(5),
+		roko.WithStrategy(roko.ExponentialSubsecond(1 * time.Second)),
+		roko.WithJitterRange(500*time.Millisecond, 1*time.Second),
+	}
+
+	defaultUserAgent = "agent-stack-k8s/" + version.Version()
+)
+
+// WithLogger sets the [*slog.Logger] for the API client. The default is a [slog.Logger] using the [slog.DiscardHandler] handler.
+func WithLogger(logger *slog.Logger) ClientOpt {
+	return func(c *Client) error {
+		c.logger = logger
+		return nil
+	}
+}
+
+// LogHTTPRequestPayloads instructs the client to log all HTTP request payloads. Note that this may log sensitive information,
+// and should be used with caution
+func LogHTTPPayloads() ClientOpt {
+	return func(c *Client) error {
+		c.logHTTPPayloads = true
+		return nil
+	}
+}
+
+// WithBaseURL sets the base URL for the API client, overriding [DefaultBaseURL]
+func WithBaseURL(baseURL *url.URL) ClientOpt {
+	return func(c *Client) error {
+		c.baseURL = baseURL
+		return nil
+	}
+}
+
+// PrependToUserAgent adds a prefix to the User-Agent header for the API client. Note that overriding of the user-agent
+// is not supported (though can probably be worked around).
+func PrependToUserAgent(prefix string) ClientOpt {
+	return func(c *Client) error {
+		c.userAgent = prefix + " " + c.userAgent
+		return nil
+	}
+}
+
+// WithHTTPClient sets the HTTP client for the API client, overriding [http.DefaultClient].
+func WithHTTPClient(httpClient *http.Client) ClientOpt {
+	return func(c *Client) error {
+		c.httpClient = httpClient
+		return nil
+	}
+}
+
+// WithRetrierOptions sets the default retrier options for the API client. These can be overridden per-request by using the [WithRetrier] RequestOpt.
+func WithRetrierOptions(opts ...roko.RetrierOpt) ClientOpt {
+	return func(c *Client) error {
+		c.retrierOpts = opts
+		return nil
+	}
+}
+
+// NewClient creates a new API client with the given token and options. Note that the token must be a Buildkite Cluster Token,
+// and that REST/GraphQL API tokens will not work.ql
+func NewClient(apiToken string, opts ...ClientOpt) (*Client, error) {
+	client := &Client{
+		httpClient:  http.DefaultClient,
+		baseURL:     urlMustParse(DefaultBaseURL),
+		userAgent:   defaultUserAgent,
+		logger:      slog.New(slog.DiscardHandler),
+		authHeader:  fmt.Sprintf("Token %s", apiToken),
+		retrierOpts: DefaultRetrierOptions,
+	}
+
+	for _, opt := range opts {
+		if err := opt(client); err != nil {
+			return nil, err
+		}
+	}
+
+	return client, nil
+}
+
+type StackAPIRequest struct {
+	*http.Request
+
+	retrier   *roko.Retrier
+	bodyBytes []byte
+}
+
+// resetBody recreates the request body from the stored bytes for retry attempts
+func (r *StackAPIRequest) resetBody() {
+	r.Body = io.NopCloser(bytes.NewReader(r.bodyBytes))
+}
+
+type RequestOption func(*StackAPIRequest) error
+
+// WithRetrier sets the retrier for the request, overriding the default retrier belonging to the [Client].
+func WithRetrier(retrier *roko.Retrier) RequestOption {
+	return func(r *StackAPIRequest) error {
+		r.retrier = retrier
+		return nil
+	}
+}
+
+// newRequest creates a new API request suitable for dispatch to the Buildkite Stacks API with the given context, method, path, and body.
+func (c *Client) newRequest(ctx context.Context, method, path string, body any, opts ...RequestOption) (*StackAPIRequest, error) {
+	fullURL := c.baseURL.ResolveReference(&url.URL{Path: path})
+
+	var bodyBytes []byte
+	if body != nil {
+		var err error
+		bodyBytes, err = json.Marshal(body)
+		if err != nil {
+			return nil, fmt.Errorf("failed to encode request body: %w", err)
+		}
+	}
+
+	req, err := http.NewRequestWithContext(ctx, method, fullURL.String(), bytes.NewReader(bodyBytes))
+	if err != nil {
+		return nil, fmt.Errorf("failed to create request: %w", err)
+	}
+
+	req.Header.Set("User-Agent", c.userAgent)
+	req.Header.Set("Authorization", c.authHeader)
+	req.Header.Set("Accept", "application/json")
+	req.Header.Set("Content-Type", "application/json")
+
+	logger := c.logger.With(
+		"method", req.Method,
+		"url", req.URL,
+	)
+
+	logger.DebugContext(ctx, "created request")
+
+	stackReq := &StackAPIRequest{
+		Request:   req,
+		retrier:   roko.NewRetrier(c.retrierOpts...),
+		bodyBytes: bodyBytes,
+	}
+
+	for _, opt := range opts {
+		err := opt(stackReq)
+		if err != nil {
+			return nil, fmt.Errorf("failed to apply request option: %w", err)
+		}
+	}
+
+	return stackReq, nil
+}
+
+// do performs the given *StackAPIRequest and returns the HTTP response, unmarshalling the body into val.
+func (c *Client) do(ctx context.Context, req *StackAPIRequest, val any) (*http.Response, error) {
+	resp, err := c.doWithRetry(ctx, req)
+	if err != nil {
+		return nil, err
+	}
+
+	if err := c.processResponse(ctx, resp, val); err != nil {
+		return nil, err
+	}
+
+	return resp, nil
+}
+
+// doWithRetry handles the retry logic for HTTP requests
+func (c *Client) doWithRetry(ctx context.Context, req *StackAPIRequest) (*http.Response, error) {
+	logger := c.logger.With(
+		"method", req.Method,
+		"url", req.URL.String(),
+	)
+
+	return roko.DoFunc(ctx, req.retrier, func(r *roko.Retrier) (*http.Response, error) {
+		req.resetBody()
+
+		sendRequestLogger := c.prepareRequestLogger(req, logger)
+		sendRequestLogger.DebugContext(ctx, "sending request")
+
+		resp, err := c.httpClient.Do(req.Request)
+		if err != nil {
+			return nil, err
+		}
+
+		logger := c.logger.With("response_status", resp.StatusCode)
+
+		responseLogger := c.prepareResponseLogger(resp, logger)
+		responseLogger.DebugContext(ctx, "received response")
+
+		if err := c.handleResponseError(ctx, resp, r, logger); err != nil {
+			return nil, err
+		}
+
+		return resp, nil
+	})
+}
+
+// handleResponseError processes error responses and determines retry behavior
+func (c *Client) handleResponseError(ctx context.Context, resp *http.Response, r *roko.Retrier, logger *slog.Logger) error {
+	err := checkResponse(resp)
+	if err != nil {
+		var errResp *ErrorResponse
+		if errors.As(err, &errResp) {
+			if errResp.IsRetryableStatus() {
+				logger = logger.With("retry_state", r.String())
+			} else {
+				r.Break()
+			}
+
+			logger.DebugContext(ctx, "request failed", "error", err)
+			return err
+		}
+
+		logger.DebugContext(ctx, "request errored", "error", err)
+		return err
+	}
+	return nil
+}
+
+// prepareRequestLogger sets up logging with optional request dump
+func (c *Client) prepareRequestLogger(req *StackAPIRequest, logger *slog.Logger) *slog.Logger {
+	if !c.logHTTPPayloads {
+		return logger
+	}
+
+	reqDump, err := httputil.DumpRequestOut(req.Request, true)
+	if err != nil {
+		logger.Warn("Failed to dump request. Log won't include request body", "error", err)
+		return logger
+	}
+
+	return logger.With("request", string(reqDump))
+}
+
+// processResponse handles response body reading and JSON unmarshaling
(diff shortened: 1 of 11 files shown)


internal/stacksapi/client.go
```diff
@@ -0,0 +1,261 @@
...
+	var bodyBytes []byte
+	if body != nil {
+		var err error
+		bodyBytes, err = json.Marshal(body)
+		if err != nil {
+			return nil, fmt.Errorf("failed to encode request body: %w", err)
+		}
+	}
+
+	req, err := http.NewRequestWithContext(ctx, method, fullURL.String(), bytes.NewReader(bodyBytes))
```
moskyb, line 164: it's not technically necessary to set the body here as we re-set it in [`client.Do`](https://github.com/buildkite/agent-stack-k8s/pull/678/files#diff-a900984307a3d164b41fc7e3049c53a6e6f469b7d50bb289e7bbf83ddb64eeb7R186), but IMO it improves the readability of the method and is less surprising than never setting a body. open to other opinions though

internal/stacksapi/client.go
```diff
@@ -0,0 +1,261 @@
...
+	DefaultBaseURL = "https://agent.buildkite.com/v3/"
+	DefaultQueue   = "_default"
+)
+
+var (
+	DefaultRetrierOptions = []roko.RetrierOpt{
+		roko.WithMaxAttempts(5),
+		roko.WithStrategy(roko.Constant(1 * time.Second)),
+		roko.WithJitterRange(500*time.Millisecond, 1*time.Second),
+	}
```
moskyb: i have no idea if this is a good default retry policy, i basically just put something on paper and went with it. happy to do something better.

internal/stacksapi/client.go
```diff
@@ -0,0 +1,261 @@
...
+}
+
+// WithRetrierOptions sets the default retrier options for the API client. These can be overridden per-request by using the [WithRetrier] RequestOpt.
+func WithRetrierOptions(opts ...roko.RetrierOpt) ClientOpt {
+	return func(c *Client) error {
+		c.retrierOpts = opts
+		return nil
+	}
+}
```
moskyb, line 106: This might seem like a somewhat odd way to define a default retry strategy (as opposed to `WithRetrier(myRokoRetrier)`), but roko retriers have internal state, so it's not safe to share them across a client that may be executing multiple requests concurrently. happy to discuss better ways of doing this!

internal/stacksapi/client.go
```diff
@@ -0,0 +1,262 @@
+package stacksapi
```
DrJosh9000, line 1: One of these files should have a package comment

```suggestion
// Package stacksapi provides a client for the Buildkite Stacks API.
package stacksapi
```

internal/stacksapi/client.go
```diff
@@ -0,0 +1,262 @@
...
+	"log/slog"
+	"net/http"
+	"net/url"
+	"time"
+
+	"github.com/buildkite/agent-stack-k8s/v2/internal/version"
+	"github.com/buildkite/roko"
+)
+
+type Client struct {
```
DrJosh9000, line 21:
```suggestion
// Client implements the API client.
type Client struct {
```

internal/stacksapi/client.go
```diff
@@ -0,0 +1,263 @@
...
+	baseURL    *url.URL
+	userAgent  string
+	authHeader string
+
+	logger      *slog.Logger
```
DrJosh9000: Oooh, is the thinking we'll centralise on `slog`? 😀

internal/stacksapi/client.go
DrJosh9000: For various reasons, I would make the default strategy `ExponentialSubsecond`.

internal/stacksapi/client.go
moskyb: god, i sure hope so.

moskyb: the benefit is that there are a million and one adapters for slog, so even if user aren't using slog, they should be able to plug their logging lib of choice in

internal/stacksapi/client.go
moskyb: 🤔 should we make `NewClient` return an interface that users can mock if necessary?

DrJosh9000: It would be much better to provide a functioning fake of the service.

internal/stacksapi/client.go
```diff
@@ -0,0 +1,264 @@
...
+// WithRetrier sets the retrier for the request, overriding the default retrier belonging to the [Client].
+func WithRetrier(retrier *roko.Retrier) RequestOption {
+	return func(r *StackAPIRequest) error {
+		r.retrier = retrier
+		return nil
+	}
+}
+
+// NewRequest creates a new API request suitable for dispatch to the Buildkite Stacks API with the given context, method, path, and body.
+func (c *Client) NewRequest(ctx context.Context, method, path string, body any, opts ...RequestOption) (*StackAPIRequest, error) {
```
DrJosh9000: We should think about which methods to export from the package. Should a user be calling `NewRequest` and `Do` themselves?

moskyb: hmm, that's a good point, these two should probably be private

internal/stacksapi/doc.go
```diff
@@ -0,0 +1,2 @@
+// The stacksapi package implements an API client for the Buildkite Stacks API.
+package stacksapi
```
zhming0, line 2: I am curious about the use of plural here. I thought we mostly do singular like "AgentAPI"?

internal/stacksapi/client.go
```diff
@@ -0,0 +1,264 @@
...
+)
+
+var (
+	DefaultRetrierOptions = []roko.RetrierOpt{
+		roko.WithMaxAttempts(5),
+		roko.WithStrategy(roko.ExponentialSubsecond(1 * time.Second)),
+		roko.WithJitterRange(500*time.Millisecond, 1*time.Second),
+	}
+
+	defaultUserAgent = "agent-stack-k8s/internal/stacksapi/" + version.Version()
```
zhming0: It isn't obvious to me if we need `internal/stacksapi/` as part of user agent. Would `agent-stack-k8s/v0.xx.xx` suffice? In the future, say, the library is extracted out, and then this `internal` bit wouldn't make sense, then we would have to careful write our data lake query to capture this detail. 

We probably won't do analytics on that 🤔

internal/stacksapi/register.go
```diff
@@ -0,0 +1,46 @@
...
+	Key      string            `json:"key,omitempty"`       // The key to call the stack. Required.
+	Type     StackType         `json:"type,omitempty"`      // A type for the stack. Required.
+	QueueKey string            `json:"queue_key,omitempty"` // The key of a queue for the stack to listen to. Required
+	Metadata map[string]string `json:"metadata"`            // Key-value metadata to associate with the queue. Required, but can be empty
+}
+
+// Stack is the output type for [RegisterStack]. It contains all the information about a registered stack
+type Stack struct {
+	ID              string            `json:"id"`
+	OrganizationID  string            `json:"organization_id"`
```
zhming0: I believe we return UUID instead.

```suggestion
	OrganizationUUID  string            `json:"organization_uuid"`
```

internal/stacksapi/client.go
```diff
@@ -0,0 +1,264 @@
...
+
+	req.Header.Set("User-Agent", c.userAgent)
+	req.Header.Set("Authorization", c.authHeader)
+	req.Header.Set("Content-Type", "application/json")
+
+	c.logger.With(
+		"method", req.Method,
+		"url", req.URL,
+		"request_headers", req.Header,
+		"request_body", string(bodyBytes),
```
zhming0: [blocking] These will result in credentials been logged. 

Also, request payload could be heavy in some cases, it's hard to predict.

moskyb: agree that it's probably best to not have this be opt-out, but i think debugging what HTTP requests are being made is probably a good idea. i'll put this behind an additional client option

internal/stacksapi/client.go
moskyb: yeah, you're probably right, i'll change this

internal/stacksapi/register.go
moskyb: ooh, good catch!

internal/stacksapi/doc.go
moskyb: i think the project being called "stacks" has crept into my brain and made `stackapi` feel weird. i think you're probably right but i also don't think it matters too much, as the REST API client's package name is just `buildkite`, and the agent API isn't exposed to customers

internal/stacksapi/type.go
```diff
@@ -0,0 +1,20 @@
+package stacksapi
+
+type StackType string
```
DrJosh9000, line 5:
```suggestion
// StackType enumerates a few broad categories of stack.
type StackType string
```

internal/stacksapi/type.go
```diff
@@ -0,0 +1,20 @@
...
+const (
+	StackTypeKubernetes StackType = "kubernetes" // Reserved for use by the first-party agent-stack-k8s
+	StackTypeElastic    StackType = "elastic"    // Reserved for use by the first-party elastic-stack-for-aws
+
+	// For use by custom stacks. If you're reading these docs and you don't work at buildkite, this is almost certainly the
+	// stack type you should be using, even if your stack is backed by Kubernetes
+	StackTypeCustom StackType = "custom"
+)
+
+type StackState string
```
DrJosh9000, line 17:
```suggestion
// StackState enumerates the main possible states a stack can be in, relative to Buildkite.
type StackState string
```

internal/stacksapi/example/main.go
```diff
@@ -0,0 +1,36 @@
...
+func main() {
+	clusterToken := os.Getenv("BUILDKITE_CLUSTER_TOKEN")
+	logger := slog.New(slog.NewTextHandler(os.Stdout, &slog.HandlerOptions{Level: slog.LevelDebug}))
+
+	client, err := stacksapi.NewClient(clusterToken,
+		stacksapi.WithLogger(logger),
+		// stacksapi.LogHTTPPayloads(),
+	)
+	if err != nil {
+		panic(err)
```
DrJosh9000: Why panic? The only error here would be from one of the variadic options, so a stack trace would be overkill.

```suggestion
		slog.Error("Couldn't create Stacks API client", "error", err)
		os.Exit(1)
```

internal/stacksapi/example/main.go
```diff
@@ -0,0 +1,36 @@
...
+	stack, err := client.RegisterStack(context.Background(), stacksapi.RegisterStackParams{
+		Key:      "example-stack",
+		Type:     stacksapi.StackTypeCustom,
+		QueueKey: stacksapi.DefaultQueue,
+		Metadata: map[string]string{
+			"test": "true",
+		},
+	})
+	if err != nil {
+		panic(err)
```
DrJosh9000: Here too (re panicking) - this is more likely to be some network failure, not programmer error.


```suggestion
		slog.Error("Couldn't register with Stacks API", "error", err)
		os.Exit(1)
```

internal/stacksapi/client.go
DrJosh9000: How about (very minor change):

```go
var bodyReader io.Reader
if body != nil {
	bodyBytes, err := json.Marshal(body)
	if err != nil {
		return nil, fmt.Errorf(...)
	}
	bodyReader = bytes.NewReader(bodyBytes)
}

req, err := http.NewRequestWithContext(ctx, method, fullURL.String(), bodyReader)
```

internal/stacksapi/client.go
```diff
@@ -0,0 +1,300 @@
...
+
+	logger := c.logger.With(
+		"method", req.Method,
+		"url", req.URL,
+	)
+	if c.logHTTPPayloads {
+		logger = logger.With(
+			"request_headers", req.Header,
+			"request_body", string(bodyBytes),
+		)
```
DrJosh9000: [Food for thought, no change requested] If we're going to log these things, is this the best choice? Over in the agent the debug logs use `httputil.DumpRequestOut` and friends to get a "wire" like representation of the request/response.

internal/stacksapi/client.go
```diff
@@ -0,0 +1,300 @@
...
+
+	data, err := io.ReadAll(r.Body)
+	if err != nil {
+		return fmt.Errorf("response failed with error %w, but reading response body failed with error %w", errorResponse, err)
+	}
+	errorResponse.RawBody = data
+
+	err = json.Unmarshal(data, errorResponse)
+	if err != nil {
+		return fmt.Errorf("response failed with error %w, but parsing response body JSON failed with error: %w. Raw body of error was: %s", errorResponse, err, string(data))
```
DrJosh9000: `%q` will be a better choice than `%s` for data that failed to unmarshal:

```suggestion
		return fmt.Errorf("response failed with error %w, but parsing response body JSON failed with error: %w. Raw body of error was: %q", errorResponse, err, string(data))
```

internal/stacksapi/client.go
```diff
@@ -0,0 +1,300 @@
...
+			return nil, fmt.Errorf("failed to decode response body: %w", err)
+		}
+	}
+
+	return resp, nil
+}
+
+// ErrorResponse provides a message.
+type ErrorResponse struct {
+	Response *http.Response // HTTP response that caused this error
```
DrJosh9000: This could go awry if the error JSON response has a field called `Response`

```suggestion
	Response *http.Response `json:"-"` // HTTP response that caused this error
```

internal/stacksapi/client.go
```diff
@@ -0,0 +1,300 @@
...
+
+		logger.DebugContext(ctx, "sending request")
+		resp, err := c.httpClient.Do(req.Request)
+		if err != nil {
+			return nil, err
+		}
+
+		err = checkResponse(resp)
+		if err != nil {
+			errResp := err.(*ErrorResponse)
```
DrJosh9000: This line would panic if `err` wasn't an `*ErrorResponse`. Which could happen, because `checkResponse` returns a `fmt.Errorf(...)` if either the response body couldn't be read to completion, or the `json.Unmarshal` failed.

internal/stacksapi/client_test.go
```diff
@@ -0,0 +1,198 @@
...
+	"github.com/buildkite/roko"
+)
+
+func TestNewClient(t *testing.T) {
+	t.Parallel()
+	t.Run("creates client with defaults", func(t *testing.T) {
+		t.Parallel()
+		client, err := NewClient("test-token")
+		if err != nil {
+			t.Fatalf("unexpected error: %v", err)
```
DrJosh9000: Unexpected error from what?

```suggestion
			t.Fatalf("NewClient('test-token') error = %v", err)
```

internal/stacksapi/client_test.go
```diff
@@ -0,0 +1,198 @@
...
+		}
+
+		if client.baseURL.String() != DefaultBaseURL {
+			t.Errorf("expected baseURL %s, got %s", DefaultBaseURL, client.baseURL)
```
DrJosh9000: I remain steadfast in advocating for "got before want" https://google.github.io/styleguide/go/decisions#got-before-want

Also `%q` is the superior choice here

```suggestion
			t.Errorf("client.baseURL = %q, want %q", client.baseURL, DefaultBaseURL)
```

internal/stacksapi/client_test.go
```diff
@@ -0,0 +1,198 @@
...
+		}
+
+		expectedUA := "agent-stack-k8s/" + version.Version()
+		if client.userAgent != expectedUA {
+			t.Errorf("expected userAgent %s, got %s", expectedUA, client.userAgent)
```
DrJosh9000: and here, and so on and so forth etc

internal/stacksapi/client_test.go
```diff
@@ -0,0 +1,198 @@
...
+			t.Fatalf("unexpected error: %v", err)
+		}
+
+		if client.baseURL.String() != customURL.String() {
+			t.Errorf("expected baseURL %s, got %s", customURL, client.baseURL)
+		}
+
+		if client.httpClient != customClient {
+			t.Error("expected custom HTTP client to be set")
+		}
```
DrJosh9000: Including the value that `client.httpClient` was, even if it turns out it's a pointer to a somewhat large struct, might make debugging test failures easier.

```suggestion
		if client.httpClient != customClient {
			t.Errorf("client.httpClient = %v, want customClient", client.httpClient)
		}
```

internal/stacksapi/client_test.go
```diff
@@ -0,0 +1,198 @@
...
+
+		expectedURL := "https://agent.buildkite.com/v3/test"
+		if req.URL.String() != expectedURL {
+			t.Errorf("expected URL %s, got %s", expectedURL, req.URL.String())
+		}
+
+		if req.Header.Get("Authorization") != "Token test-token" {
+			t.Error("Authorization header not set correctly")
+		}
```
DrJosh9000: Not set correctly, so what was it set to?

```suggestion
		if got, want := req.Header.Get("Authorization"), "Token test-token" {
			t.Error("req.Header.Get(\"Authorization\") = %q, want %q", got, want)
		}
```

internal/stacksapi/client_test.go
```diff
@@ -0,0 +1,198 @@
...
+
+		if req.Header.Get("Content-Type") != "application/json" {
+			t.Error("Content-Type header not set correctly")
+		}
```
DrJosh9000:
```suggestion
		if got, want := req.Header.Get("Content-Type"), "application/json" {
			t.Error("req.Header.Get(\"Content-Type\") = %q, want %q", got, want)
		}
```

internal/stacksapi/register.go
```diff
@@ -0,0 +1,46 @@
...
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+	"time"
+)
+
+// RegisterStackParams is the input type for [RegisterStack]. All of its fields are required.
+type RegisterStackParams struct {
```
DrJosh9000: If this is at the "API client" layer, should this be `RegisterStackRequest`, and the response type be `RegisterStackResponse` ?

DrJosh9000 (commented):
Some things to think about

internal/stacksapi/example/main.go
moskyb: yeah, this was mostly me getting lazy with example code, doing things properly is a good idea.

internal/stacksapi/client.go
moskyb: ooh this is a good call, will change

internal/stacksapi/client.go
moskyb: yep, good shout

internal/stacksapi/register.go
```diff
@@ -0,0 +1,46 @@
...
+	Type             StackType         `json:"type"`
+	Metadata         map[string]string `json:"metadata"`
+	LastConnectedOn  *time.Time        `json:"last_connected_on"`
+	State            StackState        `json:"state"`
+}
+
+// RegisterStack registers a new stack with the Buildkite Stacks API. A stack with the same key can safely be
+// re-registered as many times as necessary.
+func (c *Client) RegisterStack(ctx context.Context, body RegisterStackRequest, opts ...RequestOption) (RegisterStackResponse, error) {
+	req, err := c.newRequest(ctx, http.MethodPost, "stacks/register", body, opts...)
```
zhming0, line 33: I wonder if the newRequest should handle the `/stack` bit?

zhming0 (approved):
Let's go! 👍🏿

internal/stacksapi/register.go
moskyb: Yeah, maybe... nice to have the whole meaningful part of the URL in the method definition though

PR #682, opened 2025-09-04
Deregister endpoint

moskyb:
This PR mostly just adds a method that uses the `deregister` endpoint, but it also adds a couple of testing bits and bobs.

### DIFF
--- a/internal/stacksapi/register.go
+++ b/internal/stacksapi/register.go
@@ -47,7 +47,7 @@ func (c *Client) RegisterStack(ctx context.Context, body RegisterStackRequest, o
 
 // DeregisterStack informs the Buildkite Stacks API that a stack is exiting cleanly.
 func (c *Client) DeregisterStack(ctx context.Context, stackKey string, opts ...RequestOption) error {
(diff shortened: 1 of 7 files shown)


internal/stacksapi/register.go
```diff
@@ -38,9 +38,25 @@ func (c *Client) RegisterStack(ctx context.Context, body RegisterStackRequest, o
...
 	if err != nil {
-		return RegisterStackResponse{}, fmt.Errorf("request failed: %w", err)
+		return RegisterStackResponse{}, fmt.Errorf("register stack: %w", err)
 	}
 	defer resp.Body.Close()
 
 	return stack, nil
 }
+
+func (c *Client) DeregisterStack(ctx context.Context, stackID string, opts ...RequestOption) error {
```
zhming0: As a client library I think will be nice to have code comments here.

zhming0 (approved):
Looks good.

internal/stacksapi/register.go
moskyb: oh yup, good call. the arg should be called `stackKey` too.

PR #687, opened 2025-09-09
Register/deregister stacks

moskyb:
This PR adjusts the k8s stack to register itself as a stack on the buildkite backend, and deregister itself when it exits cleanly. if the controller is SIGKILLed or gets two SIGTERMs, the deregister call will be interrupted.

### DIFF
--- a/api/agent_client.go
+++ b/api/agent_client.go
@@ -12,6 +12,7 @@ import (
 	"time"
 
 	"github.com/buildkite/agent-stack-k8s/v2/internal/controller/agenttags"
+	"github.com/buildkite/agent-stack-k8s/v2/internal/stacksapi"
 )
 
 // This is a special keyword supported on backend for polling from the current default queue in the cluster.
@@ -25,20 +26,44 @@ func WithReservation(enable bool) AgentClientOption {
 	}
 }
 
-func NewAgentClient(token, endpoint, clusterID, queue string, agentQueryRules []string, opts ...AgentClientOption) (*AgentClient, error) {
+func WithStacksAPIClient(client *stacksapi.Client) AgentClientOption {
+	return func(c *AgentClient) {
+		c.stacksAPIClient = client
+	}
+}
+
+// AgentClient is a client for Agent API methods for retrieving jobs.
+type AgentClient struct {
+	endpoint        *url.URL
+	httpClient      *http.Client
+	stacksAPIClient *stacksapi.Client
+
+	clusterID string // or "unclustered"
+	queue     string
+	stack     stacksapi.RegisterStackResponse
+
+	// This impacts a number of endpoints' query parameters
+	reservation bool
+}
+
+func NewAgentClient(ctx context.Context, token, endpoint, clusterID, queue, stackID string, agentQueryRules []string, opts ...AgentClientOption) (*AgentClient, error) {
 	if endpoint == "" {
 		endpoint = "https://agent.buildkite.com/v3"
 	}
+
 	if clusterID == "" {
 		clusterID = "unclustered"
 	}
+
 	endpointURL, err := url.Parse(endpoint)
 	if err != nil {
 		return nil, err
 	}
+
 	if queue == "" {
 		queue = defaultQueueKey
 	}
+
 	client := &AgentClient{
 		endpoint: endpointURL,
 		httpClient: &http.Client{
@@ -48,20 +73,31 @@ func NewAgentClient(token, endpoint, clusterID, queue string, agentQueryRules []
 		clusterID: clusterID,
 		queue:     queue,
 	}
+
 	for _, opt := range opts {
 		opt(client)
 	}
-	return client, nil
-}
 
-// AgentClient is a client for Agent API methods for retrieving jobs.
-type AgentClient struct {
-	endpoint   *url.URL
-	httpClient *http.Client
-	clusterID  string // or "unclustered"
-	queue      string
-	// This impacts a number of endpoints' query parameters
-	reservation bool
+	if client.stacksAPIClient != nil {
+		stackKey := stackID
+		if stackKey == "" {
+			stackKey = "agent-stack-k8s"
+		}
+
+		stack, err := client.stacksAPIClient.RegisterStack(ctx, stacksapi.RegisterStackRequest{
+			Type:     stacksapi.StackTypeKubernetes,
+			QueueKey: queue,
+			Key:      stackKey,
+			Metadata: map[string]string{},
+		})
+		if err != nil {
+			return nil, fmt.Errorf("registering stack with Buildkite Stacks API: %w", err)
+		}
+
+		client.stack = stack
+	}
+
+	return client, nil
 }
 
 // AgentError is the JSON object of the response body returned when the HTTP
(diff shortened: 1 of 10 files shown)


internal/controller/controller.go
```diff
@@ -108,6 +113,48 @@ func Run(
...
+	})
+	if err != nil {
+		logger.Error("Couldn't register stack with Buildkite Stacks API", zap.Error(err))
+		return
+	}
+
+	logger.Info("registered stack with Buildkite Stacks API",
+		zap.String("stack_key", stack.Key),
+		zap.String("stack_id", stack.ID),
+	)
```
zhming0: [non-blocking] The LoC of Run method is getting long. Would it be helpful to make these new code happen in a few separated functions?

internal/controller/controller.go
```diff
@@ -241,7 +288,25 @@ func Run(
 
 	select {
 	case <-ctx.Done():
+		logger.Info("gracefully shutting down controller...")
+		ctx = context.WithoutCancel(ctx) // we want stack deregistration to happen, even though the context has been cancelled
+		ctx, cancel := context.WithTimeout(ctx, 5*time.Second)
+		defer cancel()
+
+		// Try best-effort to deregister the stack, but if it fails or times out, the backend will still clean it up.
+		err := stacksAPIClient.DeregisterStack(ctx, stack.Key, stacksapi.WithNoRetry())
```
zhming0: I wonder if there is a case where people don't want `WithNoRetry` for deregister?

If not, maybe we should make the WithNoRetry default in deregister?

internal/controller/controller.go
```diff
@@ -108,6 +113,48 @@ func Run(
...
+	maps.Copy(meta, agentTags)
+	stackQueueKey := queue
+	if stackQueueKey == "" {
+		stackQueueKey = stacksapi.DefaultQueue
+	}
+
+	stackKey := cfg.ID
+	if cfg.ID == "" {
+		stackKey = "agent-stack-k8s"
+	}
```
zhming0: I wonder if this will ever happen except in test environment?

internal/controller/controller.go
```diff
@@ -108,6 +113,48 @@ func Run(
...
+	stacksAPIClient, err := stacksapi.NewClient(agentToken, stacksapi.WithLogger(slog.New(zapSlogHandler)))
+	if err != nil {
+		logger.Error("Couldn't create Buildkite Stacks API client", zap.Error(err))
+		return
+	}
+
+	meta := map[string]string{
+		"source_location":    "https://github.com/buildkite/agent-stack-k8s",
+		"buildkite_official": "true",
+		"k8s_namespace":      cfg.Namespace,
```
zhming0: I wonder if these 3 are valuable to us? From our analytics purpose, I think we want to know: 

- K8s version. 
- Some config options, resource claims etc. 

But getting this right is probably out of scope.

internal/controller/controller.go
zhming0, line 1: I have a bit doubts about the overall approach. 

I wonder if we should allow coexistence of Agent API client and Stack API client? Maybe we should have a facade that can allow us better refactor this, this would reduce subsequent PRs blast radius, also ease us if we want to have a client side feature flag to de-risk things.

internal/controller/controller.go
```diff
@@ -108,6 +113,48 @@ func Run(
 		return
 	}
 
+	zapSlogHandler := slogzap.Option{Logger: logger}.NewZapHandler()
```
zhming0: [non-blocking] oh jeez is golang heading java's direction? 🙈

internal/controller/controller.go
moskyb: i doubt that there's a situation where people are going to want to exhaustively retry stack deregistration, but i think in terms of being a library, it's more important that policies (in this case retry policies) are applied everywhere. IMO it's least surprising to have the retry policy applied uniformly, and there's no harm in retrying it, so i'm inclined to keep it as is

internal/controller/controller.go
moskyb: after having looked into it, if we do a migration from agent api to stacks API over time, we're going to have to have them coexisting in some way, otherwise we'd need to implement our stacks API changes all in one go.

as it is, we should definitely make customers opt-in to stacks stuff while we're implementing this, and default to using the existing endpoints.

re: a façade, i think when we start replacing agent API functionality with stacks API functionality that's definitely a good idea, but for this PR, we're purely adding functionality that wasn't there before, so i don't think adding a façade makes sense right now. In the next couple PRs (where we implement `scheduled_jobs`, `job/~/state etc`, then i think a façade makes total sense.

internal/controller/controller.go
moskyb: yeah, good call — have done!

api/agent_client.go
```diff
@@ -25,20 +26,44 @@ func WithReservation(enable bool) AgentClientOption {
...
+	}
+}
+
+// AgentClient is a client for Agent API methods for retrieving jobs.
+type AgentClient struct {
+	endpoint        *url.URL
+	httpClient      *http.Client
+	stacksAPIClient *stacksapi.Client
+
+	clusterID string // or "unclustered"
```
zhming0: might be a good chance to remove this bit. 


```suggestion
	clusterID string
```

internal/controller/controller.go
```diff
@@ -108,13 +111,28 @@ func Run(
...
+		var err error
+		stacksAPIClient, err = stacksapi.NewClient(agentToken, stacksapi.WithLogger(slog.New(zapSlogHandler)))
+		if err != nil {
+			logger.Error("Couldn't create Buildkite Stacks API client", zap.Error(err))
+			return
+		}
+	}
+
 	agentClient, err := api.NewAgentClient(
+		ctx,
```
zhming0: [non-blocking] I wonder if it's better to get the api.WithStacksAPIClient to handle the `ctx` and hence the registration instead? 

but this isn't a big deal.

internal/controller/controller.go
```diff
@@ -108,13 +111,28 @@ func Run(
...
+	if cfg.ExperimentalStacksAPISupport {
+		zapSlogHandler := slogzap.Option{Logger: logger}.NewZapHandler()
+
+		var err error
+		stacksAPIClient, err = stacksapi.NewClient(agentToken, stacksapi.WithLogger(slog.New(zapSlogHandler)))
+		if err != nil {
+			logger.Error("Couldn't create Buildkite Stacks API client", zap.Error(err))
+			return
+		}
+	}
```
zhming0: stacksAPIClient doesn't seems to have any other usage in this file. I wonder if we should hide this bit in `agent_client.go`?

api/agent_client.go
```diff
@@ -263,6 +299,14 @@ func (c *AgentClient) ReserveJobs(ctx context.Context, ids []string) (result *Ba
 	return decodeResponse[BatchReserveJobsResult](resp)
 }
 
+func (c *AgentClient) DeregisterStack(ctx context.Context) error {
+	if c.stacksAPIClient == nil {
+		return nil
+	}
+
+	return c.stacksAPIClient.DeregisterStack(ctx, c.stack.Key, stacksapi.WithNoRetry())
+}
```
zhming0, line 312: [might be semi-blocking] There might be a risk in this approach. We can't know if the fact that `c.stacksAPIClient == nil` is an intended state or not.

I think ultimately we will need to access the feature flag in this client file anyway, so it might be worth to: 

```
if !featureFlag {
  return nil
}

return c.stacksAPIClient.DeregisterStack // empty pointer c will crash this. 
```

internal/controller/controller.go
```diff
@@ -241,7 +259,26 @@ func Run(
 
 	select {
 	case <-ctx.Done():
+		logger.Info("gracefully shutting down controller...")
+
+		// The DeregisterStack call is a no-op if stacks API support is disabled, but we don't want to log "deregistered stack"
+		// unless we actually did it.
+		if cfg.ExperimentalStacksAPISupport {
```
zhming0, line 251: Per my comment above, I think the feature flag check might be best to happen in the `agent_client.go`, hence I reckon some of these logic can move there too.

internal/controller/controller.go
moskyb: yeah, good call -- done!

PR #688, opened 2025-09-11
Add PodMonitor to deployment

DrJosh9000:
### What

Deploy a `PodMonitor` resource with the stack.

### Why

Following on from the inclusion of a Grafana dashboard, we are able to include a resource that would configure Prometheus Operator to scrape the metrics automatically. The controller doesn't come with a `Service`, so a `ServiceMonitor` won't work, but Prometheus Operator can scrape an individual pod using the very similar `PodMonitor` kind.

### DIFF
--- a/charts/agent-stack-k8s/values.schema.json
+++ b/charts/agent-stack-k8s/values.schema.json
@@ -32,10 +32,39 @@
       "title": "The image for the agent-stack-k8s controller",
       "examples": ["ghcr.io/buildkite/agent-stack-k8s/controller:latest"]
     },
-    "deployGrafanaDashboard": {
-      "type": "boolean",
-      "default": true,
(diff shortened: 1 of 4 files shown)


charts/agent-stack-k8s/values.schema.json
```diff
@@ -37,6 +37,29 @@
       "default": true,
       "title": "When enabled, installs a Grafana dashboard definition using a ConfigMap"
     },
+    "podMonitor": {
```
zhming0: getting this to the root is start to get confusing, because the k8s stack itself has a concept of pod monitor. 

maybe we should wrap all these under a `monitoring` field?

DrJosh9000: Sounds good, done

zhming0 (approved):

PR #692, opened 2025-09-16
PB-631 part 1: fail job API as part of stacks client

zhming0:
Add FailJob client function to stacksapi.


internal/stacksapi/fail.go
```diff
@@ -0,0 +1,30 @@
...
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+)
+
+type FailJobRequest struct {
+	StackKey    string // The key to call the stack. Required.
+	JobUUID     string // The uuid of a job
```
moskyb: these will need to be marked with `json:"-"` to work — otherwise they'll get marshalled into the json with the keys `StackKey` and `JobUUID`.

alternatively though, a pattern i like for API client libraries is to make path params be method arguments, and the whole request body as its own arg, eg something like

```Go
func (c *Client) FailJob(ctx context.Context, stackKey, jobUUID string, failJobReq FailJobRequest, opts ...RequestOption) error
```

internal/stacksapi/fail.go
```diff
@@ -0,0 +1,30 @@
...
+	ExitStatus  int    `json:"exit_status,omitempty"`
```
moskyb:
```suggestion
	ExitStatus  int    `json:"exit_status,omitzero"`
```

is probably a little clearer in intent (though not technically different in this case), and will make it easier for us to migrate to JSONv2 when (when!!!!) we do that

moskyb (dismissed):

internal/stacksapi/fail.go
```diff
@@ -0,0 +1,30 @@
...
+	"context"
+	"fmt"
+	"net/http"
+)
+
+type FailJobRequest struct {
+	StackKey    string `json:"-"` // The key to call the stack. Required.
+	JobUUID     string `json:"-"` // The uuid of a job
+	ExitStatus  int    `json:"exit_status,omitzero"`
+	ErrorDetail string `json:"error_detail"` // An error detail string less than 4KB.
```
SorchaAbel, line 13: should we use and validiation to error if `len(failJobReq.ErrorDetail) > 4kb`

zhming0: This is a great point. I think it warrant a separated PR to discuss this detail. 

Because I don't think it has to be as simple as throwing back the the error back to caller, depending our error handling philosophy, we have a few options. Anyway, I will raise a different PR.

zhming0: Here: https://github.com/buildkite/agent-stack-k8s/pull/698

PR #694, opened 2025-09-17
PB-680: add a test step to test with experiment feature flag on

zhming0:
[In-progress]


.buildkite/pipeline.yaml
```diff
@@ -164,6 +164,52 @@ steps:
...
+    artifact_paths: junit-*.xml
+    image: golang:latest
+    command: .buildkite/steps/tests.sh
+    env:
+      CONFIG: /etc/config.yaml
+      EXPERIMENTAL_JOB_RESERVATION_SUPPORT: true
+      EXPERIMENTAL_STACKS_API_SUPPORT: true
+    plugins:
+      - kubernetes:
+          podSpecPatch:
```
SorchaAbel, line 178: I wonder for consistency should a podSpec be used here instead of a podSpecPatch. 
I think it needs either a complete podSpec OR step-level configuration with podSpecPatch.

.buildkite/pipeline.yaml
```diff
@@ -164,6 +164,52 @@ steps:
...
+            serviceAccountName: integration-tests
+            volumes:
+              - name: agent-stack-k8s-config
+                configMap:
+                  name: agent-stack-k8s-config
+              - *go-cache-volume
+              - *go-mod-cache-volume
+            containers:
+              - name: container-0
```
SorchaAbel, line 187: could the name be something more descriptive like `tests` :)

.buildkite/pipeline.yaml
zhming0: yeah in this particular case, podSpec doesn't work. It's a rough edge I guess. 

but I am not concerned, because our end goal is to eliminate podSpec and podSpecPatch from these pipelines.

.buildkite/pipeline.yaml
zhming0: This is yet another rough edge of the product, in order to get the image attribute work, the main container has to be named container-0. 

We have an ongoing reported issue about this. The fix will come later.

.buildkite/pipeline.yaml
```diff
@@ -164,6 +164,52 @@ steps:
           files: junit-*.xml
           format: junit
 
+  - label: ":buildkite::test_tube::rocket: tests (experimental features)"
+    key: tests-experimental
```
DrJosh9000, line 168: Should this new step be a soft-fail?

```suggestion
    key: tests-experimental
    soft_fail: true
```

zhming0: I guess this depends on if we intend to merge broken code in main? I don't think we should?

.buildkite/pipeline.yaml
```diff
@@ -164,6 +164,52 @@ steps:
...
+    depends_on: agent
+    artifact_paths: junit-*.xml
+    image: golang:latest
```
DrJosh9000: It seems confusing to have both `image` here and a `podSpecPatch`, should we keep this detail in `podSpecPatch` for now? Is there still a bug with step `image` not creating image check containers?

.buildkite/pipeline.yaml
DrJosh9000: On the other hand, do we want to block a release if the experimental code isn't working?

DrJosh9000: I don't particularly mind either way, just wanted to discuss it.

zhming0: Yeah good question, IMO if our next target is releasing an agent k8s stack that has the experimental feature flag on (which I think it's true?). Then we should probably raise the bar now to avoid big bang surprise.

.buildkite/pipeline.yaml
zhming0: kk I moved the `image` bit back to `podSpecPatch`. 

I reckon I will just leave the pipeline cleanup to the future so I can eliminate all `podSpec` in one go. 

As for the image check container issue, I haven't got time to confirm that.

PR #696, opened 2025-09-18
Monitoring tweaks

DrJosh9000:
### What

- Enable Prometheus /metrics by default
- Apply the `monitoring.namespace` to both the `PodMonitor` and the Grafana dashboard `ConfigMap`
- Remove the `${DS_PROMETHEUS}` nonsense from the dashboard, as it seems to prevent the dashboard using the default Prometheus service

### Why

Things I found while trying to use these autodeployed resources.


zhming0 (approved):
I am not familiar with the Grafana dashboard configmap syntax, so I am approving on that change blindly 😬 . 

The rest look good to me.

PR #698, opened 2025-09-19
PB-631-follow-up: crop FailJob error detail when it exceed size limit

zhming0:
Subject to our error handling philosophy, this may or may not be desirable. 

Generally speaking, I personally believe: 

- Failing gracefully gets better UX. 
- Cropped diagnostic data is better than no diagnostic data. 

I also acknowledge the downside

- it may cause surprise. 
- cropped diagnostic data in some certain case, can get misleading. 

But I reckon the pros outweight the gain. I welcome feedbacks!

### DIFF
--- a/internal/stacksapi/fail.go
+++ b/internal/stacksapi/fail.go
@@ -14,6 +14,13 @@ type FailJobRequest struct {
 }
 
 func (c *Client) FailJob(ctx context.Context, failJobReq FailJobRequest, opts ...RequestOption) error {
+	const maxErrorDetailSize = 4 * 1024 // 4KB
+
+	if len(failJobReq.ErrorDetail) > maxErrorDetailSize {
+		c.logger.Warn("ErrorDetail exceeds 4KB limit, cropping", "original_size", len(failJobReq.ErrorDetail), "cropped_size", maxErrorDetailSize)
+		failJobReq.ErrorDetail = failJobReq.ErrorDetail[:maxErrorDetailSize]
+	}
+
 	path := fmt.Sprintf("/stacks/%s/jobs/%s/fail", failJobReq.StackKey, failJobReq.JobUUID)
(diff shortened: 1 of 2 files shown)


internal/stacksapi/fail.go
```diff
@@ -14,6 +14,13 @@ type FailJobRequest struct {
 }
 
 func (c *Client) FailJob(ctx context.Context, failJobReq FailJobRequest, opts ...RequestOption) error {
+	const maxErrorDetailSize = 4 * 1024 // 4KB
+
+	if len(failJobReq.ErrorDetail) > maxErrorDetailSize {
+		c.logger.Warn("ErrorDetail exceeds 4KB limit, cropping", "original_size", len(failJobReq.ErrorDetail), "cropped_size", maxErrorDetailSize)
+		failJobReq.ErrorDetail = failJobReq.ErrorDetail[:maxErrorDetailSize]
+	}
+
```
moskyb, line 24: Maybe something like:

```suggestion
func (c *Client) FailJob(ctx context.Context, failJobReq FailJobRequest, opts ...RequestOption) error {
	const croppedMessage = "… (Error message was cropped because it exceeded the max size of 4KB)"
	const maxErrorDetailSize = (4 * 1024) - len(croppedMessage) // 4KB

	if len(failJobReq.ErrorDetail) > maxErrorDetailSize {
		c.logger.Warn("ErrorDetail exceeds 4KB limit, cropping", "original_size", len(failJobReq.ErrorDetail), "cropped_size", maxErrorDetailSize)
		failJobReq.ErrorDetail = failJobReq.ErrorDetail[:maxErrorDetailSize] + croppedMessage
	}

```

to let people reading the error message know that it was cropped?

also worth noting here that this code would mean that we can never reduce the (serverside) max size of the error detail, but i think that's worth it for better error ergonomics

zhming0: Good idea!

PR #700, opened 2025-09-22
PB-628 part 1: refactor pod_watcher, splitting out bk job checker

zhming0:
This PR has no behavior change, it only decouple pod_watcher from bk job cancellation checker so we can subsequently refactor them without too much diff.

### DIFF
--- a/internal/controller/scheduler/buildkite_job_checker.go
+++ b/internal/controller/scheduler/buildkite_job_checker.go
@@ -1,0 +1,165 @@
+package scheduler
+
+import (
+	"context"
+	"sync"
+	"time"
+
+	"github.com/buildkite/agent-stack-k8s/v2/api"
+	"github.com/google/uuid"
+	"go.uber.org/zap"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+)
+
+// BuildkiteJobChecker monitors Buildkite jobs for cancellation state changes.
+type BuildkiteJobChecker struct {
+	logger      *zap.Logger
+	agentClient *api.AgentClient
+	k8s         kubernetes.Interface
+
+	// The job cancel checkers query the job state every so often.
+	jobCancelCheckerInterval time.Duration
+
+	// Channels that are closed when a cancel checker should stop.
+	cancelCheckerChsMu sync.Mutex
+	cancelCheckerChs   map[uuid.UUID]*onceChan
+}
+
+// NewBuildkiteJobChecker creates a new Buildkite job checker.
+func NewBuildkiteJobChecker(
+	logger *zap.Logger,
+	agentClient *api.AgentClient,
+	k8s kubernetes.Interface,
+	interval time.Duration,
+) *BuildkiteJobChecker {
+	return &BuildkiteJobChecker{
+		logger:                   logger,
+		agentClient:              agentClient,
+		k8s:                      k8s,
+		jobCancelCheckerInterval: interval,
+		cancelCheckerChs:         make(map[uuid.UUID]*onceChan),
+	}
+}
+
+// StartChecking starts monitoring a job for cancellation.
+// This should only be called for jobs in pending state.
+func (c *BuildkiteJobChecker) StartChecking(ctx context.Context, log *zap.Logger, podMeta metav1.ObjectMeta, jobUUID uuid.UUID) {
+	c.cancelCheckerChsMu.Lock()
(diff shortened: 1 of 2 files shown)


internal/controller/scheduler/buildkite_job_checker.go
```diff
@@ -0,0 +1,165 @@
...
+	logger      *zap.Logger
+	agentClient *api.AgentClient
+	k8s         kubernetes.Interface
+
+	// The job cancel checkers query the job state every so often.
+	jobCancelCheckerInterval time.Duration
+
+	// Channels that are closed when a cancel checker should stop.
+	cancelCheckerChsMu sync.Mutex
+	cancelCheckerChs   map[uuid.UUID]*onceChan
```
zhming0, line 26: @DrJosh9000 it isn't obvious to me if these channels are actually used? I wonder if a `bool` will suffice here. 

Do you know some context here?

DrJosh9000: The channels here are a shutdown mechanism for each cancel checker. Each worker goroutine uses a `select` to sleep (with a read on its shutdown channel as a `case`), which means it can shut down right away when the channel is closed. Compare that to using a bool: it would only be able to check the bool between sleeps, which could delay shutdown.

But actually, a bool would probably be fine. A goroutine waiting around an extra second is hardly going to kill us.

zhming0: ahh, I see the stopCh mechanism there, I don't know why I missed that. 

Yeah, once we refactor job checker into batched, we are probably going to replace these channels with a `map[string]*k8s.ObjectMeta`.

internal/controller/scheduler/bk_job_checker.go
```diff
@@ -0,0 +1,165 @@
...
+
+	"github.com/buildkite/agent-stack-k8s/v2/api"
+	"github.com/google/uuid"
+	"go.uber.org/zap"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+)
+
+// BkJobChecker monitors Buildkite jobs for cancellation state changes.
+type BkJobChecker struct {
```
DrJosh9000: Hmm. I think perhaps one of these. "Buildkite" tends to be shortened to "BK", not "Bk", don't you think?

```suggestion
type BKJobChecker struct {
```

```suggestion
type BuildkiteJobChecker struct {
```

internal/controller/scheduler/pod_watcher.go
```diff
@@ -390,19 +389,13 @@ func (w *podWatcher) forcefullyDeletePod(ctx context.Context, log *zap.Logger, p
...
 
-	if err := w.k8s.CoreV1().Pods(podMetadata.Namespace).Delete(ctx, podMetadata.Name, deleteOptions); err != nil {
+	if err := k8s.CoreV1().Pods(podMetadata.Namespace).Delete(ctx, podMetadata.Name, deleteOptions); err != nil {
 		log.Error("Couldn't forcefully delete pod", zap.Error(err))
 		forcefulPodDeletionErrorsCounter.WithLabelValues(reason, string(kerrors.ReasonForError(err))).Inc()
 		return err
 	}
 	forcefullyDeletedPodCounter.WithLabelValues(reason).Inc()
 
-	// Mark the job as ignored to avoid further processing
```
DrJosh9000, line 400: What's happening here?

zhming0: This bit of code is moved to  [here](https://github.com/buildkite/agent-stack-k8s/blob/3ea7b4e8c25f41f11525659c68c49739611e0a94/internal/controller/scheduler/pod_watcher.go#L506-L512). So this function can stay pure and easier to reuse.

DrJosh9000: I see, thanks

internal/controller/scheduler/bk_job_checker.go
zhming0: Good point, I reckon BuildkiteJobChecker is better. I renamed the file and the type.

internal/controller/scheduler/buildkite_job_checker.go
```diff
@@ -0,0 +1,165 @@
...
+}
+
+// NewBkJobChecker creates a new Buildkite job checker.
+func NewBkJobChecker(
```
DrJosh9000: This should correspond to the name of the return type:

```suggestion
func NewBuildkiteJobChecker(
```

PR #701, opened 2025-09-22
PB-628 part 2: agentclient.(GetJobState -> GetJobStates)

zhming0:
based on top of #700.

This PR changes agentclient.GetJobState -> agentclient.GetJobStates, but it doesn't change the API endpoint. It's still relying the old API.

### DIFF
--- a/api/agent_client.go
+++ b/api/agent_client.go
@@ -242,7 +242,7 @@ type AgentJobState struct {
 }
 
 // GetJobState gets the state of a specific job.
-func (c *AgentClient) GetJobState(ctx context.Context, id string) (result *AgentJobState, retryAfter time.Duration, err error) {
+func (c *AgentClient) getJobState(ctx context.Context, id string) (result *AgentJobState, retryAfter time.Duration, err error) {
 	u := c.endpoint.JoinPath(
 		"clusters", railsPathEscape(c.clusterID),
 		"queues", railsPathEscape(c.queue),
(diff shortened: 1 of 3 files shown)


api/agent_client.go
```diff
@@ -259,6 +259,88 @@ func (c *AgentClient) GetJobState(ctx context.Context, id string) (result *Agent
 	return decodeResponse[AgentJobState](resp)
 }
 
+// GetJobStats get the states of a batch of jobs.
+func (c *AgentClient) GetJobStates(ctx context.Context, ids []string) (result map[string]JobState, retryAfter time.Duration, err error) {
+	if len(ids) == 0 {
+		return make(map[string]JobState), 0, nil
+	}
+
+	const batchSize = 20
```
DrJosh9000, line 268: Why 20?

api/agent_client.go
```diff
@@ -259,6 +259,88 @@ func (c *AgentClient) GetJobState(ctx context.Context, id string) (result *Agent
...
+	stateMap := make(map[string]JobState)
+	var maxRetryAfter time.Duration
+	var firstError error
+
+	for i := 0; i < len(ids); i += batchSize {
+		end := i + batchSize
+		if end > len(ids) {
+			end = len(ids)
+		}
+		batch := ids[i:end]
```
DrJosh9000, line 278: That spells it out, but there's probably a slicier way. Something like...


```suggestion
	for len(ids) > 0 {
	    end := min(batchSize, len(ids))
		batch := ids[:end]
		ids = ids[end:]
```

zhming0:
@DrJosh9000 I had a bit think about where this is going and I don't like it. I am going to close this down and start from a new direction (part 1 stays though) 

I was hoping to make a single GetjobStates method that works for both old and new API, but now I think it introduces unnecessary risks and work.

I think a safer way would be implementing a `BatchedBuildkiteJobChecker`.

PR #703, opened 2025-09-22
PB-626 Get scheduled jobs using the stacks API

moskyb:
This PR updates the k8s stack so that when stacks API usage is enabled, the stack will use those endpoints to fetch scheduled jobs instead of the old ones under `clusters/~/queues/~/scheduled_jobs`. Its usage of the API remains the same.


internal/stacksapi/fail.go
zhming0, line 1: It isn't very obvious to me on the rationale behind getting the client method to return `http.Header`, on the first glance, it feels like an abstraction leak? 

But this PR is WIP, I will come back with another round of review when description are filled. 👀

internal/stacksapi/jobs.go
```diff
@@ -0,0 +1,130 @@
...
+	}
+
+	return listJobsResp, header, nil
+}
+
+// PaginateAllScheduledJobs retrieves all jobs in the scheduled state from a specific cluster queue, handling pagination
+// automatically. The StackKey and ClusterQueueKey fields of the request are required. This function will make multiple
+// calls to [ListScheduledJobs] as needed to retrieve all jobs, and will return a single response containing all jobs.
+// The http.Header from the last request made is also returned.
+func (c *Client) PaginateAllScheduledJobs(ctx context.Context, listReq ListScheduledJobsRequest, opts ...RequestOption) (ListScheduledJobsResponse, http.Header, error) {
```
zhming0: I don't think we want to offer this method as part of our client. Intuitively it seems useful but it might be a foot gun.

There are a few reasons: 
- This isn't atomic operation. So there really isn't a "all" concept here. In our monitor code, we are very careful constantly back tracking to ensure no missed jobs. 
- Error handling is tricky, say if users have 30 pages of jobs, if they call this method, it might never work because they would've trigger rate limit every time. 
- Last in terms of rule of thumb, we almost never want a Stack to do a big bang 100K job spawn wave.

moskyb: yeah, i get that. i worry though that this if we put this library (and this API) into the hands of customers, paginating through the list of jobs is the first thing they're going to do.

i'm on board with getting rid of this method, but i wonder if we should replace it with exporting the `JobMonitor` or something similar to it as part of this package — eg, have something like
```Go
func (c *Client) PollForScheduledJobs(ctx context.Context) (<-chan ScheduledJob, error) {
  jobsCh := make(chan ScheduledJob)
  
  go func() {
    for {
      jobs, err := somePollThingThatHandlesErrorsAndBackoffs(ctx)
      for _, j := range jobs {
        jobsCh <- j
      }
  }

  return jobsCh, nil
}

Comment (moskyb) [reply internal/stacksapi/jobs.go]: "i wonder if we should replace it with..." means "not in this PR", btw

Comment (zhming0) [reply internal/stacksapi/jobs.go]: Yeah that feels useful, in a sense we will be providing a Stack SDK rather than API client.

File: internal/stacksapi/scheduled_jobs.go
Comment (zhming0) [inline internal/stacksapi/scheduled_jobs.go:1]: [not-blocking] I wonder if we should call it `scheduled_jobs`? Just so it can be more specific and it won't raise the future later like "does reserve jobs endpoint goes to the jobs.go?"

File: internal/stacksapi/jobs.go [hunk 2]
```diff
@@ -0,0 +1,96 @@
...
+
+// Step represents the step that created a job.
+type Step struct {
+	Key string `json:"key"`
+}
+
+// ScheduledJob represents the metadata for a job that's in the scheduled state.
+type ScheduledJob struct {
+	ID              string    `json:"id"`                // The UUID of the job
+	Priority        int       `json:"priority"`          // The priority of the job; higher priority jobs are scheduled first
```
zhming0: [non-blocking] 

```suggestion
	Priority        int       `json:"priority"`          // The priority of the job; higher priority jobs should be scheduled first
```

So we made it clear its caller's responsibility to respect priority.

internal/stacksapi/jobs.go
```diff
@@ -0,0 +1,96 @@
...
+	AgentQueryRules []string  `json:"agent_query_rules"` // The agent query rules that must be matched to run the job
```
zhming0: [non-blocking] 

```suggestion
	AgentQueryRules []string  `json:"agent_query_rules"` // The agent tags that must be matched to run the job
```

might be slightly more accurate?

internal/stacksapi/jobs.go
```diff
@@ -0,0 +1,96 @@
...
+	StackKey        string // The key of the stack calling this endpoint. Required
+	ClusterQueueKey string // The key of the cluster queue to list jobs from. Required
+	PageSize        int    // The number of jobs to return. Optional, defaults to a sensible value on the backend
+	StartCursor     string // The cursor to start the page from. Optional, defaults to the start of the list
+}
+
+// ListScheduledJobs lists jobs in the scheduled state from a specific cluster queue. The StackKey and ClusterQueueKey
+// fields of the request are required. This method will only return a single page of results. To retrieve additional
+// pages, call this method again with the EndCursor field set to the EndCursor value from the previous response's PageInfo,
+// or use the [PaginateAllScheduledJobs] function to automatically handle pagination.
```
zhming0:
```suggestion
```

internal/stacksapi/register.go
```diff
@@ -29,31 +29,32 @@ type RegisterStackResponse struct {
 
 // RegisterStack registers a new stack with the Buildkite Stacks API. A stack with the same key can safely be
 // re-registered as many times as necessary.
-func (c *Client) RegisterStack(ctx context.Context, body RegisterStackRequest, opts ...RequestOption) (*RegisterStackResponse, error) {
+func (c *Client) RegisterStack(ctx context.Context, body RegisterStackRequest, opts ...RequestOption) (*RegisterStackResponse, http.Header, error) {
```
zhming0, line 32: [non-blocking] Just wanting to note down that exposing header might be a blocker if we expose the client to 3rd party use. It's good now 🚢

api/agent_client.go
```diff
@@ -185,29 +193,72 @@ func (c *AgentClient) GetScheduledJobs(ctx context.Context, afterCursor string,
...
-				// job queue key "" vs our configuration queue key "default".
-				// So this is forcing them to be equal.
-				j.AgentQueryRules = agenttags.RemoveTag(j.AgentQueryRules, "queue")
-			} else {
-				// Ensure the job has the queue tag. We queried a queue-specific
-				// endpoint, but it may be the default queue, which doesn't require
-				// `agents: queue: ...`, so the queue tag might not be present.
-				j.AgentQueryRules = agenttags.SetTag(j.AgentQueryRules, "queue", c.queue)
-			}
+			j.AgentQueryRules = c.normaliseAgentQueryRules(j.AgentQueryRules)
```
zhming0, line 201: [non-blocking, but would be nice to address] I see we duplicated QueriedAt and AgentQueryRules handling. It's not a blocker but it would be very nice to centralize it. Because this is a highly impact area that frequently incur issues.

api/agent_client.go
```diff
@@ -185,29 +193,72 @@ func (c *AgentClient) GetScheduledJobs(ctx context.Context, afterCursor string,
...
 	return result, retryAfter, err
 }
 
-// AgentJob is enough info to run a job (includes command and env).
+func (c *AgentClient) normaliseAgentQueryRules(rules []string) []string {
+	if c.queue == defaultQueueKey {
+		// When we poll from default queue, we don't know the queue key, so in rest of the system queue="".
+		// The job might contain a queue key `agents: queue: default`, in that case it will cause mismatch in local
+		// job queue key "" vs our configuration queue key "default".
+		// So this is forcing them to be equal.
```
zhming0: > So this is forcing them to be equal.

just wanting to call out that, after refactor this code comment doesn't make sense without context. 

This particular code comment really only make sense to read around the code block that mutate job agent query rules.

zhming0 (approved):
Looks good! I see no blockers therefore it's a tick from me ✅ . 

But addressing all comments would be awesome 🙏🏿

internal/stacksapi/scheduled_jobs.go
moskyb: ooh yep, good call.

api/agent_client.go
moskyb: yeah, my logic here was that the period of time that we're going to have both of these endpoints active is going to be relatively short -- better to have a little bit of duplication than the wrong abstraction, etc. when we get rid of the non-stacks API endpoints, this problem will solve itself.

PR #705, opened 2025-09-23
stacksapi: consume and close every response body

DrJosh9000:
### What

Rearrange the Stacks API client so that the HTTP response body is always fully consumed and closed. With a little generics as a treat.

### Why

The existing `Client.do` method returns a `*http.Response`. It's not documented whether the caller is responsible for closing the response body, and so already the surrounding layer of code is defensively doing a `defer resp.Body.Close()`. It's not necessary as the inner code is closing them before returning, but nevertheless it turns out we _are_ leaking response bodies in the retry loop, which the caller can't do anything about. 

Instead we could handle all that inside `do`, which is what this PR does.

### DIFF
--- a/internal/stacksapi/client.go
+++ b/internal/stacksapi/client.go
@@ -201,54 +201,58 @@ func (c *Client) newRequest(ctx context.Context, method, path string, body any, 
 	return stackReq, nil
 }
 
-// do performs the given *StackAPIRequest and returns the HTTP response, unmarshalling the body into val.
-func (c *Client) do(ctx context.Context, req *StackAPIRequest, val any) (*http.Response, error) {
-	resp, err := c.doWithRetry(ctx, req)
-	if err != nil {
-		return nil, err
-	}
-
-	if err := c.processResponse(ctx, resp, val); err != nil {
-		return nil, err
-	}
-
-	return resp, nil
-}
-
-// doWithRetry handles the retry logic for HTTP requests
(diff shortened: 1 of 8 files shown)


internal/stacksapi/client.go
```diff
@@ -201,54 +201,58 @@ func (c *Client) newRequest(ctx context.Context, method, path string, body any,
...
+// It is not suitable for large (streaming) response bodies, as it fully
+// consumes the response body into memory.
+func do[Resp any](ctx context.Context, c *Client, req *StackAPIRequest) (*Resp, http.Header, error) {
 	logger := c.logger.With(
 		"method", req.Method,
 		"url", req.URL.String(),
 	)
 
-	return roko.DoFunc(ctx, req.retrier, func(r *roko.Retrier) (*http.Response, error) {
+	return roko.DoFunc2(ctx, req.retrier, func(r *roko.Retrier) (*Resp, http.Header, error) {
```
moskyb, line 215: why return a `*Resp` rather than a `Resp`?

DrJosh9000: I guess I'm used to doing that, and initially I wrote something like `val := new(T)`, which is a `*T`. But now it could just return `val`.

A nil `*Resp` is definitely not a valid `Resp`, so returning `*Resp` might help uncover cases where calling code accidentally relying on the value when it shouldn't by making that a null pointer panic. 

I don't mind changing it though.

moskyb: yeah, i'm on board with that, let's keep it as a `*Resp`

PR #708, opened 2025-09-24
PB-628 part 2 attempt 2: BatchBuildkiteJobChecker and new GetJobStates method

zhming0:
This PR is probably one of the more evolved bit of migrating to Stack API. 

Stack API changes how client should get the job state from single fetch to batched fetch. Therefore, the existing single job oriented go routine approach won't work anymore.

This PR introduces `BatchBuildkiteJobChecker` check job stacks by 1000 per batch. 

This PR also implemented the underlying `GetJobStates` method.

### DIFF
--- a/api/agent_client.go
+++ b/api/agent_client.go
@@ -33,7 +33,7 @@ type AgentClient struct {
 
 	// This impacts a number of endpoints' query parameters
 	reservation bool
-	UseStackAPI bool
+	useStackAPI bool
 }
 
 type AgentClientOpts struct {
@@ -75,7 +75,7 @@ func NewAgentClient(ctx context.Context, opts AgentClientOpts) (*AgentClient, er
 		clusterID:   opts.ClusterID,
 		queue:       opts.Queue,
 		reservation: opts.UseReservation,
-		UseStackAPI: opts.UseStacksAPI,
+		useStackAPI: opts.UseStacksAPI,
 	}
 
 	if opts.UseStacksAPI {
@@ -249,6 +249,18 @@ type AgentJobState struct {
 
 // GetJobState gets the state of a specific job.
 func (c *AgentClient) GetJobState(ctx context.Context, id string) (result *AgentJobState, retryAfter time.Duration, err error) {
+	// Use the batch API as a shim when useStackAPI is on.
+	if c.UseStackAPI() {
+		state, retryAfter, err := c.GetJobStates(ctx, []string{id})
+		if err != nil {
+			return nil, retryAfter, err
+		}
+		return &AgentJobState{
+			ID:    id,
+			State: state[id],
+		}, retryAfter, nil
+	}
+
 	u := c.endpoint.JoinPath(
 		"clusters", railsPathEscape(c.clusterID),
 		"queues", railsPathEscape(c.queue),
@@ -263,6 +275,28 @@ func (c *AgentClient) GetJobState(ctx context.Context, id string) (result *Agent
 	defer resp.Body.Close()
 
 	return decodeResponse[AgentJobState](resp)
+}
+
+// GetJobStates gets state for a batch of jobs.
+func (c *AgentClient) GetJobStates(ctx context.Context, ids []string) (result map[string]JobState, retryAfter time.Duration, err error) {
+	if !c.UseStackAPI() {
+		// We don't expect this to happen in prod. It's mainly a defensive mechanism.
+		return nil, 0, errors.New("cannot use batch job state loading without Stack API")
+	}
+	req := stacksapi.GetJobStatesRequest{
+		StackKey: c.stack.Key,
+		JobUUIDs: ids,
+	}
+	statesResp, err := c.stacksAPIClient.GetJobStates(ctx, req)
+	if err != nil {
+		return nil, 0, err
+	}
+	result = make(map[string]JobState, len(statesResp.States))
+	for k, v := range statesResp.States {
+		result[k] = JobState(v)
+	}
+	// FIXME: fix the retryAfter later
+	return result, 0, err
 }
 
 // ReserveJobBatchResult describes the result of a batch job reservation.
@@ -318,8 +352,13 @@ func (c *AgentClient) DeregisterStack(ctx context.Context) error {
 	return c.stacksAPIClient.DeregisterStack(ctx, c.stack.Key, stacksapi.WithNoRetry())
 }
 
+// UseStackAPI returns whether the client is configured to use the Stack API.
+func (c *AgentClient) UseStackAPI() bool {
+	return c.useStackAPI
+}
+
 func (c *AgentClient) FailJob(ctx context.Context, jobUUID string, errorDetail string) error {
-	if !c.UseStackAPI {
+	if !c.UseStackAPI() {
 		return fmt.Errorf("cannot call FailJob when useStackAPI is off")
 	}
 
--- a/internal/controller/scheduler/batch_buildkite_job_checker.go
+++ b/internal/controller/scheduler/batch_buildkite_job_checker.go
@@ -1,0 +1,173 @@
+package scheduler
+
+import (
+	"context"
+	"math"
+	"slices"
+	"sync"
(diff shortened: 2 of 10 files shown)


api/agent_client.go
```diff
@@ -265,6 +277,25 @@ func (c *AgentClient) GetJobState(ctx context.Context, id string) (result *Agent
...
+		StackKey: c.stack.Key,
+		JobUUIDs: ids,
+	}
+	states, err := c.stacksAPIClient.GetJobStates(ctx, req)
+	result = make(map[string]JobState, len(states))
+	for k, v := range states {
+		result[k] = JobState(v)
+	}
+	// FIXME: fix the retryAfter later
+	return result, 0, err
```
moskyb, line 299: the FIXME here should probably be fixed, right?

internal/controller/scheduler/batch_buildkite_job_checker.go
```diff
@@ -0,0 +1,180 @@
...
+	"github.com/google/uuid"
+	"go.uber.org/zap"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+)
+
+// BuildkiteJobChecker monitors Buildkite jobs for cancellation state changes.
+// It operate on individual jobs basis, relying on legacy Agent API.
+// In the future version, we may get rid of it.
+type BatchBuildkiteJobChecker struct {
```
moskyb: this comment looks like a copypaste — is this the batch one or the individual one?

internal/controller/scheduler/batch_buildkite_job_checker.go
```diff
@@ -0,0 +1,180 @@
...
+
+	// Split jobs into batches of 1000 and process concurrently
+	const batchSize = 1000
+	batches := make([][]string, 0, (len(jobUUIDs)+batchSize-1)/batchSize)
+
+	for i := 0; i < len(jobUUIDs); i += batchSize {
+		end := i + batchSize
+		end = min(end, len(jobUUIDs))
+		batches = append(batches, jobUUIDs[i:end])
+	}
```
moskyb: we can use [`slices.Chunk`](https://pkg.go.dev/slices#Chunk) here:

```suggestion
	batches := slices.Chunk(jobUUIDs, 1000)
```

internal/controller/scheduler/batch_buildkite_job_checker.go
```diff
@@ -0,0 +1,180 @@
...
+	var wg sync.WaitGroup
+	jobStatesCh := make(chan map[string]api.JobState, len(batches))
+
+	for _, batch := range batches {
+		wg.Add(1)
+		go func(batchUUIDs []string) {
+			defer wg.Done()
+
+			jobStates, retryAfter, err := c.agentClient.GetJobStates(ctx, batchUUIDs)
+			if api.IsPermanentError(err) {
```
moskyb: i _think_ `api.IsPermanentError` will return false if the error is a `stacksapi.ResponseError` that's permanent — we might need to add a clause into it like
```Go
if stackRespErr, ok := errors.As(err, stacksapi.ResponseError); ok {
  return !stackRespErr.IsRetryableStatus()
}
```

internal/stacksapi/get_job_states.go
```diff
@@ -0,0 +1,29 @@
...
+)
+
+type GetJobStatesRequest struct {
+	StackKey string   `json:"-"`         // The key to call the stack. Required.
+	JobUUIDs []string `json:"job_uuids"` // A list of job uuids
+}
+
+type GetJobStatesResponse map[string]string
+
+func (c *Client) GetJobStates(ctx context.Context, getJobStatesReq GetJobStatesRequest, opts ...RequestOption) (GetJobStatesResponse, error) {
```
moskyb, line 19: All of these should have doc comments i think

internal/controller/scheduler/batch_buildkite_job_checker.go
```diff
@@ -0,0 +1,180 @@
...
+		k8s:                      k8s,
+		jobCancelCheckerInterval: interval,
+		checkingJobs:             make(map[uuid.UUID]metav1.ObjectMeta),
+	}
+}
+
+// Designed to run in a goroutine. It starts a loop to periodically check job states for all jobs that it manages.
+func (c *BatchBuildkiteJobChecker) StartChecking(ctx context.Context) {
+	c.logger.Debug("Starting batch job state checker")
+	defer c.logger.Debug("Stopped batch job state checker")
```
moskyb: i reckon if this is always intended to be run in a goroutine, it should just start that goroutine itself rather than relying on the caller

api/agent_client.go
```diff
@@ -249,6 +249,18 @@ type AgentJobState struct {
 
 // GetJobState gets the state of a specific job.
 func (c *AgentClient) GetJobState(ctx context.Context, id string) (result *AgentJobState, retryAfter time.Duration, err error) {
+	// Use the batch API as a shim when useStackAPI is on.
+	if c.useStackAPI {
```
moskyb: i think if we have the reader method (`c.UseStackAPI()`) we should probably use it even internally

moskyb (changes requested):
this looks good! the design of the batch checker is really nice, i like it a lot.

there's a couple things we'll need to change before we merge, but in concept this looks really good.

moskyb (changes requested):
this looks good! the design of the batch checker is really nice, i like it a lot.

there's a couple things we'll need to change before we merge, but in concept this looks really good.

api/agent_client.go
zhming0: I am planning to address it in a follow-up PR after yours being merged.

internal/controller/scheduler/batch_buildkite_job_checker.go
zhming0: As I am addressing this comment, I realize the `IsRetryableStatus` check here is unnecessary. Since we already have retry on client level, we don't really need to differentiate how we handle the final error.

internal/controller/scheduler/batch_buildkite_job_checker.go
zhming0: Very true 👍🏿

internal/stacksapi/get_job_states.go
zhming0: Added a simple doc.

internal/controller/scheduler/batch_buildkite_job_checker.go
zhming0: Neat!

api/agent_client.go
moskyb: sweet, easy as

internal/controller/scheduler/batch_buildkite_job_checker.go
```diff
@@ -0,0 +1,173 @@
...
+			defer wg.Done()
+
+			jobStates, _, err := c.agentClient.GetJobStates(ctx, batchUUIDs)
+			if err != nil {
+				c.logger.Error("Couldn't fetch states of jobs", zap.Error(err), zap.Int("batch_size", len(batchUUIDs)))
+				return
+			}
+
+			jobStatesCh <- jobStates
+		}(batch)
```
moskyb: not a blocker:
```suggestion
		go func() {
			defer wg.Done()

			jobStates, _, err := c.agentClient.GetJobStates(ctx, batch)
			if err != nil {
				c.logger.Error("Couldn't fetch states of jobs", zap.Error(err), zap.Int("batch_size", len(batch)))
				return
			}

			jobStatesCh <- jobStates
		}()
```
i think post-Go 1.23 we don't need to make this a function arg anymore

moskyb (approved):

internal/controller/scheduler/batch_buildkite_job_checker.go
zhming0: Ah right the 1.23 loop fix! Good point.

PR #711, opened 2025-09-24
Fix get job states

moskyb:
Some merge shenanigans left us with code that wasn't buildable. 

Also, make the `GetJobStates` method return a pointer to its output type, rather than the value. this doesn't have an affect, but it's more consistent with the other methods.


zhming0 (approved):
👍🏿

DrJosh9000 (approved):

PR #713, opened 2025-09-24
PB-630: Get GetJobToRun to use Stack API

zhming0:
Self explanatory 😄 .

This API changes agent_client.GetJobToRun so when UseStackAPI is true, it would use the new stack api method `GetJob` to get job.

### DIFF
--- a/api/agent_client.go
+++ b/api/agent_client.go
@@ -129,6 +129,10 @@ func IsPermanentError(err error) bool {
 	if err == nil {
 		return false
 	}
+	var stackResponseError *stacksapi.ErrorResponse
+	if errors.As(err, &stackResponseError) {
+		return !stackResponseError.IsRetryableStatus()
+	}
 	var agentErr AgentError
 	if !errors.As(err, &agentErr) {
 		return false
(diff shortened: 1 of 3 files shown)


internal/stacksapi/get_job.go
```diff
@@ -0,0 +1,34 @@
...
+
+type GetJobResponse struct {
+	ID      string            `json:"id"`
+	Env     map[string]string `json:"env"`
+	Command string            `json:"command"`
+}
+
+// GetJob return Job's runtime data (env and command) for a single job.
+func (c *Client) GetJob(ctx context.Context, getJobReq GetJobRequest, opts ...RequestOption) (*GetJobResponse, http.Header, error) {
+	path := fmt.Sprintf("/stacks/%s/jobs/%s", getJobReq.StackKey, getJobReq.JobUUID)
```
DrJosh9000: Similar comment that I made in #716: if `StackKey` is customer-supplied, then we should be sure to %-escape `.` - see https://github.com/buildkite/agent/blob/4e9dc55fb68e89a0ee9c8994f3afe48d27b7d01d/api/client.go#L418-L421

moskyb: i've fixed this in [Escape path params in the stacks API](https://github.com/buildkite/agent-stack-k8s/pull/716/commits/c96693fa6c0dedca8df7607c51cadf6d6347b610) — can rebase onto that once it's merged if you like @zhming0

zhming0: Thanks! I have updated it to use `constructPath`

internal/stacksapi/get_job.go
```diff
@@ -0,0 +1,34 @@
...
+func (c *Client) GetJob(ctx context.Context, getJobReq GetJobRequest, opts ...RequestOption) (*GetJobResponse, http.Header, error) {
+	path := constructPath("/stacks/%s/jobs/%s", getJobReq.StackKey, getJobReq.JobUUID)
+	req, err := c.newRequest(ctx, http.MethodGet, path, nil, opts...)
+	if err != nil {
+		return nil, nil, fmt.Errorf("failed to create request: %w", err)
+	}
+
+	resp, header, err := do[GetJobResponse](ctx, c, req)
+	if err != nil {
+		return nil, nil, fmt.Errorf("get job: %w", err)
```
DrJosh9000: `do` returns the response headers if they exist whether or not `err` is nil. It's not really idiomatic Go to do that, but since the retry-after header could be important to the caller, we should do the same here:

```suggestion
		return nil, header, fmt.Errorf("get job: %w", err)
```

api/agent_client.go
```diff
@@ -262,6 +266,26 @@ type AgentJob struct {
...
+			ctx,
+			stacksapi.GetJobRequest{
+				StackKey: c.stack.Key,
+				JobUUID:  id,
+			},
+			// The caller handles the retry, we may not want that in the future
+			stacksapi.WithNoRetry(),
+		)
+		if err != nil {
+			return nil, 0, err
```
```suggestion
			return nil, readRetryAfter(header), err
```

PR #716, opened 2025-09-25
PB-629 Reserve jobs using the stacks API

moskyb:
For a little while now, the k8s stack has supported reserving jobs before it spins the infrastructure up to run them. This used an endpoint that's not been migrated into the stacks API, so this PR does a couple of things:
- Folds the `--experimental-reservation-support` flag into the `--experimental-stack-api-support` flag
- Adds a new method to the `stacksapi` package to use the new endpoint
- Updates the controller to use the new endpoints when necessary


internal/controller/reserver/reserver.go
```diff
@@ -53,14 +54,14 @@ func (r *Reserver) HandleMany(ctx context.Context, jobs []*api.AgentScheduledJob
 		return r.handler.HandleMany(ctx, jobs)
 	}
 
-	var job_ids []string
+	var jobIDs []string
```
DrJosh9000: Thank you

internal/stacksapi/scheduled_jobs.go
```diff
@@ -93,3 +93,37 @@ func (c *Client) ListScheduledJobs(ctx context.Context, listReq ListScheduledJob
...
+	NotReserved []string `json:"not_reserved"` // The UUIDs of the jobs that could not be reserved (e.g. because they were already reserved by another stack, or because they're no longer in a reservable state)
+}
+
+// BatchReserveJobs attempts to reserve a list of jobs for processing by the calling stack. The StackKey and JobUUIDs
+// fields of the request are required. If a job cannot be reserved (for example, because it has already been reserved
+// by another stack, or because it is no longer in a reservable state), it will be included in the NotReserved list
+// in the response. Note that even if none of the jobs could be reserved, this method will still return a nil error,
+// as long as the request itself was valid.
+func (c *Client) BatchReserveJobs(ctx context.Context, reserveReq BatchReserveJobsRequest, opts ...RequestOption) (*BatchReserveJobsResponse, http.Header, error) {
+	path := fmt.Sprintf("/stacks/%s/scheduled_jobs/batch_reserve", reserveReq.StackKey)
```
DrJosh9000: If `StackKey` is customer-supplied, then we should be sure to %-escape `.` - see https://github.com/buildkite/agent/blob/4e9dc55fb68e89a0ee9c8994f3afe48d27b7d01d/api/client.go#L418-L421

Anywhere else we're doing this?

moskyb: ooh good call, have fixed in [Escape path params in the stacks API](https://github.com/buildkite/agent-stack-k8s/pull/716/commits/c96693fa6c0dedca8df7607c51cadf6d6347b610)

PR #717, opened 2025-09-25
PB-119: controller create stack notifications

zhming0:
Create a stack notification to when k8s job gets successfully created. 

I had a thought about creating more notifications but as I am making this, I have doubt about what truly useful information is. So to be conservative, I only added one. We can always add more later.

### DIFF
--- a/api/agent_client.go
+++ b/api/agent_client.go
@@ -31,6 +31,7 @@ type AgentClient struct {
 	queue     string
 	stack     *stacksapi.RegisterStackResponse
 
+	logger      *zap.Logger
 	useStackAPI bool
 }
 
@@ -71,6 +72,7 @@ func NewAgentClient(ctx context.Context, opts AgentClientOpts) (*AgentClient, er
 		},
 		clusterID:   opts.ClusterID,
 		queue:       opts.Queue,
+		logger:      opts.Logger,
 		useStackAPI: opts.UseStacksAPI,
 	}
 
@@ -392,6 +394,28 @@ func (c *AgentClient) FailJob(ctx context.Context, jobUUID string, errorDetail s
 
 	_, err := c.stacksAPIClient.FailJob(ctx, req)
 	return err
+}
+
+// CreateStackNotification creates a stack notification for a job.
+// This is designed to be call and forget, error will be logged because we don't anticipate error handling
+func (c *AgentClient) CreateStackNotification(ctx context.Context, jobUUID string, detail string) {
+	if !c.UseStackAPI() {
+		// No-op for the case for legacy API
+		return
+	}
+
+	req := stacksapi.CreateStackNotificationRequest{
+		StackKey: c.stack.Key,
+		JobUUID:  jobUUID,
+		Detail:   detail,
+	}
+
+	// No retry as we want to avoid any possible source of latency bump for job creation flow.
+	// In the future this will be moved into a batched stack notification worker, we can turn back auto retry then
+	_, err := c.stacksAPIClient.CreateStackNotification(ctx, req, stacksapi.WithNoRetry())
+	if err != nil {
+		c.logger.Warn("Failed creating stack notification", zap.Error(err))
+	}
 }
 
 func decodeResponse[T any](resp *http.Response) (result *T, retryAfter time.Duration, err error) {
--- a/internal/stacksapi/stack_notification.go
+++ b/internal/stacksapi/stack_notification.go
@@ -1,0 +1,36 @@
+package stacksapi
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+)
+
+type CreateStackNotificationRequest struct {
+	StackKey string `json:"-"`      // The key to call the stack. Required.
+	JobUUID  string `json:"-"`      // The uuid of a job
+	Detail   string `json:"detail"` // Short notification message (max length 255)
+}
+
+func (c *Client) CreateStackNotification(ctx context.Context, req CreateStackNotificationRequest, opts ...RequestOption) (http.Header, error) {
+	const croppedMessage = "… (cropped)"
+	const maxDetailSize = 255 - len(croppedMessage)
+
+	if len(req.Detail) > maxDetailSize {
+		c.logger.Warn("Stack notification detail exceeds character limit, cropping", "original_size", len(req.Detail))
+		req.Detail = req.Detail[:maxDetailSize] + croppedMessage
+	}
+
+	path := constructPath("/stacks/%s/jobs/%s/stack_notifications", req.StackKey, req.JobUUID)
+	httpReq, err := c.newRequest(ctx, http.MethodPost, path, req, opts...)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create request: %w", err)
+	}
+
+	_, header, err := do[struct{}](ctx, c, httpReq)
+	if err != nil {
+		return nil, fmt.Errorf("create stack notification: %w", err)
+	}
+
+	return header, nil
+}
--- a/internal/stacksapi/stack_notification_test.go
+++ b/internal/stacksapi/stack_notification_test.go
@@ -1,0 +1,84 @@
+package stacksapi
+
+import (
+	"encoding/json"
+	"net/http"
+	"strings"
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestCreateStackNotification(t *testing.T) {
+	t.Parallel()
+
+	t.Run("successful create stack notification", func(t *testing.T) {
+		t.Parallel()
+
+		server, client := setupTestServer(t, func(w http.ResponseWriter, r *http.Request) {
+			verifyAuthMethodPath(t, r, "POST", "/stacks/stack-123/jobs/456/stack_notifications")
+
+			var params CreateStackNotificationRequest
+			assert.NoError(t, json.NewDecoder(r.Body).Decode(&params))
+
+			expectedParams := CreateStackNotificationRequest{
+				Detail: "Pod is starting up",
+			}
+
+			if diff := cmp.Diff(expectedParams, params); diff != "" {
+				t.Errorf("request params mismatch (-want +got):\n%s", diff)
+			}
+
+			w.Header().Set("X-Custom-Header", "custom-value")
+			w.WriteHeader(http.StatusOK)
+		})
+		t.Cleanup(func() { server.Close() })
+
+		req := CreateStackNotificationRequest{
+			StackKey: "stack-123",
+			JobUUID:  "456",
+			Detail:   "Pod is starting up",
+		}
+
+		header, err := client.CreateStackNotification(t.Context(), req)
+		if err != nil {
+			t.Fatalf("client.CreateStackNotification returned an error: %v", err)
+		}
+
+		want, got := "custom-value", header.Get("X-Custom-Header")
+		if want != got {
+			t.Errorf("header.Get(\"X-Custom-Header\") = %q, want %q", got, want)
+		}
+	})
+
+	t.Run("crops detail when exceeds 255 characters", func(t *testing.T) {
+		t.Parallel()
+
+		server, client := setupTestServer(t, func(w http.ResponseWriter, r *http.Request) {
+			verifyAuthMethodPath(t, r, "POST", "/stacks/stack-123/jobs/456/stack_notifications")
+
+			var params CreateStackNotificationRequest
+			assert.NoError(t, json.NewDecoder(r.Body).Decode(&params))
+
+			// Verify the detail was cropped to 255 characters
+			assert.Equal(t, 255, len(params.Detail))
+			assert.True(t, strings.HasSuffix(params.Detail, "… (cropped)"))
+
+			w.WriteHeader(http.StatusOK)
+		})
+		t.Cleanup(func() { server.Close() })
+
+		// Create a detail longer than 255 characters
+		largeDetail := strings.Repeat("x", 300) // 300 characters
+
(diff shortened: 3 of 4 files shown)


api/agent_client.go
```diff
@@ -422,6 +425,28 @@ func (c *AgentClient) FailJob(ctx context.Context, jobUUID string, errorDetail s
 	return err
 }
 
+// CreateStackNotification creates a stack notification for a job.
+// This is designed to be call and forget, error will be logged but we don't anticipate error handling
+func (c *AgentClient) CreateStackNotification(ctx context.Context, jobUUID string, detail string) {
```
moskyb, line 401: even if we anticipate that in most cases it won't be useful to handle errors, we should give users the ability to make that decision by returning an error that they can choose to ignore

api/agent_client.go
```diff
@@ -422,6 +425,28 @@ func (c *AgentClient) FailJob(ctx context.Context, jobUUID string, errorDetail s
...
+	if !c.UseStackAPI() {
+		// No-op for the case for legacy API
+		return
+	}
+
+	go func() {
```
DrJosh9000: Why does this need to be concurrent?

zhming0: Since this is method is a auxiliary thing, I don't want caller to be blocked by this.

internal/stacksapi/stack_notification_test.go
```diff
@@ -0,0 +1,84 @@
...
+		}
+
+		header, err := client.CreateStackNotification(t.Context(), req)
+		if err != nil {
+			t.Fatalf("client.CreateStackNotification returned an error: %v", err)
+		}
+
+		want, got := "custom-value", header.Get("X-Custom-Header")
+		if want != got {
+			t.Errorf("header X-Custom-Header: want %q, got %q", want, got)
```
DrJosh9000: https://google.github.io/styleguide/go/decisions#got-before-want


```suggestion
			t.Errorf("header.Get(\"X-Custom-Header\") = %q, want %q", got, want)
```

internal/stacksapi/stack_notification.go
```diff
@@ -0,0 +1,36 @@
...
+
+type CreateStackNotificationRequest struct {
+	StackKey string `json:"-"`      // The key to call the stack. Required.
+	JobUUID  string `json:"-"`      // The uuid of a job
+	Detail   string `json:"detail"` // Short notification message (max length 255)
+}
+
+func (c *Client) CreateStackNotification(ctx context.Context, req CreateStackNotificationRequest, opts ...RequestOption) (http.Header, error) {
+	const croppedMessage = "… (cropped)"
+	const maxDetailSize = 255 - len(croppedMessage)
```
moskyb, line 17: [non-blocking] this message takes up a fairly significant portion of the space in the notification detail (11/255 ≈ 5%). i wonder if it makes more sense to return an error early if the size exceeds the limit rather than cropping

moskyb (changes requested):

api/agent_client.go
moskyb: oh i missed that the body of this method ran in a goroutine. i think that that too should be left in the hands of the users — they should be able to decide whether or not they await the result of this method call.

api/agent_client.go
DrJosh9000: Why _shouldn't_ the caller be blocked? All that's going to do is slow down processing of jobs. If notifications are taking a long time, then maybe we shouldn't hurry the rest of the loop.

If we do decide to not block the caller for whatever reason, then there should be a fixed size pool of goroutines for sending notifications, not one goroutine per notification, since that risks filling the runtime with goroutines.

api/agent_client.go
DrJosh9000: +1 to this. Introducing asynchronous behaviour in a library makes it harder for the caller to reason about.

Some other words and links here: https://google.github.io/styleguide/go/decisions#goroutine-lifetimes

zhming0: I agree that as a library it should stay simple and offer more control for the caller. 

But this is the `agent_client`, our internal adapter for stack api, which is above our stackapis library.

Does this change your opinions about this topic?

Alternatively, I can have this error suppression mechanism raised to another level, which feels unnecessary to me?

api/agent_client.go
zhming0: There seem to two questions in the comment, why and how. 

About why: 

Since it's hard to anticipate how many Stack notification will be created during the critical path of job provisioning, I think it's prudent to rule out impact to latency. If each stack notification takes 100ms and we have 10 notifications (which is likely) per job, that will be a 1s delay for a job, which has a non-trivial compounding impact. 

---

About how: 

My understand of user-space cooperative green thread is that we wouldn't need to do threading pool like setup.  The RAM cost per goroutine is trivial (2KB), context switching is practically free (unless we already have saturated CPU during peak load). In this PR, we introduces 1 extra goroutine goroutine per job, I don't see a risk here. 

Even if we have 20 concurrent (very unlikely) goroutines per job, in the case of 100K jobs, we have 20KB x 100K = 2GB of RAM. It's not unreasonable. 

Having said that I am open to a goroutine pool setup, it will change the 20KB -> 100 bytes in above equation. Just wondering if RAM was your specific concern here?

---

Overall speaking, I think we have a very likely latency increase vs very unlikely goroutine load issue, so I am slightly in favor of keeping the goroutine setup as it is. Let me know what you think 🙏🏿

DrJosh9000: If you feel strongly about sending the notifications concurrently to the rest of the loop, I'm happy to compromise on the "why".

I'm less compromising on the "how". If there's no need to send more than a few notifications at the same time, then there's no need to create a potentially unbounded number of goroutines to do that. What I mean is, this implementation is not creating 1 extra goroutine per _job_, it's creating 1 extra goroutine per `CreateStackNotification` call. Any bug that causes `CreateStackNotification` to be called many times repeatedly could blow up the runtime.

It may not be a problem for agent-stack-k8s if we only call this once per job, but this will certainly be treated as example code by other implementers, so I would like to do it right.

zhming0: Fair point, I've made a goroutine pool for this. 

Although, as I am making it, I realized that we should probably make that endpoint batched too.

api/agent_client.go
```diff
@@ -69,9 +80,19 @@ func NewAgentClient(ctx context.Context, opts AgentClientOpts) (*AgentClient, er
...
 		},
-		clusterID:   opts.ClusterID,
-		queue:       opts.Queue,
-		useStackAPI: opts.UseStacksAPI,
+		clusterID:        opts.ClusterID,
+		queue:            opts.Queue,
+		logger:           opts.Logger,
+		useStackAPI:      opts.UseStacksAPI,
+		notificationChan: make(chan notificationWork, 10000),
+		poolCtx:          ctx,
```
DrJosh9000: Per the context package docs, don't store the context in the struct. Pass it as an arg (to `notificationWorker`).

api/agent_client.go
```diff
@@ -21,6 +21,12 @@ import (
 // This is a special keyword supported on backend for polling from the current default queue in the cluster.
 const defaultQueueKey = "_default"
 
+type notificationWork struct {
+	ctx     context.Context
```
DrJosh9000: Don't put a context in here either. The worker should have one context, it's fine if that is the one passed in as its argument, not the one present when the notification is made.

api/agent_client.go
```diff
@@ -394,6 +436,31 @@ func (c *AgentClient) FailJob(ctx context.Context, jobUUID string, errorDetail s
...
+		jobUUID: jobUUID,
+		detail:  detail,
+	}
+
+	// Non-blocking send to avoid blocking caller if pool is busy
+	// This is to prevent stack notification increasing job latency.
+	select {
+	case c.notificationChan <- work:
+		// Successfully queued
+	default:
```
DrJosh9000: It'd be better to block until the queue has drained a bit.

api/agent_client.go
```diff
@@ -394,6 +436,31 @@ func (c *AgentClient) FailJob(ctx context.Context, jobUUID string, errorDetail s
...
+	work := notificationWork{
+		ctx:     ctx,
+		jobUUID: jobUUID,
+		detail:  detail,
+	}
+
+	// Non-blocking send to avoid blocking caller if pool is busy
+	// This is to prevent stack notification increasing job latency.
+	select {
+	case c.notificationChan <- work:
```
DrJosh9000: There should be a select case for the context here:
```suggestion
	case <-ctx.Done():
		return
	case c.notificationChan <- work:
```

internal/stacksapi/stack_notification.go
zhming0: IMO 255 is a quite an arbitrary number. Our main objective is to ensure that the message is short, whether the limit is 255 or 244 doesn't matter very much, so I guess it's okay to effectively make the limit 244. 

I am against returning error here because I think it's better to get some signal than no signal. Also the caller side probably can't do much about the error apart from logging it, in such case, we already handle that for the caller on this layer.

api/agent_client.go
zhming0: I am hesitant here as I want to avoid stack notification call blocking job creation. If, for example, stack notification was stuck in a rate limit loop, I would still want the job provisioning to happen successfully. 

I made a change to use a child context to reduce the notification dropping chance. Theoretically it won't make a big difference to raising buffer size apart from making the caller side slow down a bit, but I am not entirely sure if there is a better way?

DrJosh9000: I think it's better to block when the channel is full, to pass back-pressure on to the caller. With the current parameters, there would have to be 10k notifications in the channel (so, a few thousand jobs?). If the channel is being cleared quickly, then the jobs won't have to wait very long for the channel. If the channel is being cleared slower than notifications are coming in, then I think the job rate should slow down, even if it's because of a separate rate limit on notifications, rather than lose notifications.

zhming0: yeah, I agree on this, and I think we are on the same page about the happy path: 

> If the channel is being cleared quickly, then the jobs won't have to wait very long for the channel

But, in the event of severe backend issue, e.g. latency spike, rate limit saturation, when the buffer is full. Do we want to: 
1. Slow down or block the job? 
2. Allow some notification to be dropped? 

My understanding of this controller's top goal was to turn Job UUID -> k8s pod. Maybe it's wise to prioritize that flow so we can fail gracefully? (That was why I swallow error in this PR.)

What do you think?

DrJosh9000: I still think option 1 (block) is the right move. 

* Severe backend issue: other API calls would be more likely to fail. It's not a good time to run jobs. If the service is under heavy load, then clients should back off.
* Rate limit saturation: We can adjust the rate limit. If the notification rate limit turns out to be the bottleneck reducing job throughput, then we've learned something and we can adapt.

zhming0: But if stack can't spin up jobs, it would be P0 incident for the customer. If a stack notification gets missed, it has low impact. 

Looks like there is balance to be made re protecting customer interest vs our infra 🤔 . Just checking that we are leaning towards the latter in this particular decision? 

I am quite sensitive to anything that will block job creation 😬

DrJosh9000: Because of the channel buffer capacity, it will block spinning jobs after 10k notifications are waiting in the channel. That's still a lot of jobs that will be spun up without notifications. If those notifications can't be cleared then someone's already having a bad day. If all those jobs then fail because of backend issues, then we've wasted resources on the customer side, which may itself be a P0 incident for the customer.

zhming0: How about about raising the buffer limit to 100K and let it block? (in total 25MB RAM)

The goal post in mind was about 1K jobs per second, each job has say 2~3 notifications, 10K doesn't give much head room for transient network latency issue in this situation. 100K might be more lenient.

api/agent_client.go
```diff
@@ -21,6 +21,11 @@ import (
 // This is a special keyword supported on backend for polling from the current default queue in the cluster.
 const defaultQueueKey = "_default"
 
+type notificationWork struct {
+	jobUUID string
+	detail  string
```
DrJosh9000: It just occurred to me that if the notifications are sent asynchronously, then they may be totally out of order with other job events. Do we need to include a timestamp with the request?

zhming0: Yeah I noticed that too. I raised an independent internal ticket for this. 

I was also thinking if we should just turn the notification endpoint to batched too.

DrJosh9000: In the absence of changes to the API (timestamping, batching), should we (for the time being) make notifications synchronous?

api/agent_client.go
DrJosh9000: If you add that as a comment near the `make(chan ..)` (explaining where 100K comes from) then sounds good to me!

api/agent_client.go
zhming0: yeah timestamping is a deal break. I changed it to be synchronous.

api/agent_client.go
zhming0: Per our chat below, I've changed it to synchronous for the timebeing.

internal/stacksapi/stack_notification.go
```diff
@@ -0,0 +1,36 @@
...
+
+	if len(req.Detail) > maxDetailSize {
+		c.logger.Warn("Stack notification detail exceeds character limit, cropping", "original_size", len(req.Detail))
+		req.Detail = req.Detail[:maxDetailSize] + croppedMessage
+	}
+
+	path := fmt.Sprintf("/stacks/%s/jobs/%s/stack_notifications", req.StackKey, req.JobUUID)
```
DrJosh9000:
```suggestion
	path := constructPath("/stacks/%s/jobs/%s/stack_notifications", req.StackKey, req.JobUUID)
```

PR #718, opened 2025-09-25
Add `buildkite.com/build-repo` and `buildkite.com/pipeline-slug` annotations

adnowakodbx1:
In previous versions on k8s stack there was `BUILDKITE_REPO` env var on the containers in the pod that we used in k8s admission hook to mount additional volume containing repository to the pod to speed up clones. In latest k8s stack this is no longer exposed and would be great to have an alternative. We are also using `BUILDKITE_PIPELINE_SLUG` to emit some metrics from k8s - this variable is also no longer visible to k8s. 

There seems to be precedent for exposing some of repository and build information through annotations, for example there is `buildkite.com/build-branch` and `buildkite.com/build-url`. So I'm proposing to add 2 new annotations - `buildkite.com/build-repo` and `buildkite.com/pipeline-slug` that would contain above mentioned values.

### DIFF
--- a/internal/controller/config/config.go
+++ b/internal/controller/config/config.go
@@ -15,8 +15,10 @@ const (
 	ControllerIDLabel                     = "buildkite.com/controller-id"
 	BuildURLAnnotation                    = "buildkite.com/build-url"
 	BuildBranchAnnotation                 = "buildkite.com/build-branch"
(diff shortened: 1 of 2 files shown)


nsuma8989 (approved):
Changes look good to me.

PR #719, opened 2025-09-29
Expose termination grace period as environment variable

scadu:
Required for 
https://github.com/buildkite/agent/pull/3500#discussion_r2386951125

### DIFF
--- a/internal/controller/scheduler/scheduler.go
+++ b/internal/controller/scheduler/scheduler.go
@@ -607,6 +607,20 @@ func (w *worker) Build(podSpec *corev1.PodSpec, skipCheckout bool, inputs buildI
 		},
 	}
 
+	// Set log collection grace period based on termination grace period.
+	// This allows the agent to coordinate log collection timing with pod termination.
+	if podSpec.TerminationGracePeriodSeconds != nil {
+		// Calculate log collection grace period: termination grace period minus buffer for cleanup
+		logCollectionGracePeriod := *podSpec.TerminationGracePeriodSeconds - 10
+		if logCollectionGracePeriod < 1 {
+			logCollectionGracePeriod = 1 // Minimum 1 second
+		}
+		agentContainer.Env = append(agentContainer.Env, corev1.EnvVar{
+			Name:  "BUILDKITE_KUBERNETES_LOG_COLLECTION_GRACE_PERIOD",
+			Value: fmt.Sprintf("%ds", logCollectionGracePeriod),
+		})
+	}
+
 	// Append some agent config and checkout config to the agent container.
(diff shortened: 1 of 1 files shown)


internal/controller/scheduler/scheduler.go
```diff
@@ -705,6 +705,27 @@ func (w *worker) Build(podSpec *corev1.PodSpec, skipCheckout bool, inputs buildI
 		w.logger.Debug("Applied podSpec patch from k8s plugin", zap.Any("patched", patched))
 	}
 
+	// After all patching is complete, pass the actual termination grace period to the agent
+	// so it can coordinate log collection timing appropriately.
+	if podSpec.TerminationGracePeriodSeconds != nil {
```
DrJosh9000: I'm not sure why this needs to happen after patching. Could it be set when the agent container is configured?

scadu: @DrJosh9000 addressed it now. I initially thought to do it after patching if the configuration was fed via `podSpec` patch, but I don't think this is going to be the case (in most circumstances)?

DrJosh9000 (approved):
Thanks!

internal/controller/scheduler/scheduler.go
DrJosh9000: Hmm, maybe. If someone patches `terminationGracePeriodSeconds`, then I think they should also have to patch the environment variable. (More generally, that a patch should change exactly what is in the patch.) Maybe we should make the default configurable?

scadu: @DrJosh9000 do you suggest to decouple `BUILDKITE_KUBERNETES_LOG_COLLECTION_GRACE_PERIOD` from `podSpec.TerminationGracePeriodSeconds` and set `BUILDKITE_KUBERNETES_LOG_COLLECTION_GRACE_PERIOD` with default of `50s`?

DrJosh9000: Currently we have this line:

```go
	kjob.Spec.Template.Spec.TerminationGracePeriodSeconds = ptr.To[int64](defaultTermGracePeriodSeconds)
```

and as far as I can tell, that is the only time that `TerminationGracePeriodSeconds` is set (by us). `defaultTermGracePeriodSeconds` is a constant. I'm suggesting (this can be in the future, doesn't have to be done in this PR) that perhaps we should add a `TerminationGracePeriodSeconds` field to `Config` that can be controlled via the stack config (Helm values, etc).

scadu: @DrJosh9000 ah, cool. Will raise a separate PR with this.

PR #721, opened 2025-10-01
Set TerminationGracePeriodSeconds correctly, add config

DrJosh9000:
### What

The important bit is this:

```diff
-	kjob.Spec.Template.Spec.TerminationGracePeriodSeconds = ptr.To[int64](defaultTermGracePeriodSeconds)
+	if podSpec.TerminationGracePeriodSeconds == nil {
+		podSpec.TerminationGracePeriodSeconds = ptr.To(int64(w.cfg.DefaultTerminationGracePeriodSeconds))
+	}
```

The rest is 
- making it configurable
- some cleanups, particularly, removing the last vestiges of org and cluster-uuid.

### Why

`kjob.Spec.Template.Spec` is set to `podSpec` near the end of `Build`:

https://github.com/buildkite/agent-stack-k8s/blob/3e3fc98b3b26e0a9db401d8b9101055db20cbb7a/internal/controller/scheduler/scheduler.go#L727

So whatever changes made to `kjob.Spec.Template.Spec` directly will be overridden. This appears to be the only instance of a direct change to `kjob.Spec.Template.Spec`.


cmd/controller/controller.go
```diff
@@ -79,16 +79,16 @@ func AddConfigFlags(cmd *cobra.Command) {
...
 	cmd.Flags().Duration(
 		"poll-interval",
 		time.Second,
 		"time to wait between polling for new jobs (minimum 1s); note that increasing this causes jobs to be slower to start",
 	)
-	cmd.Flags().String(
-		"cluster-uuid",
-		"",
-		"UUID of the Buildkite Cluster. The agent token must be for the Buildkite Cluster.",
-	)
```
zhming0, line 91: Not disagreeing, but this would make this PR a breaking change? People who still passes cluster-uuid to the stack, despite it's unused, will hit error after this release.

zhming0 (approved):
LGTM with a small callout that this is potentially a light breaking change.

cmd/controller/controller.go
DrJosh9000: That's true, but it is easy for a customer to fix (just remove the flag). The `org` flag was removed in June.

PR #722, opened 2025-10-06
Add podTemplate to plugin config

DrJosh9000:
### What

Add a `podTemplate` field to the plugin config. If a `podTemplate` is provided (but not a `podSpec`), fetch it from Kubernetes and use it as the initial pod spec.

### Why

https://linear.app/buildkite/issue/PS-1233

Seems like a straightforward way to reduce in-band configuration using the built-in Kubernetes `PodTemplate` resource kind.

### TODO

- [x] Add a test

### DIFF
--- a/.buildkite/fixtures/podtemplates.yaml
+++ b/.buildkite/fixtures/podtemplates.yaml
@@ -1,0 +1,14 @@
+# This PodTemplate is used by TestPodTemplate.
+apiVersion: v1
+kind: PodTemplate
+metadata:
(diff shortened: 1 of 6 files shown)


DrJosh9000:
`podTemplate` provides the initial pod spec, similar to `podSpec`. Then the resource class gets applied to that (in the middle of `Build`). Where this could go wrong is if the template used a name for the command container other than `container-0`, but the template could just as easily set resource requests and limits directly.

PR #723, opened 2025-10-06
PB-734: use finish job endpoint instead of fail job endpoint

zhming0:
part of PB-734

This PR makes sure k8s stack use our new Finish Job endpoint instead of the old Fail Job endpoint, which no longer exist.


DrJosh9000 (approved):
Looks like a straighforward rename to me 👍 ✅

PR #731, opened 2025-10-08
PB-732: use new batched notification API

zhming0:
Accommodate our breaking changes in backend for the stack notification API. 

In this PR, the agent_client level abstraction remains unchanged. I make this so we can quickly cut over to the new API design and get our CI working unblock.

I will very quickly make a follow up PR that actually batch notification on agent_client side. But the changes in stack_api layer will stay.

### DIFF
--- a/internal/stacksapi/stack_notification.go
+++ b/internal/stacksapi/stack_notification.go
@@ -4,33 +4,40 @@ import (
 	"context"
 	"fmt"
 	"net/http"
+	"time"
 )
 
-type CreateStackNotificationRequest struct {
-	StackKey string `json:"-"`      // The key to call the stack. Required.
-	JobUUID  string `json:"-"`      // The uuid of a job
-	Detail   string `json:"detail"` // Short notification message (max length 255)
+type StackNotification struct {
+	JobUUID   string    `json:"job_uuid"`
+	Detail    string    `json:"detail"`
+	Timestamp time.Time `json:"timestamp,omitzero"`
 }
 
-func (c *Client) CreateStackNotification(ctx context.Context, req CreateStackNotificationRequest, opts ...RequestOption) (http.Header, error) {
-	const croppedMessage = "… (cropped)"
-	const maxDetailSize = 255 - len(croppedMessage)
+type CreateStackNotificationsRequest struct {
+	StackKey      string              `json:"-"`
+	Notifications []StackNotification `json:"notifications"`
+}
 
-	if len(req.Detail) > maxDetailSize {
-		c.logger.Warn("Stack notification detail exceeds character limit, cropping", "original_size", len(req.Detail))
-		req.Detail = req.Detail[:maxDetailSize] + croppedMessage
-	}
+type StackNotificationError struct {
+	Error   string `json:"error"`
+	Indexes []int  `json:"indexes"`
+}
 
-	path := constructPath("/stacks/%s/jobs/%s/stack_notifications", req.StackKey, req.JobUUID)
+type CreateStackNotificationsResponse struct {
+	Errors []StackNotificationError `json:"errors"`
+}
+
+func (c *Client) CreateStackNotifications(ctx context.Context, req CreateStackNotificationsRequest, opts ...RequestOption) (*CreateStackNotificationsResponse, http.Header, error) {
+	path := constructPath("/stacks/%s/notifications", req.StackKey)
 	httpReq, err := c.newRequest(ctx, http.MethodPost, path, req, opts...)
 	if err != nil {
-		return nil, fmt.Errorf("failed to create request: %w", err)
+		return nil, nil, fmt.Errorf("failed to create request: %w", err)
 	}
 
-	_, header, err := do[struct{}](ctx, c, httpReq)
+	resp, header, err := do[CreateStackNotificationsResponse](ctx, c, httpReq)
 	if err != nil {
-		return nil, fmt.Errorf("create stack notification: %w", err)
+		return nil, header, fmt.Errorf("create stack notifications: %w", err)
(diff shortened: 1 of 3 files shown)


internal/stacksapi/stack_notification.go
```diff
@@ -4,33 +4,40 @@ import (
...
 )
 
-type CreateStackNotificationRequest struct {
-	StackKey string `json:"-"`      // The key to call the stack. Required.
-	JobUUID  string `json:"-"`      // The uuid of a job
-	Detail   string `json:"detail"` // Short notification message (max length 255)
+type StackNotification struct {
+	JobUUID   string     `json:"job_uuid"`
+	Detail    string     `json:"detail"`
+	Timestamp *time.Time `json:"timestamp,omitempty"`
```
DrJosh9000: If you use `omitzero` instead of `omitempty` and remove the `*`, then `encoding/json` will call the `IsZero` method of `time.Time` to figure out whether or not it is set.

internal/stacksapi/stack_notification_test.go
```diff
@@ -32,53 +40,77 @@ func TestCreateStackNotification(t *testing.T) {
 
 			w.Header().Set("X-Custom-Header", "custom-value")
 			w.WriteHeader(http.StatusOK)
+			assert.NoError(t, json.NewEncoder(w).Encode(CreateStackNotificationsResponse{
+				Errors: []StackNotificationError{},
+			}))
```
DrJosh9000, line 49: Going forward, I would greatly appreciate it if test errors identified the function called and any relevant args, if not too long. `assert.NoError` doesn't log this information on its own.

```suggestion
			err := json.NewEncoder(w).Encode(CreateStackNotificationsResponse{
				Errors: []StackNotificationError{},
			})
			if err != nil {
				t.Errorf("json.NewEncoder(w).Encode(CreateStackNotificationsResponse{...}) error = %v", err)
			}
```

zhming0: Are you saying that we shouldn't use assert.NoError? I think there is a trade-off to be made here. 

Why I think assert.NoError is good: 

1. Many errors in test are so rare that having that log line is more harm than good, as it reduce code readibiliy. 
2. Even with the log line, in most case I believe people will need go back to the test file to understand the context. 

So I think there is an imbalance on the benefit vs cost here. Do you feel strongly about it? Maybe we can talk in sync, if we can get a consensus then maybe we can make a linter rule?

DrJosh9000: If it were up to me, `assert` would be banned from existence. But I realise that is an extreme opinion. 

A good error log message practically eliminates the need to re-read the test in order to understand the failure. Indenting the error flow makes it easier to "skip over" when reading. I think `assert.NoError` is _harder_ to read, because it doesn't look like the common error handling pattern in Go (`if err != nil {...}`).

zhming0: Maybe the issue here is that I use `assert.NoError` on the same line with the calling code? 

If I split them, would this be more readable? It resembles the canonical error flow. There are a few cases in codebase, which we haven't received readability complaint in previous reviews. 

```golang
err := json.NewEncoder(w).Encode(...)
assert.NoError(t, err) 
```

----

On the other hand, if we want to ban it, I will be in favor of an automatic approach, but be against manual enforcement.

internal/stacksapi/stack_notification.go
zhming0: Good catch!

internal/stacksapi/stack_notification_test.go
DrJosh9000: It's not the canonical error flow, because the error handling is not indented.

If there are a few cases already in the codebase that I haven't complained about before, it's because there's a tradeoff with shipping code quickly and maintaining a pre-existing style. But I felt it was finally time to say something.

zhming0: Would you be open for us to address this outside of this PR? 

So we can one get everyone to agree (which I am very open to), two find a scalable solution. 

I don't have strong opinion on work related code style, but I feel strongly about automatic enforcement.

DrJosh9000: Sure, no problem.

internal/stacksapi/stack_notification_test.go
```diff
@@ -32,53 +44,78 @@ func TestCreateStackNotification(t *testing.T) {
...
 		}
 
-		_, err := client.CreateStackNotification(t.Context(), req)
+		resp, _, err := client.CreateStackNotifications(t.Context(), req)
 		assert.NoError(t, err)
+		assert.Len(t, resp.Errors, 2)
+		assert.Equal(t, "detail is required", resp.Errors[0].Error)
+		assert.Equal(t, []int{1}, resp.Errors[0].Indexes)
+		assert.Equal(t, "detail exceeds its length limit of 256", resp.Errors[1].Error)
+		assert.Equal(t, []int{3}, resp.Errors[1].Indexes)
```
DrJosh9000: This would be nicer with `cmp.Diff` (https://google.github.io/styleguide/go/decisions#full-structure-comparisons):

```suggestion
		wantErrors := []StackNotificationError{
			{
				Error:   "detail is required",
				Indexes: []int{1},
			},
			{
				Error:   "detail exceeds its length limit of 256",
				Indexes: []int{3},
			},
		}
		if diff := cmp.Diff(resp.Errors, wantErrors); diff != "" {
			t.Errorf("resp.Errors diff (-got +want):\n%s", diff)
		}
```

DrJosh9000 (approved):

PR #732, opened 2025-10-08
PS-1250 Fix bug where resource classes with hugepages wouldn't work

moskyb:
Viper, our config parsing library, kindly downcases all keys of maps in config. Unfortunately, the resource-classes config item has maps that have case-sensitive keys, but only in one case. One of the available resource requests is `hugepages-2Mi` (where both the number and the unit are variable), and when this was getting parsed by viper, it was downcasing it to `hugepages-2mi` (lowercase m). This would cause k8s to not recognise the resource claim.

This PR fixes this case by going in and doing surgery on hugepages resource claims, ensuring that the unit is uppercased.

Ticket: PS-1250

### DIFF
--- a/cmd/controller/controller.go
+++ b/cmd/controller/controller.go
@@ -6,6 +6,7 @@ import (
 	"log"
 	"os"
 	"reflect"
+	"regexp"
 	"strconv"
 	"strings"
 
@@ -24,6 +25,7 @@ import (
 	"github.com/spf13/pflag"
 	"github.com/spf13/viper"
 	"go.uber.org/zap"
+	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 	"k8s.io/apimachinery/pkg/util/intstr"
 	"k8s.io/client-go/kubernetes"
@@ -161,6 +163,13 @@ func ParseAndValidateConfig(v *viper.Viper) (*config.Config, error) {
 		}
 	}
 
+	for name, rc := range cfg.ResourceClasses {
+		err := recapitalizeHugePagesResourceClasses(rc)
+		if err != nil {
+			return nil, fmt.Errorf("invalid resource class %q: %w", name, err)
+		}
+	}
+
 	if _, err := resource.ParseQuantity(cfg.ImageCheckContainerCPULimit); err != nil {
 		return nil, fmt.Errorf("invalid CPU resource limit defined: %s", cfg.ImageCheckContainerCPULimit)
 	}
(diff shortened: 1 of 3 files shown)


cmd/controller/controller.go
```diff
@@ -337,6 +348,58 @@ func ParseAndValidateConfig(v *viper.Viper) (*config.Config, error) {
...
+
+	rc.Resource.Requests, err = recapitalizeHugePagesResourceMap(rc.Resource.Requests)
+	if err != nil {
+		return fmt.Errorf("invalid resource class requests: %w", err)
+	}
+
+	return nil
+}
+
+var hugepagesSizeRE = regexp.MustCompile(`^(\d+)([KMGPEkmgpe])i$`)
```
DrJosh9000, line 202: Oh, to have hugepages exabytes in size!

cmd/controller/controller.go
```diff
@@ -326,6 +328,15 @@ func ParseAndValidateConfig(v *viper.Viper) (*config.Config, error) {
 		}
 	}
 
+	if cfg.ResourceClasses != nil {
```
DrJosh9000: Ranging over a nil map is no problem.

cmd/controller/controller.go
```diff
@@ -337,6 +348,58 @@ func ParseAndValidateConfig(v *viper.Viper) (*config.Config, error) {
...
+		if !strings.HasPrefix(stringRes, corev1.ResourceHugePagesPrefix) {
+			newResourceList[res] = qty
+			continue
+		}
+
+		hugepagesSize := strings.TrimPrefix(stringRes, corev1.ResourceHugePagesPrefix)
+
+		matched := hugepagesSizeRE.FindStringSubmatch(hugepagesSize)
+		if len(matched) != 3 {
+			return nil, fmt.Errorf("couldn't match hugepages size %s to regex %s. This is a bug, please contact support@buildkite.com", hugepagesSize, hugepagesSizeRE.String())
```
DrJosh9000: - `%q` in case `hugepagesSize` is empty
- `fmt` functions know how to call `String`

```suggestion
			return nil, fmt.Errorf("couldn't match hugepages size %q to regex %s. This is a bug, please contact support@buildkite.com", hugepagesSize, hugepagesSizeRE)
```

cmd/controller/controller_test.go
```diff
@@ -18,160 +18,175 @@ func ptr[T any](v T) *T {
...
-		PaginationPageSize:                   1000,
-		PaginationDepthLimit:                 5,
-		QueryResetInterval:                   10 * time.Second,
-		DefaultImagePullPolicy:               "Never",
-		DefaultImageCheckPullPolicy:          "IfNotPresent",
-		EnableQueuePause:                     true,
-		WorkQueueLimit:                       2_000_000,
-		ImageCheckContainerCPULimit:          "201m",
-		ImageCheckContainerMemoryLimit:       "129Mi",
+	t.Run("worked example config", func(t *testing.T) {
```
DrJosh9000: I'm not a huge fan of adding a `t.Run` just to wrap a single test case.

cmd/controller/controller.go
moskyb: we're moore's law ready!

cmd/controller/controller_test.go
moskyb: oh, yep, this was a relic from when i was trying to add it in its own test case, but i realised that it's easier to cram everything into one file 🙃 

will remove the extraneous case

cmd/controller/controller.go
```diff
@@ -337,6 +348,58 @@ func ParseAndValidateConfig(v *viper.Viper) (*config.Config, error) {
 	return cfg, nil
 }
 
+// Viper downcases keys in maps used in inputs, but in the case of resource classes where hugepages are used, it
+// gets downcased to "hugepages-2mi". This function fixes that by recapitlizing the "Mi" part for any hugepages resource
+// claims. This is important because the hugepages resource name must be capitalized as "hugepages-2Mi" (or similar),
+// otherwise k8s won't recognize the resource claim.
+func recapitalizeHugePagesResourceClasses(rc *config.ResourceClass) error {
```
zhming0, line 188: Is this possible to move these to another file? `recapticalize_resource.go?` `controller.go` is getting big. And this logic is quite isolated and narrow.

moskyb: i solved this by moving the flags to their own file, as "what flags we have" and "how we process their values" feel like two separate things to me

PR #733, opened 2025-10-08
PB-732 final: Stack notification batcher

zhming0:
Based on too of #731.

This is the final piece of PB-732. It changed agent_client so it sends Stack notifications in batch.

Every 100m, the batcher will send stack notifications to backend, 1000 notifications per call in parallel.

There might be a small risk of notification buffer overflow but I think the likelihood is negligible.

### DIFF
--- a/api/agent_client.go
+++ b/api/agent_client.go
@@ -33,6 +33,9 @@ type AgentClient struct {
 
 	logger      *zap.Logger
 	useStackAPI bool
+
+	notificationBatcher         *notificationBatcher
+	notificationBatcherCancelFn context.CancelFunc
 }
 
 type AgentClientOpts struct {
@@ -104,9 +107,30 @@ func NewAgentClient(ctx context.Context, opts AgentClientOpts) (*AgentClient, er
 		}
 
 		client.stack = stack
+		client.notificationBatcher = newNotificationBatcher(stackKey, client.stacksAPIClient, opts.Logger)
 	}
 
 	return client, nil
+}
+
+// Starting internal goroutines for AgentClient.
+func (e *AgentClient) Start(ctx context.Context) error {
+	if e.UseStackAPI() {
+		ctxWithCancel, cancel := context.WithCancel(ctx)
+		e.notificationBatcherCancelFn = cancel
+		if err := e.notificationBatcher.start(ctxWithCancel); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// Ends all internal goroutines and wait for them to finish pending works.
+func (e *AgentClient) Stop() {
+	if e.UseStackAPI() {
+		e.notificationBatcherCancelFn()
+		e.notificationBatcher.waitDone()
+	}
 }
 
 // AgentError is the JSON object of the response body returned when the HTTP
@@ -421,11 +445,9 @@ func (c *AgentClient) FailJob(ctx context.Context, jobUUID string, errorDetail s
 	return err
 }
 
-// CreateStackNotification creates a stack notification for a job.
-// This is designed to be call and forget, error will be logged because we don't anticipate error handling
+// CreateStackNotification queues a stack notification for a job to be sent in a batch.
 func (c *AgentClient) CreateStackNotification(ctx context.Context, jobUUID string, detail string) {
-	if !c.UseStackAPI() {
-		// No-op for the case for legacy API
+	if !c.UseStackAPI() || c.notificationBatcher == nil {
 		return
 	}
 
@@ -437,25 +459,14 @@ func (c *AgentClient) CreateStackNotification(ctx context.Context, jobUUID strin
 		detail = detail[:maxDetailSize] + croppedMessage
 	}
 
-	req := stacksapi.CreateStackNotificationsRequest{
-		StackKey: c.stack.Key,
-		Notifications: []stacksapi.StackNotification{
-			{
-				JobUUID: jobUUID,
-				Detail:  detail,
-			},
-		},
-	}
-
-	// No retry as we want to avoid any possible source of latency bump for job creation flow.
-	// In the future this will be moved into a batched stack notification worker, we can turn back auto retry then
-	resp, _, err := c.stacksAPIClient.CreateStackNotifications(ctx, req, stacksapi.WithNoRetry())
+	err := c.notificationBatcher.add(ctx, stacksapi.StackNotification{
+		JobUUID: jobUUID,
+		Detail:  detail,
+		// These notifications may be ingested out of order
+		Timestamp: time.Now(),
+	})
 	if err != nil {
-		c.logger.Warn("Failed creating stack notification", zap.Error(err))
-	}
-
-	if len(resp.Errors) > 0 {
-		c.logger.Warn("Failed creating stack notification", zap.String("error", resp.Errors[0].Error))
+		c.logger.Error("Abort sending Stack notification", zap.Error(err))
 	}
 }
 
--- a/api/stack_notification_batcher.go
+++ b/api/stack_notification_batcher.go
@@ -1,0 +1,142 @@
+package api
+
+import (
+	"context"
+	"errors"
+	"slices"
+	"sync"
+	"time"
+
+	"github.com/buildkite/agent-stack-k8s/v2/internal/stacksapi"
+	"go.uber.org/zap"
+)
+
+const (
+	// Extreme situation, we don't anticipate people to encounter.
+	// This will consume ~1GB memory
+	// (The number is from: 1000 jobs per second, 100 notifications per jobs, 30 seconds)
+	maxBufferSize = 3_000_000
+	// Combining maxNotificationsPerInterval and maxNotificationsPerBatch
+	// We limited the max concurrent requests to our backend to: maxNotificationsPerInterval / maxNotificationsPerBatch = 100.
+	// It also limits the max throughput to maxNotificationsPerInterval / notificationBatchInterval.
+	maxNotificationsPerBatch    = 1000
+	maxNotificationsPerInterval = 100_000
+	notificationBatchInterval   = 100 * time.Millisecond
+)
+
+var ErrAlreadyRunning = errors.New("notification batcher is already running")
+
+type notificationBatcher struct {
+	mu            sync.Mutex
+	notifications chan stacksapi.StackNotification
+	stackKey      string
+	client        *stacksapi.Client
+	logger        *zap.Logger
+	done          chan struct{}
+	running       bool
+}
+
+func newNotificationBatcher(stackKey string, client *stacksapi.Client, logger *zap.Logger) *notificationBatcher {
+	nb := &notificationBatcher{
+		notifications: make(chan stacksapi.StackNotification, maxBufferSize),
+		stackKey:      stackKey,
+		client:        client,
+		logger:        logger,
+	}
+	return nb
+}
+
+func (nb *notificationBatcher) add(ctx context.Context, notification stacksapi.StackNotification) error {
+	select {
+	case nb.notifications <- notification:
+		return nil
+	case <-ctx.Done():
+		return ctx.Err()
+	}
+}
+
+func (nb *notificationBatcher) start(ctx context.Context) error {
+	nb.mu.Lock()
+	if nb.running {
+		nb.mu.Unlock()
+		return ErrAlreadyRunning
+	}
+	nb.running = true
+	nb.done = make(chan struct{})
+	nb.mu.Unlock()
+
+	go func() {
+		defer func() {
+			nb.mu.Lock()
+			nb.running = false
+			nb.mu.Unlock()
+			close(nb.done)
+		}()
+
+		ticker := time.NewTicker(notificationBatchInterval)
+		defer ticker.Stop()
+
+		for {
+			select {
+			case <-ctx.Done():
+				// We could do a final flush here.
+				// But it still would be an incomplete graceful shutdown without connection tracking.
+				// So we decided not to do it.
+				return
+			case <-ticker.C:
+				nb.flush(ctx)
+			}
+		}
+	}()
+
+	return nil
+}
+
+func (nb *notificationBatcher) flush(ctx context.Context) {
+	if len(nb.notifications) == 0 {
+		return
+	}
+
+	count := min(len(nb.notifications), maxNotificationsPerInterval)
+	notifications := make([]stacksapi.StackNotification, count)
+	for i := range count {
+		notifications[i] = <-nb.notifications
+	}
+
+	var wg sync.WaitGroup
+	for batch := range slices.Chunk(notifications, maxNotificationsPerBatch) {
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			nb.sendBatch(ctx, batch)
+		}()
+	}
+	wg.Wait()
+}
+
+func (nb *notificationBatcher) sendBatch(ctx context.Context, notifications []stacksapi.StackNotification) {
+	req := stacksapi.CreateStackNotificationsRequest{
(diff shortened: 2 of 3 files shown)


api/stack_notification_batcher.go
```diff
@@ -0,0 +1,115 @@
...
+		stackKey:      stackKey,
+		client:        client,
+		logger:        logger,
+		ctx:           batcherCtx,
+		cancel:        cancel,
+		done:          make(chan struct{}),
+	}
+	go nb.run()
+	return nb
+}
```
moskyb, line 47: contexts [shouldn't be stored in structs](https://go.dev/blog/context-and-structs), and the fact that this one is present and necessary is a sign that there's a bit of a leaky abstraction here. 

with the addition of the notification batcher as something that's directly "owned" and monitored by the `AgentClient`, the agent client has moved from being (more or less) a service object (a struct to hold some state and some methods you can call on it) to being a stateful goroutine monitor that has to look after its children (admittedly, this rule was probably first broken by me in adding the `RegisterStack` call to `NewAgentClient`).

my recommendation here is:
- update the notification batcher so that it doesn't run on the `new` call, and update the `run()` func to take a context
- enforce that callers of `NewAgentClient` pass in a started notification batcher
- then delegate the handling of closing etc the notification batcher to the caller, who has a better idea of the lifecycle of all of the different moving pieces

zhming0: I agree that we shouldn't store context in structs. I fixed that.

More importantly, I also agree that we shouldn't mingle object creation with object's runtime start/shutdown. However, the distinction really starts manifest when have more elaborate lifecycle control... which we don't 😬 . 

In this particular case, the new batcher depends on Stack registration to work, so without more refactor, its lifecycle is bound with `agent_client`. 

I don't think there is a harm to make `AgentClient` stateful because it rightfully is. But it's the "how" that's a bit questionable, fixing that is probably out of scope. 

Anyway, I tried as I might to make it look "correct" and more "future-proof". Please take another look 🙏🏿

api/agent_client.go
```diff
@@ -104,11 +107,32 @@ func NewAgentClient(ctx context.Context, opts AgentClientOpts) (*AgentClient, er
...
+		e.notificationBatcherCancelFn = cancel
+		if err := e.notificationBatcher.start(ctxWithCancel); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// Ends all internal goroutines and wait for them to finish pending works.
+func (e *AgentClient) Stop(ctx context.Context) {
```
DrJosh9000: If the arg is unused, and it's not needed to satisfy an interface, remove it:

```suggestion
func (e *AgentClient) Stop() {
```

If it _is_ needed to satisfy an interface, remove its name:

```suggestion
func (e *AgentClient) Stop(context.Context) {
```

internal/controller/controller.go
```diff
@@ -132,6 +132,11 @@ func Run(
 		return
 	}
 
+	if err = agentClient.Start(ctx); err != nil {
```
DrJosh9000: [An incredibly trivial thing] I prefer `:=` instead of `=` in this situation, since it makes it slightly clearer that the `err` inside the `if` has no meaningful relation to any previously-declared `err`: 

```suggestion
	if err := agentClient.Start(ctx); err != nil {
```

api/stack_notification_batcher.go
```diff
@@ -0,0 +1,132 @@
...
+		client:        client,
+		logger:        logger,
+	}
+	return nb
+}
+
+func (nb *notificationBatcher) add(notification stacksapi.StackNotification) {
+	nb.mu.Lock()
+	defer nb.mu.Unlock()
+	nb.notifications = append(nb.notifications, notification)
```
DrJosh9000: Am I right in understanding that the number of pending notifications is unbounded?

zhming0: Yeah, I briefly mentioned it in PR description. Since the backlog will be continuously dropped by the batcher, I believe the likelihood of this buffer RAM usage to be an issue is negligible. It would need a instant huge influx of notifications.

api/stack_notification_batcher.go
```diff
@@ -0,0 +1,132 @@
...
+	nb.notifications = nb.notifications[:0]
+	nb.mu.Unlock() // Unlock as soon as we pull notifications from buffer
+
+	var wg sync.WaitGroup
+	for batch := range slices.Chunk(notifications, maxNotificationsPerBatch) {
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			nb.sendBatch(ctx, batch)
+		}()
```
DrJosh9000, line 112: This also creates a (hypothetically) unbounded number of goroutines to upload chunks concurrently. I agree it's much less of a problem than goroutine-per-notification, but why not just block the goroutine created in `start`?

api/stack_notification_batcher.go
DrJosh9000: Or a steady influx of notifications while the connection times out. But sure, it's probably not a big problem.

api/stack_notification_batcher.go
zhming0: It is blocking the main batcher goroutine with `wg` currently. 

If I don't do so, then each goroutine is subject to roughly 10K notification per second limit (assuming round trip time of 100ms). So I will need to carefully tune a static pool of batcher goroutine to ensure the number is adequate to handle typical job throughput. 

By letting it self adapt (number of goroutine = number of batches), it adds little risk but a lot easier to maintain.

api/stack_notification_batcher.go
zhming0: I figure adding a limit is small effort but could prevent major problem. I added a very naive throttling mechanism. Another look please 🙏🏿

api/stack_notification_batcher.go
```diff
@@ -0,0 +1,142 @@
...
+	"time"
+
+	"github.com/buildkite/agent-stack-k8s/v2/internal/stacksapi"
+	"go.uber.org/zap"
+)
+
+const (
+	// Extreme situation, we don't anticipate people to encounter.
+	// This will consume ~1GB memory
+	maxBufferSize             = 1000 * 100 * 30
```
DrJosh9000: I'd probably write this as:

```suggestion
	maxBufferSize             = 3_000_000
```

zhming0: Good point, 1000 * 100 * 30 was my mental model of 1000 jobs per second, 100 notifications per jobs, 30 seconds. 

I turned that into comments.

api/stack_notification_batcher.go
DrJosh9000: Well, the sending side of a bounded-size queue would be more efficiently implemented with a channel, since that wouldn't require a sleep in a loop (with locking). A send could proceed as soon as the runtime finds a matching receive. Of course then you can't grab chunks like you can with a slice, but I notice you're copying the slice anyway.

zhming0: No problem. I was chasing minimum code change as I think the situation is so rare but channel is trivial to write too 👍🏿 . 

Another look please?

api/stack_notification_batcher.go
```diff
@@ -0,0 +1,133 @@
...
+		stackKey:      stackKey,
+		client:        client,
+		logger:        logger,
+	}
+	return nb
+}
+
+func (nb *notificationBatcher) add(notification stacksapi.StackNotification) {
+	nb.notifications <- notification
+}
```
DrJosh9000: Now that this blocks in a selectable way, the method could take a context to abandon the send early.

```suggestion
func (nb *notificationBatcher) add(ctx context.Context, notification stacksapi.StackNotification) error {
	select {
	case nb.notifications <- notification:
		return nil
	case <-ctx.Done():
		return ctx.Err()
	}
}
```

api/stack_notification_batcher.go
```diff
@@ -0,0 +1,133 @@
...
+
+func (nb *notificationBatcher) flush(ctx context.Context) {
+	if len(nb.notifications) == 0 {
+		return
+	}
+
+	count := len(nb.notifications)
+	notifications := make([]stacksapi.StackNotification, count)
+	for i := 0; i < count; i++ {
```
DrJosh9000:
```suggestion
	for i := range count {
```

api/stack_notification_batcher.go
```diff
@@ -0,0 +1,133 @@
...
+			close(nb.done)
+		}()
+
+		ticker := time.NewTicker(notificationBatchInterval)
+		defer ticker.Stop()
+
+		for {
+			select {
+			case <-ctx.Done():
+				// Use an uninterrupted new context here for the final flush
```
DrJosh9000: If this is exiting because the context is cancelled, that could happen while `flush` is already running, which would interrupt some number of calls to `nb.sendBatch`, which would probably cause those notifications to be discarded.

So do we need to wait for this flush?

zhming0: `flush` is blocking, so I don't think we will have two `flush` happen at the same time? (unless I understand goroutine's mechanism horribly wrong 😬 )

`nb.sendBatch` can happen in parallel there is no concern since it works on copied data. 

(This is the moment I miss immutable data structure)

api/stack_notification_batcher.go
zhming0: ohh I kinda miss the old go when where is no syntax sugar at all. But no problem.

api/stack_notification_batcher.go
zhming0: Ah good point! Done

api/stack_notification_batcher.go
DrJosh9000: It's not about data structures. Maybe we don't need to call `nb.flush` in this codepath at all. The HTTP calls in the flush in the other path (ticker channel receive) will have been interrupted by the context.

zhming0: Yeah I think I understand, is your main point that since we would've dropped some notifications anyway, so maybe this final flush aren't necessary?

Yeah, you are not wrong. The "complete" way to do graceful shutdown in this case would be maintaining a client connections tracker etc, which is a tad bit more tedious to write. I don't plan to do that (yet). 

So I guess the real question is whether the final flush help or not. My take is that it doesn't harm? wdty?

DrJosh9000: My main problem with the final flush as it currently exists is that it's detached from the parent context. There's no way for the caller to reliably abort the goroutine via the context it passed in. So if the final flush isn't necessary, do we need it? Or if we need a graceful shutdown, can we implement that?

api/stack_notification_batcher.go
DrJosh9000: > It is blocking the main batcher goroutine with wg currently.

Yes that is true. 

> By letting it self adapt (number of goroutine = number of batches), it adds little risk but a lot easier to maintain.

I think this is understating potential risks. With 1 API request per goroutine, there's a lot more happening than pure goroutines in the runtime, e.g. network connections and bandwidth to transmit them all, and threads in the backend to handle the requests. 

If the default limit on concurrent batches is 10, that's still 100K notifications/sec by your estimate. Perhaps instead of tuning it ourselves, we make it configurable?

zhming0: Maybe it would help if we can measure the risk here non-abstractly? 

With one goroutine per batch + `WaitGroup` setup we will be having at most `number of stack notification in one interval (150ms) / 1000` concurrent goroutines. Even if we have 1 millions notifications per interval (that is, roughly 10M notifications per second), we will have a 1000 parallel Goroutines in total, capped, which is 1000 HTTP connections outbound underneath the hood.  I would've assume that isn't a risk 🙂 . 

But you may have a different risk model, I am all 👂🏿 .

api/stack_notification_batcher.go
zhming0: that's okay, I removed the final flush. We will cross the bridge when the unclean shutdown becomes an issue. 

---- the below is purely for my learning. 

I am pretty curious about the "caller to reliably abort the goroutine via the context it passed in" rule. I guess it implied that context cancellation can't be used for graceful shutdown purpose? Since the semantic of cancel kinda implied that it needs to be instantaneous?

api/stack_notification_batcher.go
DrJosh9000: 1000 HTTP connections is very hefty for 1 client. That's at least 1000 TCP connections in HTTP/1. Importantly, if they translate into 1000 threads on the backend, that could be a significant chunk of serving capacity.

Strictly speaking, it could be higher. `flush` (which there is only one of running at a time) first reads `len(nb.notifications)`, and then reads that number of notifications into the slice. The channel capacity is 3 million, and there are 1000 notifications per batch, so there could be up to 3000 concurrent requests. (So I suppose it is implicitly bounded by 3000 right now.)

Your 100ms or 150ms estimate RTT for the request is probably reasonable, but glosses over some things. What happens if the backend starts taking longer to complete those requests? Those backend threads will be tied up for longer, which would have knock-on effects (more notifications backed up in the channel -> more batches -> more concurrent requests in the next iteration -> more threads in use).

api/stack_notification_batcher.go
DrJosh9000: The context package docs say: "A Context may be canceled to indicate that work done on its behalf should stop.". So context cancellation _can_ be implemented as either an immediate or a graceful shutdown. I think it should _always_ be made clear which is implemented. ("When you spawn goroutines, make it clear when or whether they exit." https://google.github.io/styleguide/go/decisions#goroutine-lifetimes).

What you had was a mix - requests during an in-progress flush would be cancelled immediately, followed up with an attempt at a graceful shutdown with the final flush. But there was no timeout or other way to cancel the final flush, because it was called with the background context. The graceful shutdown couldn't itself be gracefully or immediately shut down - it could have run a fairly long time after the original context was cancelled.

api/stack_notification_batcher.go
zhming0: I see! You are more thinking about the backend risk. Hmmm, I would've thought that our main protection is through rate limit on that front. (as we know, the default limit is way lower than the 3000 limit)

It's true that if a single client behave like this, compounded with other factors, we could be in trouble. Although, since our API is publicly documented, it's hard mitigate this risk from k8s stack alone. 

How about this? I could make a maximum batch number limit per interval to be, say 100. That will cap the open connection number to be 100 at most, and it would also cap the the throughput to be 100K/interval, but overall throughput is still substantial enough.

api/stack_notification_batcher.go
zhming0: Yeah, I wasn't a fan with the mix approach either. But my main thinking was that it was better losing some than losing all data during shutdown. 

Designing a graceful shutdown that itself can be graceful shutdown was also an blindspot on my end, in the past I've always rely on OS, k8s controller, global shutdown timer to handle that. I've yet to encounter a case where gracefully shutting down graceful shutdown is useful. 

Good learning

api/stack_notification_batcher.go
DrJosh9000: Yes, a cap would be good. I don't mind whether that would be in the form of `flush` only reading 100 chunks per call, or creating `min(100, len(chunks))` goroutines and farming out all pending notifications in chunks. 

I'm less certain about the number 100, but that's a fair bit safer than 1000, and I imagine it could be made configurable easily.

zhming0: Done. Another look please 🙏🏿 ?

DrJosh9000: Can you add a comment in the code that makes it clear that concurrent requests are being limited this way? i.e. that 100k notifications per interval / 1000 notifications per batch = at most 100 concurrent requests?

zhming0: Good idea, I added some comments in the top where the constants are defined.

PR #739, opened 2025-10-10
Don't deploy tags to the CI k8s stack (only main commits)

moskyb:


.buildkite/steps/deploy.sh
```diff
@@ -2,6 +2,23 @@
...
+  # $ git branch
+  #   main
+  # * feature-one
+  #   feature-two
+  # would mean we're on the feature-one branch
+  tag_branch=$(git branch --contains "tags/${BUILDKITE_TAG}" --sort committerdate | head -n 1 | cut -b 3-)
+  if [ "$tag_branch" != "main" ]; then
+    echo "Tag ${BUILDKITE_TAG} was not created from main branch (found: $tag_branch), skipping deployment"
+    exit 0
+  fi
```
zhming0: I wonder if we need to do branch check here at all? When we create a tag, it will always be from HEAD in main, which would have been deployed already. 

It's also an issue if we have this in main branch: 

```
commit A -> B - C
``` 

and we decided to make a tag for B, in such case we don't really want a deployment. 


Would this work? 

```suggestion
  exit 0
```

moskyb: > When we create a tag, it will always be from HEAD in main, which would have been deployed already.

not when we're tagging off of, for example, `v0.32.x`, which is a release branch for the 0.32 series

zhming0: but in that case, since it has a tag, we can filter it out? 

It seems this logic would suffice? `if tag -> no deploy`

moskyb: hmmm, good point. i think we can push this a level higher and do it using `if` in the pipeline. will investigate...

PR #741, opened 2025-10-10
PS-1176: detect command containter using BUILDKITE_COMMAND EnvVar

### LINKED ISSUE #697: [BUG] Resource classes are ignored if container name is set
## Describe the bug

If you use a resource class (as introduced in #668), the resources block is not respected for build steps that specify a container name other than `container-0`.

## To Reproduce

See pipeline:

```yaml
steps:
  - agents:
      resource_class: gpu-1
      queue: gpu-1
    plugins:
      - kubernetes:
          podSpec:
            containers:
              - image: nvidia/cuda:12.6.3-devel-ubuntu24.04
                command:
                  - nvidia-smi
  - agents:
      resource_class: gpu-1
      queue: gpu-1
    plugins:
      - kubernetes:
          podSpec:
            containers:
              - image: nvidia/cuda:12.6.3-devel-ubuntu24.04
                name: nvidia-smi # setting container name breaks resource allocation
                command:
                  - nvidia-smi
```

## Expected behavior

The first step works as expected. The second fails because it lacks the gpu resources. I expect both steps to work the same way.

## Environment
- agent-stack-k8s version: 0.31.0
- Kubernetes version: 1.31.11
- Deployment method: helm chart

## Additional context

This is noted as a limitation in a code comment: https://github.com/buildkite/agent-stack-k8s/blob/352706238d223e77eed4123c7f01f886f28a9c20/internal/controller/config/resource_class.go#L23-L24

However it's not documented in the [resource class documentation](https://buildkite.com/docs/agent/v3/agent-stack-k8s/container-resource-limits#using-resource-class).

We hit this problem because we wanted to use [sidecars](https://buildkite.com/docs/agent/v3/agent-stack-k8s/sidecars), and the first code example for a sidecar specifies a container name.

I think either the code should be fixed to work when you change the container name, or the limitation needs to be documented better.

zhming0:
When we introduced the image attribute and resource class, we (I) incorrectly assume that command container would all have a simple "name: container-0" pattern. 

This PR breaks this wrong assumption by using `BUILDKITE_COMMAND` and `BUILDKITE_BOOTSTRAP_PHASES` as the main detection mechanism. 

Closes #697

### DIFF
--- a/internal/controller/model/model.go
+++ b/internal/controller/model/model.go
@@ -8,6 +8,7 @@ import (
 	"github.com/buildkite/agent-stack-k8s/v2/api"
 
 	batchv1 "k8s.io/api/batch/v1"
+	corev1 "k8s.io/api/core/v1"
 )
 
 // ErrDuplicateJob is a sentinel error returned when a job has already been
@@ -38,5 +39,17 @@ func JobFinished(job *batchv1.Job) bool {
 			return true
 		}
 	}
+	return false
+}
+
+// Detect if a container is a buildkite command container.
+// The detection logic is a heuristic, but there is a very low likelihood of false positives.
+func IsCommandContainer(container *corev1.Container) bool {
+	for _, env := range container.Env {
+		if env.Name == "BUILDKITE_COMMAND" && env.Value != "" {
+			return true
+		}
+	}
+
 	return false
 }
--- a/internal/controller/scheduler/scheduler.go
+++ b/internal/controller/scheduler/scheduler.go
@@ -43,7 +43,7 @@ const (
 	CopyAgentContainerName        = "copy-agent"
 	ImageCheckContainerNamePrefix = "imagecheck-"
 	CheckoutContainerName         = "checkout"
-	CommandContainerName          = "container-0"
+	DefaultCommandContainerName   = "container-0"
 )
 
 var errK8sPluginProhibited = errors.New("the kubernetes plugin is prohibited by this controller, but was configured on this job")
@@ -450,7 +450,7 @@ func (w *worker) Build(podSpec *corev1.PodSpec, skipCheckout bool, inputs buildI
 	if len(podSpec.Containers) == 0 {
 		// Create a default command container
 		c := corev1.Container{
-			Name:         CommandContainerName,
+			Name:         DefaultCommandContainerName,
 			Image:        w.cfg.Image,
 			Command:      commandContainerCommand,
 			Args:         commandContainerArgs,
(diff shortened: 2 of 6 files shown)


internal/controller/scheduler/scheduler.go
```diff
@@ -43,7 +43,7 @@ const (
 	CopyAgentContainerName        = "copy-agent"
 	ImageCheckContainerNamePrefix = "imagecheck-"
 	CheckoutContainerName         = "checkout"
-	CommandContainerName          = "container-0"
```
moskyb, line 46: given that we no longer rely directly on the name being `container-0`, could we rename this to be `command` or something?

internal/controller/model/model.go
```diff
@@ -40,3 +42,24 @@ func JobFinished(job *batchv1.Job) bool {
 	}
 	return false
 }
+
+// Detect if a container is a buildkite command container.
+// The detection logic is a hueristic, meaning there is a very low likelihood of false positive.
+func IsCommandContainer(container *corev1.Container) bool {
+	var hasBuildkiteCommand bool
+	var hasBuildkiteBootstrapPhases bool
```
moskyb: [non-blocking]

it's equivalent, but IMO a little bit clearer to have

```suggestion
	hasBuildkiteCommand := false
	hasBuildkiteBootstrapPhases := false
```

as it more clearly shows we're starting from a point of assuming it's not the command container, then proving that it is

zhming0: I ended up realizing that we don't always set the bootstrap phases env var, so the code is simplified now.

internal/controller/scheduler/scheduler.go
zhming0: The name "container-0" is referenced in our doc in a few spots, changing this might be breaking change depending how podSpec is used. 

So the answer is probably "not now" :)

moskyb: sweetas! sounds good.

internal/controller/model/model.go
```diff
@@ -40,3 +41,15 @@ func JobFinished(job *batchv1.Job) bool {
 	}
 	return false
 }
+
+// Detect if a container is a buildkite command container.
+// The detection logic is a hueristic, meaning there is a very low likelihood of false positive.
```
DrJosh9000: A heuristic can be good or bad. In this case, I think the heuristic is good.

```suggestion
// The detection logic is a heuristic, but there is a very low likelihood of false positives.
```

internal/controller/scheduler/scheduler.go
DrJosh9000: I think it's still needed, since a `podSpecPatch`-only setup will be patching containers by name, right?

PR #743, opened 2025-10-13
PB-714 Stacks api failure message testing

moskyb:
At the moment, in the integration tests, when checking for failures to launch pods, we just check the logs. However, when the stacks API is enabled, we don't put pod launch failures into the logs, instead they go into a `stack_error` job event.

To get around this, when testing under the stacks API, we would  simply not check the error messages when testing pod launch failures -- there was a lot of code like
```Go
if !useStacksAPI {
  tc.AssertLogsContain(t, "oh no it failed")
}
```

and similar.

We're going to be globally enabling the stacks API soon, and we don't want to reduce test coverage, so we need to find a way to test pod launch failures under the stacks API.

This PR does just that! it adds a new query to the graphQL API to get all of the job events for a given job, and exposes a helper under `integration.testcase` to check the failure message of the last `stack_error` event.

This helper will check the logs if the stacks API is disabled, but that feature is not long for this world.

Ticket: [PB-714](https://linear.app/buildkite/issue/PB-714/remove-experimentalstackapissupport-flag-for-agent-k8s-stack-and-make)

### DIFF
--- a/internal/integration/integration_test.go
+++ b/internal/integration/integration_test.go
@@ -473,7 +473,6 @@ func TestSidecars(t *testing.T) {
 	tc.StartController(ctx, cfg)
 	build := tc.TriggerBuild(ctx, pipelineID)
 	tc.AssertSuccess(ctx, build)
-	time.Sleep(15 * time.Second)
 	tc.AssertLogsContain(build, "Welcome to nginx!")
 }
 
@@ -582,7 +581,6 @@ func TestEnvVariables(t *testing.T) {
 	tc.StartController(ctx, cfg)
 	build := tc.TriggerBuild(ctx, pipelineID)
 	tc.AssertSuccess(ctx, build)
-	time.Sleep(15 * time.Second) // trying to reduce flakes: logs not immediately available
 	tc.AssertLogsContain(build, "Testing some env variables: set")
 }
 
@@ -689,7 +687,6 @@ func TestInterposerBuildkite(t *testing.T) {
 	tc.StartController(ctx, cfg)
 	build := tc.TriggerBuild(ctx, pipelineID)
 	tc.AssertSuccess(ctx, build)
-	time.Sleep(15 * time.Second) // trying to reduce flakes: logs not immediately available
 	logs := tc.FetchLogs(build)
 	assert.Contains(t, logs, "Hello World!")
 	assert.Contains(t, logs, "Goodbye World!")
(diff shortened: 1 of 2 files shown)


internal/integration/testcase_test.go
```diff
@@ -341,6 +341,57 @@ func (t testcase) AssertCancelled(ctx context.Context, build api.Build) {
...
+	}
+
+	cj, ok := jobEvents.Job.(*api.GetJobEventsJobJobTypeCommand)
+	if !ok {
+		t.Fatalf("unexpected job type: %T", jobEvents.Job)
+	}
+
+	stackErrorEvents := make([]string, 0, 1) // Theoretically there can only ever be one errored event, but use a slice just in case.
+	for _, edge := range cj.Events.Edges {
+		if event, ok := edge.Node.(*api.GetJobEventsJobJobTypeCommandEventsJobEventConnectionEdgesJobEventEdgeNodeJobEventStackError); ok {
```
moskyb, line 386: perhaps one of the most beautiful type names i've ever laid eyes on (don't shoot me, it's generated)

internal/integration/integration_test.go
```diff
@@ -302,25 +299,6 @@ func TestControllerSetsAdditionalRedactedVars(t *testing.T) {
...
-	ctx := context.Background()
-	pipelineID := tc.PrepareQueueAndPipelineWithCleanup(ctx)
-
-	tc.StartController(ctx, cfg)
-	build := tc.TriggerBuild(ctx, pipelineID)
-	tc.AssertSuccess(ctx, build)
-	logs := tc.FetchLogs(build)
-	assert.Contains(t, logs, "The pre-checkout hook ran!")
-	assert.Contains(t, logs, "The post-checkout hook ran!")
-}
```
zhming0, line 322: Any reason to remove this test case?

internal/integration/testcase_test.go
```diff
@@ -341,6 +349,57 @@ func (t testcase) AssertCancelled(ctx context.Context, build api.Build) {
...
+	return ""
+}
+
+func (t testcase) FailureMessage(build api.Build, jobID string, useStacksAPI bool) string {
+	t.Helper()
+
+	if !useStacksAPI {
+		time.Sleep(10 * time.Second) // Wait for logs to be available
+		return t.FetchLogs(build)
+	}
```
zhming0, line 371: Combining with the sleep in FetchLogs, we have 25 seconds wait. Is this necessary?

zhming0 (commented):
Looks good to me overall (thank you!). With two semi blocking comments.

internal/integration/integration_test.go
moskyb: ah, it's in the commit message but the tl;dr is that this behaviour is already covered by another test, `TestHooksAndPlugins`

internal/integration/testcase_test.go
moskyb: oh good call! i thought i got rid of all the sleeps but i missed the most important one 😅

PR #745, opened 2025-10-14
Update k8s stack to always use the stacks API

moskyb:
The stacks API, which we've been working on in the background now for a little while, is ready for primetime! While previously enabled in an experimental fashion, this PR makes the stacks API the default, with no way to opt out to the old APIs (without downgrading the stack, anyway — the old APIs will continue to be supported for the foreseeable future, and nothing is going to break in older versions).

The use of the stacks API should only have a few user-facing changes, mostly around the job timeline of jobs picked up by the k8s stack — now, they'll have a little more info given by the k8s stack as to their scheduling process: 
CleanShot 2025-10-15 at 13 49 06

When the k8s stack fails to launch a job, the details of that failure will now be in the job timeline. for example, here's what the output for a job with a malformed `kubernetes` plugin config looks like: 
CleanShot 2025-10-15 at 13 56 50

However, the main benefits from the k8s stack using the stacks API come from the fact that it's designed to be much more scalable, and should be able to handle a higher job throughput.

As a procedural note, for the purposes of our CI instance, i've kept the CLI flag that enables the stacks API but it no longer does anything. it'll get removed in a later PR.

### DIFF
--- a/internal/controller/controller.go
+++ b/internal/controller/controller.go
@@ -29,12 +29,7 @@ import (
 
 const agentTokenKey = "BUILDKITE_AGENT_TOKEN"
 
-func Run(
-	ctx context.Context,
-	logger *zap.Logger,
-	k8sClient kubernetes.Interface,
-	cfg *config.Config,
-) {
+func Run(ctx context.Context, logger *zap.Logger, k8sClient kubernetes.Interface, cfg *config.Config) {
 	httpMuxes := make(map[string]*http.ServeMux)
 
 	if cfg.ProfilerAddress != "" {
@@ -125,7 +120,6 @@ func Run(
 		StackID:         cfg.ID,
 		AgentQueryRules: cfg.Tags,
 		Logger:          logger,
-		UseStacksAPI:    cfg.ExperimentalStacksAPISupport,
 	})
 	if err != nil {
 		logger.Error("Couldn't create Agent API client", zap.Error(err))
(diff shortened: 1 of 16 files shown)


internal/controller/controller.go
```diff
@@ -29,12 +29,7 @@ import (
 
 const agentTokenKey = "BUILDKITE_AGENT_TOKEN"
 
-func Run(
-	ctx context.Context,
-	logger *zap.Logger,
-	k8sClient kubernetes.Interface,
-	cfg *config.Config,
-) {
```
zhming0, line 37: [nit] it's probably unnecessary diff?  But it also reduce readability too

zhming0 (approved):
LGTM! with a nit.

internal/controller/controller.go
DrJosh9000: I think it's an improvement (see also https://google.github.io/styleguide/go/guide#line-length)

zhming0: It might depends on how we read it: 

>  If a line feels too long, prefer refactoring instead of splitting it. 

It still imply that long line isn't ideal.

> as short as it is practical for it to be

This is the gray area I guess, subject to personal preference.

In this case either way is fine. But if it has 5+ args I might feel slightly stronger about splitting.

moskyb: yeah, that's generally where i tend to create args structs etc, but this one is fine for now IMO

PR #748, opened 2025-10-21
Set {signal,cancel} grace period agent config

DrJosh9000:
### What

Instead of setting a Kubernetes-specific env var to set a grace period, use the agent's usual signal- and cancel-grace-period controls.

### Why

Over in buildkite/agent#3534, I determined that BUILDKITE_KUBERNETES_LOG_COLLECTION_GRACE_PERIOD didn't accomplish its goal, and removed it. But its heart is in the right place. We just need to extend the right grace periods. 

Signal-grace-period controls how long the agent will wait for a job to exit once it is interrupted before trying to kill it, and cancel-grace-period controls how long the agent will keep running things like the log streamer before giving up entirely. In the usual case these should be tied to the pod's `terminationGracePeriodSeconds`. Signal-grace-period should be a few seconds short of `terminationGracePeriodSeconds` in order to limit the worst case amount of time spent waiting for the job to wrap up, but still provide time to upload logs and mark the job as finished. Cancel-grace-period should extend all the way to `terminationGracePeriodSeconds` so that the agent continues trying to upload job logs and finish the job after the job is "killed". All containers get killed by Kubernetes at `terminationGracePeriodSeconds`.

### DIFF
--- a/internal/controller/scheduler/scheduler.go
+++ b/internal/controller/scheduler/scheduler.go
@@ -630,15 +630,33 @@ func (w *worker) Build(podSpec *corev1.PodSpec, skipCheckout bool, inputs buildI
 	// Set log collection grace period based on termination grace period.
 	// This allows the agent to coordinate log collection timing with pod termination.
 	if podSpec.TerminationGracePeriodSeconds != nil {
-		// Calculate log collection grace period: termination grace period minus buffer for cleanup
-		logCollectionGracePeriod := *podSpec.TerminationGracePeriodSeconds - 10
-		if logCollectionGracePeriod < 1 {
-			logCollectionGracePeriod = 1 // Minimum 1 second
-		}
-		agentContainer.Env = append(agentContainer.Env, corev1.EnvVar{
-			Name:  "BUILDKITE_KUBERNETES_LOG_COLLECTION_GRACE_PERIOD",
-			Value: fmt.Sprintf("%ds", logCollectionGracePeriod),
-		})
+		termGraceSecs := int(*podSpec.TerminationGracePeriodSeconds)
+		// When the agent cancels the job (e.g. when it receives SIGTERM), it
+		// first interrupts the job, waits for signal-grace-period, then
+		// terminates the job before waiting the remainder of the cancel-grace-
+		// period.
+		// When the pod is deleted, Kubernetes first sends SIGTERM to all
+		// containers. From Agent v3.110.0, kubernetes-bootstrap absorbs that
+		// signal and instead waits for the agent container to send an interrupt
+		// over the socket.
+		// Kubernetes will then take care of killing the job after
+		// TerminationGracePeriodSeconds is up. But we should still configure
+		// signal-grace-period so that the agent has time to upload logs and
+		// mark the job as finished, and also cancel-grace-period in case it
+		// needs to know how long it has left.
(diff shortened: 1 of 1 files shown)


internal/controller/scheduler/scheduler.go
```diff
@@ -630,15 +630,31 @@ func (w *worker) Build(podSpec *corev1.PodSpec, skipCheckout bool, inputs buildI
...
+		// When the pod is deleted, Kubernetes first sends SIGTERM to all
+		// containers. From Agent v3.110.0, kubernetes-bootstrap absorbs that
+		// signal and instead waits for the agent container to send an interrupt
+		// over the socket.
+		// Kubernetes will then take care of killing the job after
+		// TerminationGracePeriodSeconds is up. But we should still configure
+		// signal-grace-period so that the agent has time to upload logs and
+		// mark the job as finished.
+		signalGracePeriod := max(0, termGraceSecs-10)
+		cancelGracePeriod := max(termGraceSecs, signalGracePeriod+1)
```
zhming0: Since `signalGracePeriod = termGraceSecs - 10`, isn't this line effectively? This is kinda linked to my question below too.

```suggestion
		cancelGracePeriod := termGraceSecs
```

internal/controller/scheduler/scheduler.go
```diff
@@ -630,15 +630,31 @@ func (w *worker) Build(podSpec *corev1.PodSpec, skipCheckout bool, inputs buildI
...
+		agentContainer.Env = append(agentContainer.Env,
+			corev1.EnvVar{
+				Name:  "BUILDKITE_CANCEL_GRACE_PERIOD",
+				Value: strconv.Itoa(cancelGracePeriod),
+			},
+			corev1.EnvVar{
+				Name:  "BUILDKITE_SIGNAL_GRACE_PERIOD_SECONDS",
+				Value: strconv.Itoa(signalGracePeriod),
+			},
```
zhming0, line 658: ~~This is outside of this PR, but do we need the `BUILDKITE_CANCEL_GRACE_PERIOD` concept at all? Shouldn't an agent take as long as it wants to finish uploading logs, and leave it to upper layer to decide its patience?~~

actually, having a circuit breaker is good

zhming0 (approved):
👍🏿  with a minor question

internal/controller/scheduler/scheduler.go
DrJosh9000: It's an edge case, but has an effect. If `termGraceSecs` is 0 (valid, but seems silly to me to do that), then `signalGracePeriod` is 0 (not negative). But if `cancelGracePeriod` is also 0, then [the existing check in the agent](https://github.com/buildkite/agent/blob/a4dffdbed6b36623d52dfa1e2b3d677d6c53a96a/clicommand/cancel_signal.go#L63) will prevent it starting up.

zhming0: right, hmm, then should this be: `cancelGracePeriod := max(termGraceSecs, 1)`?

DrJosh9000: Maybe this would be simpler:

```go
// Note that the agent requires cancelGracePeriod > signalGracePeriod.
cancelGracePeriod := max(termGraceSecs, 1)
```

DrJosh9000: Done!

PR #752, opened 2025-10-22
Run release after pushing helm chart

DrJosh9000:
### What

Make `release` depend on `push-main-or-tag`.

### Why

The things that happen in `release` need the Helm chart to have been pushed. Previously this happened because of the transitive dependency via `deploy`, but as of #739, `deploy` no longer runs on tag builds.


moskyb:
😬 my bad, sorry about this!

PR #753, opened 2025-10-23
Use the standalone stacksapi module

moskyb:
We've published what used to be in the `internal/stacksapi` path in this repo to [its own repo](https://github.com/buildkite/stacksapi), so let's use that instead.


zhming0 (approved):
👍🏿

PR #754, opened 2025-10-23
PB-851 part 1: use Buildkite Secret in favor of K8s Secret

zhming0:
This PR replaces all k8s secret reference to Buildkite secrets with proper access policy. 

Note, because current Buildkite Secret restrict BUILDKITE_** env var names, I had to rename some env var names.

### DIFF
--- a/.buildkite/steps/deploy.sh
+++ b/.buildkite/steps/deploy.sh
@@ -18,7 +18,7 @@ helm upgrade agent-stack-k8s "${helm_repo_pecr}/agent-stack-k8s" \
   --install \
   --create-namespace \
   --wait \
-  --set agentToken="${BUILDKITE_AGENT_TOKEN}" \
+  --set agentToken="${DEFAULT_CLUSTER_TOKEN}" \
   --set config.image="${agent_image}" \
   --set config.debug=true \
   --set config.profiler-address=localhost:6060 \
--- a/cmd/controller/controller_test.go
+++ b/cmd/controller/controller_test.go
@@ -159,12 +159,8 @@ func TestReadAndParseConfig(t *testing.T) {
 		},
 	}
 
-	// The buildkite token is required, but it is set from a Kubernetes secret, not the config file,
-	// which is itself set from a config map that is used to create env variables in the controller
-	// container. As this is required, we set it here to avoid the validation error.
-	t.Setenv("BUILDKITE_TOKEN", "my-graphql-enabled-token")
-
 	// These need to be unset to as it is set in CI which pollutes the test environment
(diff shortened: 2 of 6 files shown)


.buildkite/steps/deploy.sh
```diff
@@ -18,7 +18,7 @@ helm upgrade agent-stack-k8s "${helm_repo_pecr}/agent-stack-k8s" \
   --install \
   --create-namespace \
   --wait \
-  --set agentToken="${BUILDKITE_AGENT_TOKEN}" \
+  --set agentToken="${DEFAULY_CLUSTER_TOKEN}" \
```
moskyb:
```suggestion
  --set agentToken="${DEFAULT_CLUSTER_TOKEN}" \
```

cmd/controller/controller_test.go
```diff
@@ -162,7 +162,7 @@ func TestReadAndParseConfig(t *testing.T) {
 	// The buildkite token is required, but it is set from a Kubernetes secret, not the config file,
 	// which is itself set from a config map that is used to create env variables in the controller
 	// container. As this is required, we set it here to avoid the validation error.
```
moskyb: This comment is a lie now, right? It's set from a buildkite secret during the deploy process

cmd/controller/flags.go
```diff
@@ -17,7 +17,7 @@ func AddConfigFlags(cmd *cobra.Command) {
 		"buildkite-agent-token",
 		"name of the Buildkite agent token secret",
 	)
-	cmd.Flags().String("buildkite-token", "", "Deprecated - Buildkite API token with GraphQL scopes")
+	cmd.Flags().String("integration-test-buildkite-token", "", "Deprecated - Buildkite API token with GraphQL scopes")
```
moskyb, line 20: i think now that we're only using the graphQL token in tests, we should remove it from the stack entirely and pass it to the tests by environment variable or similar — its presence in the stack configuration is only going to confuse people

.buildkite/steps/deploy.sh
swaller-bk: smol typo here I think: 
```--set agentToken="${DEFAULT_CLUSTER_TOKEN}" \```

cmd/controller/controller_test.go
zhming0: This comment is confusing, I removed it and reworked this bit of code.

cmd/controller/flags.go
zhming0: Totally agree, but that's probably out-of-scope for this PR.

PR #755, opened 2025-10-23
PB-851 part 2: use podTemplate + image attribute to hide podSpec

zhming0:
use podTemplate + image attribute to hide podSpecs, largely simplify our `pipeline.yaml`!

### DIFF
--- a/.buildkite/pipeline.yaml
+++ b/.buildkite/pipeline.yaml
@@ -13,110 +13,38 @@ x-anchors:
           checkout:
             fetchFlags: -v --tags
 
-  go-cache-env: &go-cache-env
-    name: GOCACHE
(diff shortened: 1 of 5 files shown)


.buildkite/pipeline.yaml
```diff
@@ -125,35 +53,13 @@ steps:
...
     plugins:
       - kubernetes:
-          podSpec:
+          podTemplate: go-with-cache
+          podSpecPatch:
             serviceAccountName: integration-tests
-            volumes:
-              - name: agent-stack-k8s-config
-                configMap:
-                  name: agent-stack-k8s-config
```
zhming0, line 135: There is no need for our integration test to access our production controller config.

PR #756, opened 2025-10-23
Ground out zap

moskyb:
This PR does two things:
1. Replace `zap` with go's builtin `log/slog`
2. Adds a couple of logging UX improvements:
- Adds a `--log-format` flag which takes either `logfmt` (default) or `json` (new!), allowing us to output JSON logs
- Colours the logs by default when the k8s stack is run in a TTY (which in production is almost never, but makes local dev more bearable)
- Adds a `--log-level` flag, allowing users to customise how many logs the k8s stack outputs into their environment

The zap-to-slog transition is basically a giant find-and-replace, which Amp mostly handled on my part. The new flags are good old-fashioned meat code.

**This PR will be easier to review commit-by-commit**

Thanks to @wolfeidau for the inspiration (https://github.com/buildkite/zstash/pull/81)

### DIFF
--- a/cmd/controller/controller.go
+++ b/cmd/controller/controller.go
@@ -16,6 +16,8 @@ import (
 	"github.com/buildkite/agent-stack-k8s/v2/internal/controller"
 	"github.com/buildkite/agent-stack-k8s/v2/internal/controller/config"
 	"github.com/buildkite/agent-stack-k8s/v2/internal/controller/scheduler"
+	"github.com/lmittmann/tint"
+	"github.com/mattn/go-isatty"
 
 	"github.com/go-playground/locales/en"
 	ut "github.com/go-playground/universal-translator"
@@ -281,12 +283,37 @@ func New() *cobra.Command {
 				return fmt.Errorf("failed to parse config: %w", err)
 			}
 
-			var logger *slog.Logger
-			if cfg.Debug {
-				logger = slog.New(slog.NewJSONHandler(os.Stdout, &slog.HandlerOptions{Level: slog.LevelDebug}))
-			} else {
-				logger = slog.New(slog.NewJSONHandler(os.Stdout, &slog.HandlerOptions{Level: slog.LevelInfo}))
+			level := slog.LevelInfo
+			if cfg.LogLevel != "" {
+				err := level.UnmarshalText([]byte(cfg.LogLevel))
+				if err != nil {
+					return fmt.Errorf("invalid log level %q: %w", cfg.LogLevel, err)
+				}
 			}
+
+			if cfg.Debug {
+				// The debug flag trumps the log level flag
+				level = slog.LevelDebug
+			}
+
+			var handler slog.Handler
+			switch cfg.LogFormat {
+			case "json":
+				handler = slog.NewJSONHandler(os.Stdout, &slog.HandlerOptions{Level: level})
+
+			case "logfmt":
+				handler = tint.NewHandler(os.Stdout, &tint.Options{
+					AddSource: true,
+					Level:     level,
+					NoColor:   cfg.NoColor || !isatty.IsTerminal(os.Stdout.Fd()),
+				})
+
+			default:
+				// Handled by the config validator above, but nice to have just in case.
+				return fmt.Errorf("unknown log format: %s", cfg.LogFormat)
+			}
+
+			logger := slog.New(handler)
 			logger.Info("configuration loaded", "config", cfg)
 
 			clientConfig := restconfig.GetConfigOrDie()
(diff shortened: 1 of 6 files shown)


cmd/controller/controller.go
```diff
@@ -281,15 +283,37 @@ func New() *cobra.Command {
...
+			var handler slog.Handler
+			switch cfg.LogFormat {
+			case "json":
+				handler = slog.NewJSONHandler(os.Stdout, &slog.HandlerOptions{Level: level})
+
+			case "console":
+				handler = tint.NewHandler(os.Stdout, &tint.Options{
+					Level:   level,
+					NoColor: cfg.NoColor || !isatty.IsTerminal(os.Stdout.Fd()),
+				})
```
zhming0: I believe the default key value text log format is commonly referred to as `logfmt`. 

```suggestion
			case "logfmt":
				handler = tint.NewHandler(os.Stdout, &tint.Options{
					Level:   level,
					NoColor: cfg.NoColor || !isatty.IsTerminal(os.Stdout.Fd()),
				})
```

internal/controller/config/config.go
```diff
@@ -114,70 +113,10 @@ type Config struct {
...
-type stringSlice []string
 
-func (s stringSlice) MarshalLogArray(enc zapcore.ArrayEncoder) error {
-	for _, x := range s {
-		enc.AppendString(x)
-	}
-	return nil
-}
-
-func (c Config) MarshalLogObject(enc zapcore.ObjectEncoder) error {
```
zhming0, line 128: [question] Is this capability no longer needed?

internal/controller/monitor/monitor.go
```diff
@@ -151,13 +151,13 @@ func (m *Monitor) Start(ctx context.Context, handler model.ManyJobHandler) <-cha
...
 				continue
-			}
+				}
 
-			m.logger.Info("job processing of Agent API results completed...",
-				zap.Int("agent-api-jobs-processed", len(jobs)),
-			)
+				m.logger.Info("job processing of Agent API results completed...",
+				"agent-api-jobs-processed", len(jobs),
+				)
```
zhming0: Amp did some funky indentation change?

internal/controller/controller.go
```diff
@@ -149,12 +149,13 @@ func Run(ctx context.Context, logger *zap.Logger, k8sClient kubernetes.Interface
...
 	})
 	if err != nil {
-		logger.Fatal("failed to create monitor", zap.Error(err))
+		logger.Error("failed to create monitor", "error", err)
+		return
 	}
 
 	// Scheduler does the complicated work of converting a Buildkite job into
 	// a pod to run that job. It talks to the k8s API to create pods.
-	sched := scheduler.New(logger.Named("scheduler"), k8sClient, agentClient, scheduler.Config{
```
zhming0, line 157: I didn't verify if the original `logger.Named` is equavelent to `logger.With("component", ...)`, hopefully it's the same?

moskyb: in zap, the name of the logger is included in the log line. it's what's highlighted here:
CleanShot 2025-10-24 at 13 47 49@2x

In the new `slog` logs, it looks like this:
CleanShot 2025-10-24 at 13 49 11@2x

(the equivalent being `component=monitor`).

The two formats aren't the same (which can be seen just by looking at the screenshots), but i don't think that the format of our logs is an API surface we need to maintain just yet, and looking forward, i would expect the `json` format to stay stable while the `logfmt` format may change

internal/controller/monitor/monitor.go
moskyb: oh yes indeed. lemme go in and fix those...

internal/controller/config/config.go
moskyb: not really. `MarshalLogObject` (and the `MarshalLogArray` method above) are used by zap to encode custom types into log values. slog is smart enough to just call `fmt.Sprintf("%v", val)` on it, and it gets output like this:

Image

not exactly pretty, but then it doesn't have to be. if we wanted to duplicate the behaviour using `slog`, we could do something like:
```Go
func (c Config) LogValue() slog.Value {
    return slog.GroupValue(
        slog.String("agent-token-secret", c.AgentTokenSecret),
        slog.Bool("debug", c.Debug),
        slog.String("image", c.Image),
        // ... etc
    )
}
```

... but tbh i don't really think it's worth it, as it wouldn't make the outputted log much prettier, or much more useful.

cmd/controller/controller.go
moskyb: how do you feel about `text`? it matches the agent, and saying `--log-format logfmt` looks a little silly

zhming0: I personally feel `logfmt` is a widely recognized word while `text` is pretty ambiguous, JSON is `text` too 😬.

moskyb: sweet, `logfmt` it is

PR #759, opened 2025-10-27
PB-862: use Kubernetes native sidecar mechanism

zhming0:
Historically,  k8s didn't have a main container vs sidecar container concept. As a result we maintain a in-house built sidecar pattern for agent-k8s stack, it means we have to implement many code to make a container sidecar: it includes relying on `activeDeadlineSeconds` to kill the pod after "main" container is done etc.

It also isn't ideal because it can leave a k8s job in a status of `type: Failed with reason: DeadlineExceeded` despite it successfully handled the bk job. 

This PR changes k8s stack to use Kuberenetes' [native sidecar pattern](https://kubernetes.io/docs/concepts/workloads/pods/sidecar-containers/).


internal/integration/fixtures/extra-volume-mounts-sidecar.yaml
```diff
@@ -23,6 +23,10 @@ steps:
                   mv /tmp/extra-volume-mount-sidecar/pass-the-parcel /tmp/extra-volume-mount/pass-the-parcel
                   ls -lah /tmp/extra-volume-mount-sidecar
                   ls -lah /tmp/extra-volume-mount
+                  # So Kubernetes can mark this as Started
+                  # In reality a sidecar should not be used like this.
+                  # As this is effectively an ad-hoc initContainer.
+                  sleep 120
```
zhming0: This is a potential breaking change. In the new sidecar implementation, the sidecar container is expected to be long running and never finish.

internal/controller/scheduler/completions.go
```diff
@@ -68,15 +68,22 @@ func (w *completionsWatcher) OnUpdate(old any, new any) {
...
+//
+// Whether we still need it is subject to debate.
+// In the past, before k8s introduced sidecar concept and we adopted it, we rely on this to clean up those pods with
+// (our in-house built) sidecar containers.
+// Now it's no longer necessary.
+//
+// I am leaning to keep it because:
+// 1. To capture the edge case where agent container ends successfully, but command containers somehow left hanging.
+// 2. This is very cheap operation.
+func (w *completionsWatcher) setJobDealine(ctx context.Context, pod *v1.Pod) {
```
DrJosh9000:
```suggestion
func (w *completionsWatcher) setJobDeadline(ctx context.Context, pod *v1.Pod) {
```

internal/controller/scheduler/completions.go
```diff
@@ -68,15 +68,22 @@ func (w *completionsWatcher) OnUpdate(old any, new any) {
 	}
 
 	newPod := new.(*v1.Pod)
-	w.cleanupSidecars(w.resourceEventHandlerCtx, newPod)
+	w.setJobDealine(w.resourceEventHandlerCtx, newPod)
 }
 
-// cleanupSidecars first checks if the container status of the agent container
+// setJobDealine first checks if the container status of the agent container
```
DrJosh9000:
```suggestion
// setJobDeadline first checks if the container status of the agent container
```

internal/controller/scheduler/pod_watcher.go
```diff
@@ -14,9 +14,10 @@ import (
 	"github.com/buildkite/agent/v3/agent"
 	"github.com/buildkite/roko"
 
+	"log/slog"
```
DrJosh9000: This actually belongs in the first group (standard library packages).

internal/controller/scheduler/completions.go
```diff
@@ -68,15 +68,22 @@ func (w *completionsWatcher) OnUpdate(old any, new any) {
...
-// (So this is not actually sidecar-specific, but is needed because sidecars
-// would otherwise cause the pod to continue running.)
-func (w *completionsWatcher) cleanupSidecars(ctx context.Context, pod *v1.Pod) {
+//
+// Whether we still need it is subject to debate.
+// In the past, before k8s introduced sidecar concept and we adopted it, we rely on this to clean up those pods with
+// (our in-house built) sidecar containers.
+// Now it's no longer necessary.
+//
+// I am leaning to keep it because:
```
DrJosh9000: I vote in favour of deleting the whole completion watcher. 

If the containers are misbehaving so badly that they are blocking the pod finishing, that sounds like a serious bug we should fix. Leaving the pod running would enable more ways to diagnose.

internal/integration/fixtures/extra-volume-mounts-sidecar.yaml
DrJosh9000: Hmm, true. We'll have to call this out carefully in the release notes.

internal/controller/scheduler/completions.go
zhming0: No objection from me. It's gone now 🫡

DrJosh9000:
> This change also adds a new minimum cluster version requirement of `1.33.0`. 

[the feature is active by default since Kubernetes v1.29](https://kubernetes.io/docs/concepts/workloads/pods/sidecar-containers/)

> We will need to ensure this is distinctly called out in our docs, as a controller upgrade in an older K8s cluster would break the customer's deployment. Perhaps we should add a minimum cluster version startup "check" to the controller and exit early?

Yes and yes.

zhming0:
I am hoping to do a scan to find out version distribution in our existing users. But we don't seem to collect this information. 

Nor do we have a way to engage with the crowd en mass. 

So we will have to make a small bet. If it helps, `1.31` EOL is yesterday. 

> 1.31
> Latest Release:1.31.13 (released: 2025-09-09)
> End of Life:2025-10-28

but yeah, 100% on a `kubeVersion` field.

zhming0:
ready for another review @DrJosh9000 🙏🏿

internal/controller/scheduler/scheduler.go
```diff
@@ -24,9 +24,10 @@ import (
 	"github.com/buildkite/agent/v3/agent"
 	"github.com/buildkite/agent/v3/clicommand"
 
+	"log/slog"
```
DrJosh9000: Put with the stdlib imports

charts/agent-stack-k8s/Chart.yaml
```diff
@@ -22,3 +22,5 @@ version: 0.1.0
 # follow Semantic Versioning. They should reflect the version the application is using.
 # It is recommended to use it with quotes.
 appVersion: "1.16.0"
+
+kubeVersion: ">=1.29.0"
```
DrJosh9000, line 28:
```suggestion
# Sidecar support requires the SidecarContainers feature gate, first enabled
# by default in Kubernetes 1.29.
kubeVersion: ">=1.29.0"
```

internal/integration/fixtures/extra-volume-mounts-sidecar.yaml
DrJosh9000: A quick play with `alpine:latest` shows that it supports `infinity` - maybe we should do this?

```suggestion
                  sleep infinity
```

DrJosh9000 (approved):
I wish there were a way to test for individual feature gates in Chart.yaml rather than k8s versions

Anyway, LGTM with some minor comments

PR #762, opened 2025-10-30
Add explicit queue config option

moskyb:
Specifying the queue based on the tags is weird and busted, and is based on agent functionality that we've [since simplified](https://github.com/buildkite/agent/pull/2648) by... adding a `--queue` flag. Let's echo what the agent does here, and make life a bit easier for our users.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,7 +28,7 @@ To specify a non-default queue:
 ```sh
 helm install agent-stack-k8s oci://ghcr.io/buildkite/helm/agent-stack-k8s \
     --set agentToken=<buildkite-agent-token> \
-    --set-json 'config.queue="arm64"'
+    --set config.queue=arm64
 ```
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -23,12 +23,12 @@ helm install agent-stack-k8s oci://ghcr.io/buildkite/helm/agent-stack-k8s \
...
 ```
 
-Specify a different queue using tags:
+To specify a non-default queue:
 
 ```sh
 helm install agent-stack-k8s oci://ghcr.io/buildkite/helm/agent-stack-k8s \
     --set agentToken=<buildkite-agent-token> \
-    --set-json 'config.tags=["queue=arm64"]'
+    --set-json 'config.queue="arm64"'
```
DrJosh9000: This can probably be simplified now:

```suggestion
    --set config.queue=arm64
```

DrJosh9000 (approved):
Nice work!

PR #771, opened 2025-11-23
PB-946 part 1: backfill unit test for reserver

zhming0:
Add unit test for reserver, locking in the batching behavior. 

As part of it, introducing fake agent server, that can be handy during test. (this fake agent server was _mostly_ written by Claude)

### DIFF
--- a/api/fake_agent_server.go
+++ b/api/fake_agent_server.go
@@ -1,0 +1,148 @@
+package api
+
+import (
+	"bytes"
+	"encoding/json"
+	"log"
+	"net/http"
+	"net/http/httptest"
+	"sync"
+
+	"github.com/buildkite/stacksapi"
+)
+
+// FakeAgentServer is an httptest.Server that implements the Buildkite Agent API.
+// It records all requests for verification in tests.
+// Use with a real AgentClient for realistic integration testing.
+type FakeAgentServer struct {
+	server *httptest.Server
+	mu     sync.Mutex
+
+	// ReserveCalls records the job IDs passed to each ReserveJobs call.
+	ReserveCalls [][]string
+
+	// ReserveResponse configures the response for ReserveJobs.
+	// If nil, returns all job IDs as reserved.
+	ReserveResponse *stacksapi.BatchReserveJobsResponse
+
+	// ReserveStatusCode configures the HTTP status code for ReserveJobs.
+	// Default is 200.
+	ReserveStatusCode int
+
+	// ReserveError configures an error message to return.
+	ReserveError string
+}
+
+// NewFakeAgentServer creates and starts a fake agent API server.
+// Use server.URL() to get the endpoint for creating a real AgentClient.
+func NewFakeAgentServer() *FakeAgentServer {
+	fake := &FakeAgentServer{
+		ReserveStatusCode: http.StatusOK,
+	}
+
+	mux := http.NewServeMux()
+	mux.HandleFunc("/stacks/register", fake.handleRegisterStack)
+	mux.HandleFunc("/stacks/test-stack/scheduled-jobs/batch-reserve", fake.handleReserveJobs)
+
+	fake.server = httptest.NewServer(mux)
+	return fake
+}
+
+// URL returns the base URL of the fake server.
+// Use this as the Endpoint when creating a real AgentClient.
+func (f *FakeAgentServer) URL() string {
+	return f.server.URL
+}
+
+// Close shuts down the fake server.
+func (f *FakeAgentServer) Close() {
+	f.server.Close()
+}
+
+func (f *FakeAgentServer) handleRegisterStack(w http.ResponseWriter, r *http.Request) {
+	if r.Method != http.MethodPost {
+		http.Error(w, "method not allowed", http.StatusMethodNotAllowed)
+		return
+	}
+
+	var req stacksapi.RegisterStackRequest
+	if err := json.NewDecoder(r.Body).Decode(&req); err != nil {
+		http.Error(w, err.Error(), http.StatusBadRequest)
+		return
+	}
+
+	resp := stacksapi.RegisterStackResponse{
+		Key:             req.Key,
+		ClusterQueueKey: req.QueueKey,
+		Metadata:        req.Metadata,
+	}
+
+	var buf bytes.Buffer
+	if err := json.NewEncoder(&buf).Encode(resp); err != nil {
+		http.Error(w, "fake: failed to encode response: "+err.Error(), http.StatusInternalServerError)
+		return
+	}
+
+	w.Header().Set("Content-Type", "application/json")
+	if _, err := w.Write(buf.Bytes()); err != nil {
+		log.Printf("fake: failed to write register stack response: %v", err)
+	}
+}
+
+func (f *FakeAgentServer) handleReserveJobs(w http.ResponseWriter, r *http.Request) {
+	if r.Method != http.MethodPut && r.Method != http.MethodPost {
+		http.Error(w, "method not allowed", http.StatusMethodNotAllowed)
+		return
+	}
+
+	var req stacksapi.BatchReserveJobsRequest
+	if err := json.NewDecoder(r.Body).Decode(&req); err != nil {
+		http.Error(w, err.Error(), http.StatusBadRequest)
+		return
+	}
+
+	// Record the call (mutex protects concurrent access from HTTP handler)
+	f.mu.Lock()
+	f.ReserveCalls = append(f.ReserveCalls, req.JobUUIDs)
+	f.mu.Unlock()
+
+	// Return configured error if set
+	if f.ReserveError != "" {
+		var buf bytes.Buffer
+		if err := json.NewEncoder(&buf).Encode(map[string]string{
+			"message": f.ReserveError,
+		}); err != nil {
+			http.Error(w, "fake: failed to encode error response: "+err.Error(), http.StatusInternalServerError)
+			return
+		}
+
+		w.Header().Set("Content-Type", "application/json")
+		w.WriteHeader(f.ReserveStatusCode)
+		if _, err := w.Write(buf.Bytes()); err != nil {
+			log.Printf("fake: failed to write error response: %v", err)
+		}
+		return
+	}
+
+	// Return configured response
+	resp := f.ReserveResponse
+	if resp == nil {
+		// Default: all jobs reserved successfully
+		resp = &stacksapi.BatchReserveJobsResponse{
+			Reserved:    req.JobUUIDs,
+			NotReserved: []string{},
+		}
+	}
+
+	var buf bytes.Buffer
+	if err := json.NewEncoder(&buf).Encode(resp); err != nil {
+		http.Error(w, "fake: failed to encode response: "+err.Error(), http.StatusInternalServerError)
+		return
+	}
+
+	w.Header().Set("Content-Type", "application/json")
+	w.WriteHeader(f.ReserveStatusCode)
+	if _, err := w.Write(buf.Bytes()); err != nil {
+		log.Printf("fake: failed to write reserve jobs response: %v", err)
+	}
+}
--- a/api/fake_agent_server_test.go
+++ b/api/fake_agent_server_test.go
@@ -1,0 +1,144 @@
+package api_test
+
+import (
+	"context"
+	"log/slog"
+	"testing"
+
+	"github.com/buildkite/agent-stack-k8s/v2/api"
+	"github.com/buildkite/stacksapi"
+	"github.com/google/go-cmp/cmp"
+)
+
+func TestFakeAgentServer_DefaultBehavior(t *testing.T) {
+	ctx := context.Background()
+
+	// Create fake server
+	server := api.NewFakeAgentServer()
+	defer server.Close()
+
+	// Create REAL AgentClient pointing to fake server
+	client, err := api.NewAgentClient(ctx, api.AgentClientOpts{
+		Token:    "fake-token",
+		Endpoint: server.URL(),
+		StackID:  "test-stack",
+		Logger:   slog.Default(),
+	})
+	if err != nil {
+		t.Fatalf("NewAgentClient() error = %v", err)
+	}
+
+	jobIDs := []string{"job-1", "job-2", "job-3"}
+	result, retryAfter, err := client.ReserveJobs(ctx, jobIDs)
+	if err != nil {
+		t.Errorf("ReserveJobs(ctx, %q) error = %v, want nil", jobIDs, err)
+	}
+	if retryAfter != 0 {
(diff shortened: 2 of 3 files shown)


api/fake_agent_client.go
```diff
@@ -0,0 +1,89 @@
...
+	"fmt"
+	"sync"
+	"time"
+
+	"github.com/buildkite/stacksapi"
+)
+
+// FakeAgentClient is a test double for AgentClient. It records all method calls
+// and allows configuring return values for testing.
+type FakeAgentClient struct {
```
DrJosh9000: I think we definitely should not fake the client side of the API. I think we should fake the server side - run a httptest server in each test.

zhming0: We already have pretty extensive integration test coverage that includes real server connectivity check. For the interest of easier writing unit test, is client level mock harmful? 

I could make FakeAgentClient spawn a httptest server, though it probably add negligible gain in our testability or coverage?

DrJosh9000: Faking the client side is basically a test that the code is calling some methods. It's low fidelity and requires introducing an interface or some other method for injecting the dependency at the code level.

Faking the server side is scarcely any more effort to create, yet ensures that an actual HTTP request is sent and received, request and response bodies are correctly encoded and decoded, and only requires reconfiguring the client side to use the test server.

zhming0: I changed how the fake client is implemented. I am happy to use httptest to increase fidelity. 

But, there are two things we aren't aligned on. Are you happy with following statements? 

- The intention of reserver test is specifically about verifying a number of method calls with correct parameters, regardless whether the mock happens on interface level or lower http level. 
- The new interface isn't harmful. Per [go styleguide](https://google.github.io/styleguide/go/decisions#interfaces):
> Go interfaces generally belong in the package that consumes values of the interface type, not a package that implements the interface type.

DrJosh9000: > The intention of reserver test is specifically about verifying a number of method calls with correct parameters, regardless whether the mock happens on interface level or lower http level.

We don't need tests at all to verify things. A formal proof would suffice. But that's too onerous. So instead we have tests. Tests don't formally verify anything, but they do provide evidence that some things work as designed. 

I don't think the intention of any test should be to check some specific number of method calls with some specific parameters, I think it's better to check that the code under test does what it is intended to do.

> The new interface isn't harmful.

I'm not arguing that the new interface is _harmful_, I'm saying that by faking the server side it becomes _unnecessary_.

---

Please delete the interface and do not use `FakeAgentClient` as a wrapper that hides `httptest`.

zhming0: My intention was to use a test case to prove that the reserver's side effect meets our expectation. The side effect happen to be a number of a API calls with specific parameters. 

I am happy to meet people's preference on how things should be written, but some pros and cons analysis will be very appreciated 🙂 . SWE is 10% about science 90% about people, so things can go so many different directions. 

Are you happy with the following plan?

1. Implement a fake agent server. 
2. During reserver test, spawn real agent server, but with its endpoint pointing to fake agent server.
3. During test, tap into the fake agent server to assert API calls.

zhming0: I did the change above, another look please. I guess the benefit is that now reserver remains unchanged 👍🏿 .

DrJosh9000: > The side effect happen to be a number of a API calls with specific parameters.

It's true that both approaches could be characterised as testing for "a number of (method/API) calls with specific parameters". But if we're doing pros/cons:

### Fake client with internal interface

Pros:

- Doesn't require anything listening on a network port
- Avoids considering the visible API, only internal details

Cons:

- Requires adding an interface to enable dependency injection, but there's only one "real" implementation of the interface
- Doesn't test that actual API requests are generated and received
- Doesn't test that actual API responses are received and understood
- Fake client very likely to need updating as internal details change (e.g. suppose the client's methods need refining; the interface and fake then need updating as well)

### Real client, fake server

Pros:

- No new interface needed
- Tests that actual API requests are generated and sent
- Tests that actual API responses are received and understood
- Fake server less likely to need updating as client-side internal details change (the methods on the client can change but this doesn't necessarily impact the API)

Cons:

- Requires starting a test server during tests
- Requires thinking about the API (this might be a pro rather than a con?)


> Are you happy with the following plan?

That plan is definitely more agreeable, though I think you mean

> During reserver test, spawn real agent **client**, but with its endpoint pointing to fake agent server.

zhming0: Thank you for writing it up 💚💚, I appreciate it and I have updated the PR per your feedback. 

---

I agree the most of them, except on one topic, which I think is our real divergence: 

> Requires adding an interface to enable dependency injection, but there's only one "real" implementation of the interface

I was under the impression that having smaller interface is generally encouraged in Go? 

In this particular case, faking our own API is very doable, but in different context, say if the reserver is calling the k8s API or some gRPC service, would you agree that mocking on client interface level is a good middle ground?

DrJosh9000: > I was under the impression that having smaller interface is generally encouraged in Go?

_If you need an interface,_ then having an interface that specifies fewer methods is generally better. Avoiding adding an unnecessary interface in the first place avoids breakage due to looser coupling, e.g. during refactoring.

> if the reserver is calling the k8s API or some gRPC service, would you agree that mocking on client interface level is a good middle ground?

No, not at all! If we can't or don't want to call the real k8s API or the real gRPC service in question, then we should look for configurable fakes/test doubles provided by k8s or the service being called, and if they don't exist, write our own, which is usually quite easy because the API or .proto files are available and we can generate stubs.

Where the style guide says:

> To test packages that use Stubby RPC clients, use a real client connection. If a real server cannot be run in the test, Google’s internal practice is to obtain a real client connection to a local [test double](https://abseil.io/resources/swe-book/html/ch13.html#basic_concepts) using the internal rpctest package (coming soon!).

substitute "Stubby" with "gRPC" (because they are practically the same thing, outside of Google anyway).

api/fake_agent_server.go
```diff
@@ -0,0 +1,176 @@
...
+		resp = stacksapi.BatchReserveJobsResponse{
+			Reserved:    jobIDsCopy,
+			NotReserved: []string{},
+		}
+	}
+
+	w.Header().Set("Content-Type", "application/json")
+	w.WriteHeader(f.ReserveStatusCode)
+	if err := json.NewEncoder(w).Encode(resp); err != nil {
+		panic(fmt.Sprintf("fake: failed to encode reserve jobs response: %v", err))
```
DrJosh9000: Don't panic - instead consider encoding the response to a `bytes.Buffer` (which prevents any `io.Writer` related errors) and then if there's a JSON encoding error, use `http.Error`.

api/fake_agent_server.go
```diff
@@ -0,0 +1,176 @@
...
+	var resp stacksapi.BatchReserveJobsResponse
+	if f.ReserveResponse != nil {
+		resp = *f.ReserveResponse
+	} else {
+		// Default: all jobs reserved successfully
+		resp = stacksapi.BatchReserveJobsResponse{
+			Reserved:    jobIDsCopy,
+			NotReserved: []string{},
+		}
+	}
```
DrJosh9000: A small reduction, but makes it feel like "nil handling".

```suggestion
	resp := f.ReserveResponse
	if resp == nil {
		// Default: all jobs reserved successfully
		resp = &stacksapi.BatchReserveJobsResponse{
			Reserved:    jobIDsCopy,
			NotReserved: []string{},
		}
	}
```

api/fake_agent_server.go
```diff
@@ -0,0 +1,176 @@
...
+// GetReserveCalls returns a copy of all recorded ReserveJobs calls.
+func (f *FakeAgentServer) GetReserveCalls() [][]string {
+	f.mu.Lock()
+	defer f.mu.Unlock()
+	calls := make([][]string, len(f.ReserveCalls))
+	for i := range f.ReserveCalls {
+		calls[i] = make([]string, len(f.ReserveCalls[i]))
+		copy(calls[i], f.ReserveCalls[i])
+	}
+	return calls
```
DrJosh9000: I'm not convinced these methods (ReserveCallCount, GetReserveCallIDs, GetReserveCalls) are worthwhile. They seem to exist to avoid data races. If you are only checking them after finishing the client call, then they shouldn't be concurrent with any API handling. (You could call `Close` to be sure.) If they are meant to be concurrent with API handling then the return values are unreliable.

api/fake_agent_server.go
```diff
@@ -0,0 +1,176 @@
...
+
+func (f *FakeAgentServer) handleStacksAPI(w http.ResponseWriter, r *http.Request) {
+	// Route based on path
+	if r.URL.Path == "/stacks/test-stack/scheduled-jobs/batch-reserve" {
+		f.handleReserveJobs(w, r)
+		return
+	}
+
+	// Default: not found
+	http.Error(w, "not found: "+r.URL.Path, http.StatusNotFound)
```
DrJosh9000: Why not register this particular path (`"/stacks/test-stack/scheduled-jobs/batch-reserve"`) on the mux, so that you get 404 Not Found everywhere else for free?

api/fake_agent_server_test.go
```diff
@@ -0,0 +1,154 @@
...
+	}
+	if retryAfter != 0 {
+		t.Errorf("ReserveJobs() retryAfter = %v, want 0", retryAfter)
+	}
+	if len(result.Reserved) != 3 {
+		t.Errorf("len(result.Reserved) = %d, want 3", len(result.Reserved))
+	}
+	if len(result.NotReserved) != 0 {
+		t.Errorf("len(result.NotReserved) = %d, want 0", len(result.NotReserved))
+	}
```
DrJosh9000: It would be more thorough to test the entire contents of the response:

```suggestion
	want := &stacksapi.BatchReserveJobsResponse{
		Reserved: jobIDs,
		NotReserved: []string{},
	}
	if diff := cmp.Diff(result, want); diff != "" {
		t.Errorf("ReserveJobs(ctx, %q) diff (-got +want):\n%s", jobIDs, diff) 
	}
```

api/fake_agent_server_test.go
```diff
@@ -0,0 +1,154 @@
...
+		t.Errorf("len(call0) = %d, want 2", len(call0))
+	}
+
+	call1, err := server.GetReserveCallIDs(1)
+	if err != nil {
+		t.Fatalf("GetReserveCallIDs(1) error = %v", err)
+	}
+	if len(call1) != 1 {
+		t.Errorf("len(call1) = %d, want 1", len(call1))
+	}
```
DrJosh9000: Similarly, what about diffing `server.ReserveCalls`?

```suggestion
	want := [][]string{
		{"job-1", "job-2"},
		{"job-3"},
	}
	if diff := cmp.Diff(server.ReserveCalls, want); diff != "" {
		t.Errorf("after ReserveJobs, server.ReserveCalls diff (-got +want):\n%s", diff)
	}
```

api/fake_agent_client.go
zhming0: It's interesting, I don't see this choice as black and white. I believe there is a line somewhere. 

As you noted: 

> Tests that actual API requests are generated and sent
> Tests that actual API responses are received and understood

If a 3rd party API client is trustworthy (like k8s client), the value gain in the areas above is negligible. So it makes the trade-off largely "extra complexity" vs "extra interface", which I think depending on the complexity of test double, can go either way.

api/fake_agent_server.go
```diff
@@ -0,0 +1,145 @@
...
+
+	var buf bytes.Buffer
+	if err := json.NewEncoder(&buf).Encode(resp); err != nil {
+		http.Error(w, "fake: failed to encode response: "+err.Error(), http.StatusInternalServerError)
+		return
+	}
+
+	w.Header().Set("Content-Type", "application/json")
+	w.WriteHeader(f.ReserveStatusCode)
+	_, _ = w.Write(buf.Bytes()) // Write error cannot be handled meaningfully after headers sent
```
DrJosh9000: The error should still be logged.

DrJosh9000: Don't forget to do the same with the other instances above.

api/fake_agent_server.go
```diff
@@ -0,0 +1,145 @@
...
+	var req stacksapi.BatchReserveJobsRequest
+	if err := json.NewDecoder(r.Body).Decode(&req); err != nil {
+		http.Error(w, err.Error(), http.StatusBadRequest)
+		return
+	}
+
+	// Record the call (mutex protects concurrent access from HTTP handler)
+	f.mu.Lock()
+	jobIDsCopy := make([]string, len(req.JobUUIDs))
+	copy(jobIDsCopy, req.JobUUIDs)
```
DrJosh9000: Why does the slice need to be deep cloned? req.JobUUIDs isn't mutated after the request was decoded

api/fake_agent_server_test.go
```diff
@@ -0,0 +1,146 @@
...
+		Endpoint: server.URL(),
+		StackID:  "test-stack",
+		Logger:   slog.Default(),
+	})
+	if err != nil {
+		t.Fatalf("NewAgentClient() error = %v", err)
+	}
+
+	result, _, err := client.ReserveJobs(ctx, []string{"job-1", "job-2", "job-3"})
+
```
DrJosh9000:
```suggestion
```

api/fake_agent_server_test.go
```diff
@@ -0,0 +1,146 @@
...
+	if err != nil {
+		t.Errorf("ReserveJobs() error = %v, want nil", err)
+	}
+	if retryAfter != 0 {
+		t.Errorf("ReserveJobs() retryAfter = %v, want 0", retryAfter)
+	}
+
+	if diff := cmp.Diff(result, want); diff != "" {
+		t.Errorf("ReserveJobs(ctx, %q) diff (-got +want):\n%s", jobIDs, diff)
+	}
```
DrJosh9000, line 46: 1. Don't separate `..., err := ...` from `if err != nil`
2. Include the argument to the function being called (`ReserveJobs`) in output

```suggestion
	result, retryAfter, err := client.ReserveJobs(ctx, jobIDs)
	if err != nil {
		t.Errorf("ReserveJobs(%q) error = %v, want nil", jobIDs, err)
	}
	if retryAfter != 0 {
		t.Errorf("ReserveJobs(%q) retryAfter = %v, want 0", jobIDs, retryAfter)
	}
	
	want := &stacksapi.BatchReserveJobsResponse{
		Reserved:    jobIDs,
		NotReserved: []string{},
	}
	if diff := cmp.Diff(result, want); diff != "" {
		t.Errorf("ReserveJobs(ctx, %q) diff (-got +want):\n%s", jobIDs, diff)
	}
```

DrJosh9000: Don't forget to apply throughout

PR #776, opened 2025-11-26
PB-946 part 2: backfill test for stack notification batcher

zhming0:
Back fill test coverage for notification batcher.

### DIFF
--- a/api/fake_agent_server.go
+++ b/api/fake_agent_server.go
@@ -31,6 +31,9 @@ type FakeAgentServer struct {
 
 	// ReserveError configures an error message to return.
 	ReserveError string
+
+	// NotificationCalls records all notification batches sent to the server.
+	NotificationCalls [][]stacksapi.StackNotification
 }
 
 // NewFakeAgentServer creates and starts a fake agent API server.
@@ -43,6 +46,7 @@ func NewFakeAgentServer() *FakeAgentServer {
 	mux := http.NewServeMux()
 	mux.HandleFunc("/stacks/register", fake.handleRegisterStack)
 	mux.HandleFunc("/stacks/test-stack/scheduled-jobs/batch-reserve", fake.handleReserveJobs)
+	mux.HandleFunc("/stacks/test-stack/notifications", fake.handleNotifications)
 
 	fake.server = httptest.NewServer(mux)
 	return fake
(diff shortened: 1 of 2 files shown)


api/fake_agent_server.go
```diff
@@ -146,3 +150,33 @@ func (f *FakeAgentServer) handleReserveJobs(w http.ResponseWriter, r *http.Reque
...
+	var req stacksapi.CreateStackNotificationsRequest
+	if err := json.NewDecoder(r.Body).Decode(&req); err != nil {
+		http.Error(w, err.Error(), http.StatusBadRequest)
+		return
+	}
+
+	f.mu.Lock()
+	batch := make([]stacksapi.StackNotification, len(req.Notifications))
+	copy(batch, req.Notifications)
+	f.NotificationCalls = append(f.NotificationCalls, batch)
```
DrJosh9000: Same comment as last PR - I don't see a point in carefully copying the slice (that was created a few lines above with the JSON decode), since nothing else will be mutating it.

api/stack_notification_batcher_test.go
```diff
@@ -0,0 +1,120 @@
...
+		if err := nb.add(ctx, stacksapi.StackNotification{
+			JobUUID: "job-" + string(rune('A'+i%26)),
+			Detail:  "detail",
+		}); err != nil {
+			t.Fatalf("add() error = %v", err)
+		}
+	}
+
+	// Wait for ticker to flush (interval is 100ms)
+	time.Sleep(250 * time.Millisecond)
```
DrJosh9000, line 97: Do we have a better mechanism for ensuring the notifications have flushed?

Could we take advantage of https://pkg.go.dev/testing/synctest ?

api/stack_notification_batcher_test.go
```diff
@@ -0,0 +1,120 @@
...
+		t.Fatalf("start() error = %v", err)
+	}
+
+	// Add more than maxNotificationsPerBatch (1000) notifications
+	for i := range 2500 {
+		if err := nb.add(ctx, stacksapi.StackNotification{
+			JobUUID: "job-" + string(rune('A'+i%26)),
+			Detail:  "detail",
+		}); err != nil {
+			t.Fatalf("add() error = %v", err)
```
DrJosh9000: If this fails, it would be good to know how far through the 2500 it got and what the notification object actually was:

```suggestion
		note := stacksapi.StackNotification{
			JobUUID: "job-" + string(rune('A'+i%26)),
			Detail:  "detail",
		}
		if err := nb.add(ctx, note); err != nil {
			t.Fatalf("i=%d add(%v) error = %v", i, note, err)
```

api/stack_notification_batcher_test.go
zhming0: TIL synctest, I gave it a good play. I don't think it works in our use. The basic premise of synctest is like poorman's DST, within `synctest.Test`'s when all goroutines are "durably blocked", time flies fast, it fast track to a time where at least one goroutine can be blocked, and then fast forward to next etc.  As if a single threaded cooperative concurrency model.

The test goroutine can also enter durably blocked state by issuing `time.Sleep` or `synctest.Wait`. 

Sadly "durably blocked" does not include: 
> System calls like file or network I/O
> External event handling (such as reading from a socket)

This is precisely what the agent faker server is doing. 

As a result, our fake agent server (blocked on the real IO connection accept) and client (blocked on readLoop/writeLoop) will never enter the "durably blocked" state, meaning the entire synctest block will be blocked. (Remember, it's single threaded, one running routine can cause the entire thread to block).

I think it's in theory possible to instrument fake agent server and client so it can become "durably blocked" during test, but I don't think the benefits justify the effort yet.

api/stack_notification_batcher_test.go
```diff
@@ -0,0 +1,121 @@
...
+	defer cancel()
+
+	if err := nb.start(ctx); err != nil {
+		t.Fatalf("start() error = %v", err)
+	}
+
+	// Add more than maxNotificationsPerBatch (1000) notifications
+	for i := range 2500 {
+		note := stacksapi.StackNotification{
+			JobUUID: "job-" + string(rune('A'+i%26)),
```
DrJosh9000: What about:
 
```suggestion
			JobUUID: fmt.Sprintf("job-%d", i),
```

PR #791, opened 2025-12-23
Fix multiple calls to t.Parallel in cleanup test

phantomwhale:
t.Parallel() was being called multiple times on the same parent test t. The testcase.Init() method calls t.Parallel(), and it was being called in a loop with the parent test, causing a panic to occur.

This moves the testcase initialization inside the t.Run callback so that Init() (and thus t.Parallel()) is called on each subtest's *testing.T instead of the parent test.

This also removes the unused sync.WaitGroup variable and import.

### DIFF
--- a/charts/agent-stack-k8s/values.schema.json
+++ b/charts/agent-stack-k8s/values.schema.json
@@ -85,7 +85,7 @@
       "default": [],
       "title": "K8s EnvVar environment variables to add to the agent-stack-k8s controller container",
       "items": {
-        "$ref": "https://kubernetesjsonschema.dev/master/_definitions.json#/definitions/io.k8s.api.core.v1.EnvVar"
+        "$ref": "https://raw.githubusercontent.com/yannh/kubernetes-json-schema/master/v1.35.0/_definitions.json#/definitions/io.k8s.api.core.v1.EnvVar"
       }
     },
     "nodeSelector": {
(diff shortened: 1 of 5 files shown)


phantomwhale:
This SEEMED to fix it up; it tidied up 100 pipelines when it ran (https://buildkite.com/buildkite-kubernetes-stack/kubernetes-agent-stack/builds/3418/steps/canvas?sid=019b497f-5377-4bf9-a82f-e7535c5bee37)

But my Golang is weak, and perhaps I'm misreading the code (should it have cleaned up the other 350 pipelines? What about the 500 queues?)

CleanShot 2025-12-23 at 15 48 13

internal/integration/cleanup_queues_test.go
```diff
@@ -0,0 +1,145 @@
...
+	"context"
+	"strings"
+	"testing"
+
+	"github.com/Khan/genqlient/graphql"
+	"github.com/buildkite/agent-stack-k8s/v2/internal/integration/api"
+	"github.com/stretchr/testify/require"
+)
+
+func TestCleanupOrphanedQueues(t *testing.T) {
```
moskyb, line 13: So my PR (#792) added similar functionality to the `CleanupOrphanedResources` test, because we can't delete queues that have jobs in them, so there's a light dependency between cancelling jobs (which happens as part of pipeline cleanup) and deleting queues.

because run order is randomised, this means that sometimes this "test" will run in parallel with `TestCleanupOrphanedResources`, which might lead to contention. i think it's best that we keep all of the cleanup logic in one test so that we can control order more easily.

PR #792, opened 2025-12-23
Cleanup cleanup

moskyb:
The cleanup orphaned pipelines test (that's not really a test, more a maintenance task we need to run every once in a while) needs some serious TLC. It uses the testcase struct as a concurrency controller, and its logic is generally really confusing.

This PR updates the cleanup test to use more standard go idioms for concurrency control, generally makes it a bit easier to read, and updates it so that it also deletes orphaned cluster queues.

It also updates `go-buildkite` to the latest version -- we were fairly seriously behind

### DIFF
--- a/.buildkite/pipeline.yaml
+++ b/.buildkite/pipeline.yaml
@@ -13,6 +13,22 @@ x-anchors:
           checkout:
             fetchFlags: -v --tags
 
+  run-tests: &run-tests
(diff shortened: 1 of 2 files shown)


phantomwhale:
Love it!

phantomwhale:
Looks like it's failing to clean up queues in the tests?

PR #795, opened 2025-12-29
Update Kubernetes schema references

JoeColeman95:
Replaced references to kubernetesjsonschema.dev with direct links to the v1.35.0 schema definitions from yannh/kubernetes-json-schema.

As of the time of this PR, kubernetesjsonschema.dev is down and has been for over a day, with it looking like it's not coming back, or not soon at least.

This will prevent schema validation erroring out with a 404 when specified, or when there's a tool used such as Terraform to deploy, which is pretty standard.

We were previously pinning to `master`, which would could also be breaking. To avoid schema validation breaking in future for older releases, I've opted to pin to `v1.35.0`, which is the latest version currently anyway.

This adds maintenance burden with the benefit of stability for past releases. Feel free to flag this if you would prefer `master/master` over `master/v1.35.0`.

I've opted for the source of [yannh/kubernetes-json-schema](https://github.com/yannh/kubernetes-json-schema) as this seems to be the most up-to-date and maintained source, with it being highly used throughout the K8's community. Again, if you have any issues, just flag which source you'd prefer and I can change it.

### DIFF
--- a/charts/agent-stack-k8s/values.schema.json
+++ b/charts/agent-stack-k8s/values.schema.json
@@ -85,7 +85,7 @@
       "default": [],
       "title": "K8s EnvVar environment variables to add to the agent-stack-k8s controller container",
       "items": {
-        "$ref": "https://kubernetesjsonschema.dev/master/_definitions.json#/definitions/io.k8s.api.core.v1.EnvVar"
+        "$ref": "https://raw.githubusercontent.com/yannh/kubernetes-json-schema/master/v1.35.0/_definitions.json#/definitions/io.k8s.api.core.v1.EnvVar"
       }
     },
     "nodeSelector": {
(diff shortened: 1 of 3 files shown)


moskyb (approved):
This is breaking certain helm installs, including all of those using terraform, as the helm terraform provider doesn't provide an option to skip json schema verification. For this reason, i'm going to merge this in and ship a patch release, potentially backporting to a previous version or two. 

i do want to note that we're essentially moving from importing our schema definitions from some person's website to importing them from some other person's website. there's every chance that a similar thing happens to us again, where the owner of the new json schema repository yanks their repo, or makes breaking changes, or something similarly bad for us as software maintainers.

in the interest of getting fixes out promptly (ideally in time for most people to returning to work after the holidays), i think we should merge this, but in the short-to-medium term, i'd love to move toward hosting the json schema repository ourselves — be that through a fork, or perhaps some higher-tech thing that we can figure out later.

PR #798, opened 2026-01-07
Add default-resource-class-name config option

gempesaw:
## Background

We really like using `resource_classes` from https://github.com/buildkite/agent-stack-k8s/pull/668 to be able to vary CPU/memory requests based off a single agent template, but currently if a pipeline step doesn't specify a `resource_class: foo`, they don't get any resource requests at all. I was hoping perhaps specifying resources in `pod-spec-patch` would be the fallback/default in that case, but pod-spec-patch's resources override anything from `resource_class`:

applyResourceClass happens around 524:

https://github.com/buildkite/agent-stack-k8s/blob/4beedd0532263961a398be5634eefd539cbabfb4/internal/controller/scheduler/scheduler.go#L524-L527

and then later on the pod-spec-patch and plugin spec patches are applied on top:

https://github.com/buildkite/agent-stack-k8s/blob/4beedd0532263961a398be5634eefd539cbabfb4/internal/controller/scheduler/scheduler.go#L739-L742

https://github.com/buildkite/agent-stack-k8s/blob/4beedd0532263961a398be5634eefd539cbabfb4/internal/controller/scheduler/scheduler.go#L753-L755

it'd be nice to have a default that applies when a step doesn't remember to specify a `resource_class`

## Changes

Adds a new `default-resource-class-name` config option that references a named resource class from `resource-classes`. When a job doesn't have a `resource_class` agent tag, the default kicks in; jobs that explicitly set `resource_class` still take priority.

usage looks like:

```yaml
resource-classes:
  small:
    resource:
      requests:
        cpu: "500m"
        memory: "512Mi"
  large:
    resource:
      requests:
        cpu: "2"
        memory: "4Gi"

default-resource-class-name: "small"
```

I also added `resource-classes` to the helm schema since it seemed to be missing?

## Testing

- Unit tests for `applyResourceClass` covering the various combinations (no tag/no default, tag specified, default fallback, tag overrides default, error cases)
- Integration test that verifies the default resource class actually gets applied by checking cgroup limits

### DIFF
--- a/cmd/controller/controller.go
+++ b/cmd/controller/controller.go
@@ -167,6 +167,15 @@ func ParseAndValidateConfig(v *viper.Viper) (*config.Config, error) {
 		}
 	}
 
+	if cfg.DefaultResourceClassName != "" {
+		if cfg.ResourceClasses == nil {
+			return nil, fmt.Errorf("default-resource-class-name %q specified but no resource-classes defined", cfg.DefaultResourceClassName)
+		}
+		if _, exists := cfg.ResourceClasses[cfg.DefaultResourceClassName]; !exists {
+			return nil, fmt.Errorf("default-resource-class-name %q not found in resource-classes", cfg.DefaultResourceClassName)
+		}
+	}
+
 	if _, err := resource.ParseQuantity(cfg.ImageCheckContainerCPULimit); err != nil {
 		return nil, fmt.Errorf("invalid CPU resource limit defined: %s", cfg.ImageCheckContainerCPULimit)
 	}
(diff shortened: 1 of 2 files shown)


gempesaw:
@moskyb 

> adding a validation in [controller.ParseAndValidateconfig](https://github.com/buildkite/agent-stack-k8s/blob/main/cmd/controller/controller.go#L136) to ensure that the default resource class is included in the list of resource classes provided

oo yea that's a great idea, tyvm b6087f9cd788a7f7cc59cd8afa892da6354f2e39

PR #799, opened 2026-01-08
PS-1530: Viper -> Kong, rework controller entrypoint

### LINKED ISSUE #797: [BUG] CSI volume attributes are incorrectly lowercased
## Describe the bug
When providing a `pod-spec-patch` with a CSI volume such as:
```
pod-spec-patch:
  ...
  volumes:
    - csi:
        driver: secrets-store-gke.csi.k8s.io
        readOnly: true
        volumeAttributes:
          secretProviderClass: buildkite-job-default
      name: secrets
```

The job is incorrectly created with a volume with lowercased attributes as follows:
```
...
spec:
  spec:
    ...
    volumes:  
      - csi:
          driver: secrets-store-gke.csi.k8s.io
          readOnly: true
          volumeAttributes:
            secretproviderclass: buildkite-job-default
        name: secrets
```

## To Reproduce
Create agent stack with a pod-spec-patch including a CSI driver volume with camel case volume attributes and run a job. The agent config map is created correctly but the resulting job has lowercased volume attribute keys.

## Expected behavior
CSI volume attribute casing should remain as provided.

## Environment
- agent-stack-k8s version: 0.36.2
- Kubernetes version: 1.34.1
- Deployment method: helm chart

## Logs
n/a

## Additional context
I imagine this relates to the viper lowercasing issue addressed for resource requests in https://github.com/buildkite/agent-stack-k8s/pull/738

zhming0:
Fixes #797. 

Viper has a known limitation, it [does not support case sensitive keys in config file](https://github.com/spf13/viper?tab=readme-ov-file#does-viper-support-case-sensitive-keys). 

In the past when we encounter such issue, we had a number of bandaid solution: #732, #738.  This is clearly not sustainable as K8s spec is fundamentally case sensitive. 

This PR replaced Viper with Kong + k8s native unmarshaller, solving this problem categorically, and at the same time, getting rid of many viper specific patches from the past. 

The diff of this PR is unfortunately larger than what I wanted. Despite we have test coverage on various config code path, running this controller locally checking things will be better approach. 

Hopefully it will be worth it


cmd/controller/flags.go
```diff
@@ -1,187 +1,486 @@
...
+	// Logging settings
+	LogFormat string `kong:"name='log-format',help='Log format: logfmt or json (default: logfmt)',env='LOG_FORMAT'"`
+	NoColor   *bool  `kong:"name='no-color',help='Disable colored log output',env='NO_COLOR'"`
+	LogLevel  string `kong:"name='log-level',help='Log level: debug, info, warn, error (default: info)',env='LOG_LEVEL'"`
+}
+
+// applyCLIOverrides applies CLI values to the config when explicitly set.
+// Pointer fields allow distinguishing "not set" (nil) from "set to zero/false".
+// This ensures precedence: defaults < config file < env vars/CLI flags
+func applyCLIOverrides(cfg *config.Config, cli *CLI) {
```
moskyb: this is an important function, but i think the repetitiveness makes it harder to read. it feels like it'd be really easy to miss adding something to this function, or to misunderstand its usage at a glance.

it feels like one of the few cases where something might actually be more readable if it used reflection, rather than less. it might be worth giving it a shot?

cmd/controller/flags.go
```diff
@@ -1,187 +1,486 @@
...
+	if file.Namespace != "" {
+		base.Namespace = file.Namespace
+	}
+	if file.Debug {
+		base.Debug = file.Debug
+	}
+	if file.MaxInFlight != 0 {
+		base.MaxInFlight = file.MaxInFlight
+	}
+	if file.JobTTL != 0 {
```
trvrnrth: There's a subtle change in behaviour whereby It's no longer possible to set a job TTL of "0s" which I suspect is due to this check.

My use case is that I want the job pods to be deleted ASAP as I'm using https://github.com/seatgeek/buildkit-operator to manage per-step buildkit instances and it uses a webhook on deletion to manage cleanup.

For now a TTL of 1s is good enough but it would probably be good to retain the old behaviour if possible.

zhming0: Great catch. Thank you! I refactored how things are done, added more test harness. The old behavior is retained now. 

With regard to `buildkit-operator`, I am curious, do you provision ephemeral `Buildkit` instance? I would've thought non-ephemeral setup are more desirable for most use cases due to better caching etc.

internal/integration/main_test.go
```diff
@@ -32,14 +31,24 @@ var (
...
+			if i+1 < len(os.Args) {
+				configArgs = append(configArgs, "--config="+os.Args[i+1])
+			}
+		}
+		if strings.HasPrefix(arg, "-f=") {
+			configArgs = append(configArgs, "--config="+strings.TrimPrefix(arg, "-f="))
+		}
+		if strings.HasPrefix(arg, "--config=") {
+			configArgs = append(configArgs, arg)
+		}
```
zhming0, line 48: I am not sure if we ever use --config to impact our test environment config, maybe I can remove this whole bit. But it doesn't have much harm keeping it either.

cmd/controller/flags.go
```diff
@@ -1,187 +1,299 @@
...
+// "not set" (nil) from "explicitly set to zero/false". This allows CLI/env vars
+// to override config file values with zero/false values.
+// Defaults are applied via newConfigWithDefaults().
+type CLI struct {
+	ConfigFile string `kong:"name='config',short='f',help='Path to config file',env='CONFIG'"`
+
+	Debug *bool `kong:"name='debug',help='Sets log level to debug',env='DEBUG'"`
+
+	// Job / Pod settings
+	JobTTL                               *time.Duration `kong:"name='job-ttl',help='Time to retain kubernetes jobs after completion (default: 10m)',env='JOB_TTL'"`
```
moskyb: were these envars previously namespaced into `BUILDKITE_KUBERNETES_STACK_*` or something? if so, will this change change that?

cmd/controller/flags.go
```diff
@@ -1,187 +1,299 @@
...
+			if !cliField.IsNil() {
+				cfgField.Set(cliField.Elem())
+			}
+		case reflect.String:
+			// String types (empty = not set)
+			if cliField.String() != "" {
+				cfgField.Set(cliField)
+			}
+		case reflect.Slice:
+			// Slice types (empty = not set)
```
moskyb: > empty = not set

i just want to double check that this is the right semantic to check — in the agent, there are slice fields with defaults (eg vars to redact) that have a sensible set of default values, but that can manually be set to `[]` if the user desires. Using array emptiness as the metric for a field not being set essentially gates us out of having behaviour like that.

i think if we're making that decision then that's all good, but i also think it's probably good to think about whether checking nil-ness might give us better semantics.

cmd/controller/flags.go
```diff
@@ -1,187 +1,299 @@
...
+			if cliField.Len() > 0 {
+				cfgField.Set(cliField)
+			}
+		}
+	}
+}
+
+// newConfigWithDefaults returns a config with default values.
+func newConfigWithDefaults() *config.Config {
```
moskyb, line 133: it's probably worth calling out that kong does have a `default` directive in its struct tag schema, but i actually think that this is significantly more readable, so no action necessary here.

cmd/controller/flags.go
```diff
@@ -1,187 +1,299 @@
...
+	// First pass: get the set of keys present in the file
+	var rawMap map[string]any
+	if err := yaml.Unmarshal(data, &rawMap); err != nil {
+		return nil, nil, fmt.Errorf("failed to parse config file: %w", err)
+	}
+
+	keys := make(map[string]bool)
+	for k := range rawMap {
+		keys[k] = true
+	}
```
moskyb: nit: use `map[string]struct{}` instead of `map[string]bool` for sets, as the semantics of whether something is present is clearer — ie, if we have an entry `"some_string": false` in the map, what does that mean? using `struct{}` makes this state unrepresentable

cmd/controller/flags.go
```diff
@@ -1,187 +1,299 @@
...
+
+	decoder := json.NewDecoder(strings.NewReader(string(jsonData)))
+	decoder.DisallowUnknownFields()
+	return decoder.Decode(cfg)
+}
+
+// convertDurations recursively converts duration strings (e.g., "5m", "30s") to nanoseconds.
+func convertDurations(m map[string]any) {
+	durationFields := map[string]bool{
```
moskyb: as above re: `map[string]bool` -> `map[string]struct{}`

cmd/controller/flags.go
```diff
@@ -1,187 +1,299 @@
...
+	}
+
+	convertDurations(rawMap)
+
+	jsonData, err := json.Marshal(rawMap)
+	if err != nil {
+		return err
+	}
+
+	decoder := json.NewDecoder(strings.NewReader(string(jsonData)))
```
moskyb:
```suggestion
	decoder := json.NewDecoder(bytes.NewBuffer(jsonData))
```
to avoid a needless cast to `string`

cmd/controller/flags.go
```diff
@@ -1,187 +1,299 @@
...
+// mergeConfigFromFile merges file config values into the base config.
+// The keys map indicates which fields were present in the file, allowing
+// zero values (like max-in-flight: 0) to override defaults.
+// Uses reflection to iterate over fields based on their JSON tags.
+func mergeConfigFromFile(base, file *config.Config, keys map[string]bool) {
+	baseVal := reflect.ValueOf(base).Elem()
+	fileVal := reflect.ValueOf(file).Elem()
+	baseType := baseVal.Type()
+
+	for i := 0; i < baseType.NumField(); i++ {
```
moskyb:
```suggestion
	for i := range baseType.NumField() {
```

moskyb (commented):
very much LGTM in concept, viper really annoys me. i'm sure Kong will grow to annoy me in different ways, but for now it's very exciting.

i have a couple of notes on the config parsing logic, but nothing major.

cmd/controller/flags.go
zhming0: We don't use env var prefix on agent-k8s-stack, I think we should, but this isn't something we can/should change at this point. So no behavior change 👍🏿

cmd/controller/flags.go
zhming0: Yeah I thought about using this `default` directive. But decided against that for the reason you mentioned. 

Also, after this refactor, unlike viper/cobra, the kong bit is only a part of the config loading process, the config file parsing and CLI + EnvVar parsing are separated. So it no long make sense to use `default` directive in struct tag scheme.

cmd/controller/flags.go
moskyb: sweet, good to know

cmd/controller/flags.go
zhming0: This is a great catch! I've changed to checking `IsNil` to determine if it's unset, and added a test case. This would match the old viper behavior and I also think it's generally better 👍🏿 .

moskyb (approved):
LGTM!

PR #804, opened 2026-01-15
PB-1062: dogfooding resource class

zhming0:
We should use this feature 😁 . 
There may be a subsequent PR to add resource_class agent tag in pipeline.yml


moskyb (approved):
🤙

PR #807, opened 2026-01-22
PB-1069: vendor k8s json schema

zhming0:
Instead relying on 3rd party url during schema validation, vendor the schema locally. This mitigate a supply chain risk.


charts/agent-stack-k8s/schemas/extract-k8s-schema.py
```diff
@@ -0,0 +1,122 @@
+#!/usr/bin/env python3
```
moskyb: what's the origin of this script? is it from yannh's repo, or is it something we wrote?

if it's from yannh's repo, we should probably call that out.

zhming0: It's vomited out from a LLM 😁 (Opus 4.5)

zhming0: I've removed it.

charts/agent-stack-k8s/schemas/update-k8s-schema.sh
```diff
@@ -0,0 +1,23 @@
...
+#   ./update-k8s-schema.sh 1.36.0
+#
+set -euo pipefail
+
+VERSION="${1:-1.35.0}"
+URL="https://raw.githubusercontent.com/yannh/kubernetes-json-schema/master/v${VERSION}/_definitions.json"
+OUTPUT="$(dirname "$0")/kubernetes-v${VERSION}.json"
+
+echo "Downloading Kubernetes schema v${VERSION}..."
+curl -fsSL "$URL" -o "$OUTPUT"
```
DrJosh9000:
```suggestion
curl -fsSL "${URL}" -o "${OUTPUT}"
```

charts/agent-stack-k8s/schemas/update-k8s-schema.sh
```diff
@@ -0,0 +1,23 @@
...
+
+echo "Written to $OUTPUT"
+echo "File size: $(du -h "$OUTPUT" | cut -f1)"
```
DrJosh9000:
```suggestion
echo "Written to ${OUTPUT}"
echo "File size: $(du -h "${OUTPUT}" | cut -f1)"
```

charts/agent-stack-k8s/schemas/update-k8s-schema.sh
zhming0: I am curious what pitfall this is preventing, if it's useful I am keen to add it to my personal shellcheck rule

DrJosh9000: I am used to https://google.github.io/styleguide/shellguide.html#s5.6-variable-expansion , but what you have here is probably fine (there is no confusion over what variable is referred to in the variable expansion based on what is after).

Issue #808, opened 2026-01-22
[BUG] agent-stack-k8s does not work with ubuntu based agent images


seemethere:
## Describe the bug
## agent-stack-k8s Alpine Hardcoding Issue

Version: v0.37.0 (current latest)

Location: `internal/controller/scheduler/scheduler.go`

### Problem

The controller hardcodes Alpine-specific shell and commands, making it impossible to use Ubuntu/Debian-based agent images:

1. Shell hardcoded to ash (Alpine shell):
```go
// Line 987 - copy-agent init container
Command: []string{"ash"}
Args:    []string{"-cefx", containerArgs.String()}

// Lines 1111, 1129, 1142 - checkout container
Command: []string{"ash", "-c"}
```
2. Alpine-specific user/group commands:
```go
// Lines 1112-1115
checkoutContainer.Args = []string{fmt.Sprintf(`set -exufo pipefail
addgroup -g %d buildkite-agent
adduser -D -u %d -G buildkite-agent -h /workspace buildkite-agent
su buildkite-agent -c "%s && buildkite-agent-entrypoint kubernetes-bootstrap"`,
```
- addgroup / adduser -D are BusyBox/Alpine commands
- Ubuntu/Debian use groupadd / useradd

### Why This Matters

Some Kubernetes environments require the use-vc resolv.conf option to force TCP-based DNS queries. musl libc (Alpine) doesn't support use-vc, causing DNS resolution to fail. glibc-based images (Ubuntu, Rocky) work correctly. In general I feel as though it'd be good for all images published by `buildkite/agent` to be compatible with this stack

### Requested Enhancement

Add configuration option to specify shell and use POSIX-compatible user creation, or detect the image type and adapt accordingly. Example:
```
config:
  shell: "/bin/bash"  # or auto-detect
```
Or use POSIX-compatible approach that works on both:
```
# Instead of Alpine-specific adduser/addgroup
getent group buildkite-agent || groupadd -g $GID buildkite-agent
getent passwd buildkite-agent || useradd -u $UID -g buildkite-agent -d /workspace buildkite-agent
```

## To Reproduce
1. Deploy with configuration '...':
```yaml
  # Helm values for agent-stack-k8s
  config:
    # Custom agent image (Ubuntu-based instead of default Alpine)
    image: "ghcr.io/buildkite/agent:3.115.4-ubuntu-24.04"

    # Required for our environment - forces TCP DNS queries
    pod-spec-patch:
[...]

PR #811, opened 2026-01-27
PS-1583: Replace Alpine-specific shell commands with portable alternatives

zhming0:
This PR builds on top of the awesome #809(@seemethere). Solve #808. 

The goal remain the same: 
> Use /bin/sh instead of ash and add auto-detection for user/group creation commands to support both Alpine (BusyBox) and glibc-based images (Ubuntu, Debian, Rocky Linux). This enables use of non-Alpine agent images which may be needed for glibc-specific features like TCP-based DNS queries (use-vc resolv.conf option).

I fixed a number of paper cuts issue in #809: 
- improved naming. 
- added an integration test.
- increase readability on the script constructing code path, unify some duplicated condition. 
- fix the incorrect tinit-static path.

### DIFF
--- a/internal/controller/scheduler/scheduler.go
+++ b/internal/controller/scheduler/scheduler.go
@@ -919,36 +919,6 @@ func PatchPodSpec(original *corev1.PodSpec, patch *corev1.PodSpec, cmdParams *co
 	return &patchedSpec, nil
 }
 
-// generatePortableUserScript generates a shell script that creates a user
-// and optionally a group, working on both Alpine (BusyBox) and glibc-based
-// systems (Debian, Ubuntu, RHEL).
-func generatePortableUserScript(uid, gid int64, username, groupname string) string {
-	var sb strings.Builder
-	sb.WriteString("# Portable user/group creation\n")
-	sb.WriteString("if command -v adduser >/dev/null 2>&1 && adduser --help 2>&1 | grep -q BusyBox; then\n")
-
-	// BusyBox (Alpine) branch
-	if groupname != "" && groupname != "root" {
-		fmt.Fprintf(&sb, "    addgroup -g %d %s 2>/dev/null || true\n", gid, groupname)
-	}
-	fmt.Fprintf(&sb, "    adduser -D -u %d -G %s -h /workspace %s\n", uid, groupname, username)
-
-	sb.WriteString("elif command -v useradd >/dev/null 2>&1; then\n")
-
-	// shadow-utils (Debian/Ubuntu/RHEL) branch
-	if groupname != "" && groupname != "root" {
-		fmt.Fprintf(&sb, "    groupadd -g %d %s 2>/dev/null || true\n", gid, groupname)
-	}
-	fmt.Fprintf(&sb, "    useradd -M -u %d -g %s -d /workspace -s /bin/sh %s\n", uid, groupname, username)
-
-	sb.WriteString("else\n")
-	sb.WriteString(`    echo "Error: No supported user creation tool found" >&2` + "\n")
-	sb.WriteString("    exit 1\n")
-	sb.WriteString("fi\n")
-
-	return sb.String()
-}
-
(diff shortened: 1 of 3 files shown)


internal/controller/scheduler/shell.go
```diff
@@ -0,0 +1,36 @@
...
+
+import (
+	"fmt"
+	"strings"
+)
+
+// generateCreateUserScript generates a shell script that creates a user
+// and optionally a group, working on both Alpine (BusyBox) and glibc-based
+// systems (Debian, Ubuntu, RHEL).
+func generateCreateUserScript(uid, gid int64, username, groupname string) string {
```
moskyb, line 28: this might be easier to read as a template rather than a string builder, switching from `sb.WriteString` to `fmt.Fprintf` and the various quote changes makes it hard to get into the flow of reading this script IMO

```Go
var createUserScriptTmpl = template.Must(template.New("createUser").Parse(`# Portable user/group creation
if command -v adduser >/dev/null 2>&1 && adduser --help 2>&1 | grep -q BusyBox; then
{{- if .CreateGroup }}
    addgroup -g {{.GID}} {{.Groupname}} 2>/dev/null || true
{{- end }}
    adduser -D -u {{.UID}} -G {{.Groupname}} -h /workspace {{.Username}}
elif command -v useradd >/dev/null 2>&1; then
{{- if .CreateGroup }}
    groupadd -g {{.GID}} {{.Groupname}} 2>/dev/null || true
{{- end }}
    useradd -M -u {{.UID}} -g {{.Groupname}} -d /workspace -s /bin/sh {{.Username}}
else
    echo "Error: No supported user creation tool found" >&2
    exit 1
fi
`))

// generateCreateUserScript generates a shell script that creates a user
// and optionally a group, working on both Alpine (BusyBox) and glibc-based
// systems (Debian, Ubuntu, RHEL).
func generateCreateUserScript(uid, gid int64, username, groupname string) string {
	data := struct {
		UID         int64
		GID         int64
		Username    string
		Groupname   string
		CreateGroup bool
	}{
		UID:         uid,
		GID:         gid,
		Username:    username,
		Groupname:   groupname,
		CreateGroup: groupname != "" && groupname != "root",
	}

	var buf bytes.Buffer
	_ = createUserScriptTmpl.Execute(&buf, data)
	return buf.String()
}
```

internal/controller/scheduler/scheduler.go
```diff
@@ -1101,50 +1111,36 @@ func (w *worker) createCheckoutContainer(
...
-		checkoutContainer.Command = []string{"ash", "-c"}
-		checkoutContainer.Args = []string{fmt.Sprintf(`set -exufo pipefail
-adduser -D -u %d -G root -h /workspace buildkite-agent
-su buildkite-agent -c "%s && buildkite-agent-entrypoint kubernetes-bootstrap"`,
-			podUser,
-			gitConfigCmd,
-		)}
+		createUserScript := generateCreateUserScript(podUser, gid, "buildkite-agent", groupname)
+		bootstrapCmd := fmt.Sprintf(`su buildkite-agent -c "%s && buildkite-agent-entrypoint kubernetes-bootstrap"`, gitConfigCmd)
+		checkoutScript := strings.Join([]string{"set -exuf", createUserScript, bootstrapCmd}, "\n")
```
moskyb, line 1130: worth noting that we're losing `-o pipefail` here — is that what we want?

moskyb (approved):

internal/controller/scheduler/scheduler.go
zhming0: Yeah good catch. I tried to get a portable `set -o pipefail` but failed, because for example in `dash`, `set -o pipefail` will end the shell even though I tried to capture it `set -o pipefail || true`. 
 
I think it's fine. Our own script here don't use pipe syntax. It's arguable though, whether `agent` should handle the pipefail? But I think it's out of scope.

internal/controller/scheduler/shell.go
zhming0: good idea!

PR #822, opened 2026-02-10
Add HTTP timeout and logger for Stacks API client

adnowakodbx1:
In our environment we noticed cases where controller seemed to hang until it was manually restarted. I noticed that stacksapi client is using `http.DefaultClient` which has no timeout and no logging of requests. Wondering whether this is on purpose? Because it seems the `AgentClient` is  using the timeout and request logging.

Also wondering whether maybe these timeouts should be configurable? Happy to modify this if you deem it might be useful.


adnowakodbx1:
If it looks reasonable please feel free to merge the PR yourself. I would also appreciate if you could create release including this change since it's quite important for us.

PR #823, opened 2026-02-12
HTTP Timeouts

moskyb:
This PR adds HTTP timeouts to the stacks api client, and also makes the default HTTP timeout configurable. By default, it'll stay at the old (previously unconfigurable) value of 60 seconds.

Many thanks to @adnowakodbx1 for suggesting this.


api/agent_client.go
zhming0, line 89: are we supposed to change here to apply the timeout setting? 🤔

moskyb: oop! yes 😶

moskyb: fixed

PR #824, opened 2026-02-13
Better request logging

moskyb:
Preivously, the k8s stack allowed users to log request payloads when the `DEBUG` environment variable was set, though this didn't affect calls to the stacks API (which represents the vast majority of the requests made by the k8s stack). The logs were also not formatted in the same way as the existing logs, using the `log` rather than the `slog` library.

This PR makes a variety of changes to this system, primarily:
- Removing the `DEBUG` flag check, and replacing it with a dedicated flag `--log-http-payloads`
- When this flag is enabled, requests and their payloads will be logged for calls to the stacks API, as well as any other API calls.
- Updated the logging roundtripper to use `slog` rather than `log`

To do this, i've done a little bit of surgery and cleanup in `AgentTokenClient` — previously, it was authenticating via an authenticating roundtripper, which was overkill for a struct with one method. I removed this authenticating roundtripper.


api/log_transport.go
```diff
@@ -0,0 +1,82 @@
...
+		"duration", duration,
+	)
+
+	if t.logPayloads {
+		b, err := httputil.DumpResponse(out, true)
+		if err != nil {
+			logger.Warn("Failed to dump response", "error", err)
+		}
+
+		logger.With("response_body", string(b))
```
blaknite: Blocking: Looks like the `logger.With(...)` result needs to be assigned back to `logger` here, same as the request side on line 48.

```suggestion
		logger = logger.With("response_body", string(b))
```

moskyb: ah i fixed one of these earlier but missed this one! curses

PR #855, opened 2026-04-07
Allow safe-to-evict annotation to be overridden

ceciliaschudel:
The `cluster-autoscaler.kubernetes.io/safe-to-evict` annotation is currently hardcoded to "false", preventing it from being overridden in Helm values. This causes jobs targeting spot nodes to never start in a GKE Autopilot cluster, because the cluster autoscaler refuses to scale up spot nodes for pods that disallow eviction.

Moves the annotation before the metadata merges so it acts as a default that can be overridden via DefaultMetadata or k8s plugin annotations.

Fixes https://github.com/buildkite/agent-stack-k8s/issues/849

### DIFF
--- a/internal/controller/scheduler/scheduler.go
+++ b/internal/controller/scheduler/scheduler.go
@@ -299,6 +299,10 @@ func (w *worker) Build(podSpec *corev1.PodSpec, skipCheckout bool, inputs buildI
 		},
 	}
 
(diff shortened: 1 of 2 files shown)


zhming0 (approved):
@ceciliaschudel thank you very much for the contribution! LGTM!

PR #860, opened 2026-04-14
A-1099: allow container start timeout to be configured

zhming0:
This PR is a follow up to https://github.com/buildkite/agent/pull/3822. 

It allow user to pass kubernetes container start timeout to agent, most importantly, this PR adds test coverage for this change. 

as part of this change we also upgrade Agents to v3.123.0


charts/agent-stack-k8s/values.schema.json
```diff
@@ -441,6 +441,10 @@
             "debug-signing": {
               "type": "boolean"
             },
+            "kubernetes-container-start-timeout": {
```
DrJosh9000: Perhaps the config should be called "container-start-timeout" ? Mainly because everything here is Kubernetes, so I don't think "kubernetes-" is adding any value.

```suggestion
            "container-start-timeout": {
```

zhming0: It makes sense 👍🏿

DrJosh9000 (approved):
LGTM

PR #862, opened 2026-04-17
A-1136: stop detecting init container failure for sidecar containers

### LINKED ISSUE #827: [BUG] failOnInitContainerFailure retries indefinitely on 404 API::Error::NotFound, causing zombie job/pod accumulation
## Describe the bug

When an init container fails, `podWatcher.failOnInitContainerFailure` calls `failForK8sObject` to fail the job via the Buildkite Agent REST API. If the Buildkite API returns `404 API::Error::NotFound` (the job has already been cleaned up on Buildkite's side), the function logs the error and returns **without adding the job to `ignoredJobs`** or deleting the pod/job from Kubernetes.

On the next informer re-sync, the same pod triggers `failOnInitContainerFailure` again, hitting the same 404, creating an infinite retry loop. Over time, this accumulates thousands of zombie Jobs and Pods that are never cleaned up, saturating etcd and the Kubernetes API server.

## Root cause

In [`internal/controller/scheduler/pod_watcher.go` (failOnInitContainerFailure)](https://github.com/buildkite/agent-stack-k8s/blob/main/internal/controller/scheduler/pod_watcher.go):

```go
if err := failForK8sObject(ctx, log, pod, failureInfo, w.agentClient); err != nil {
    log.Error("Could not fail Buildkite job", "error", err)
    podWatcherBuildkiteJobFailErrorsCounter.Inc()
    return // BUG: does not call w.ignoreJob(jobUUID) or forcefullyDeletePod
}
```

Compare with `failForImageFailure` which correctly handles terminal job states:

```go
case api.JobStateCanceling, api.JobStateCanceled, api.JobStateFinished, api.JobStateSkipped:
    log.Debug("Job not acquirable or cancelable")
    w.ignoreJob(jobUUID) // correctly prevents further processing
```

## To Reproduce

1. Run agent-stack-k8s at scale with spot instances (we run ~11,500 max-in-flight)
2. Spot nodes get reclaimed mid-init, causing init containers to fail
3. By the time the controller tries to fail the job on Buildkite, the job has already been cleaned up/cancelled on Buildkite's side
4. The Buildkite API returns `404 API::Error::NotFound`
5. The controller retries on every informer re-sync, never cleaning up the K8s Job/Pod

## Impact (observed at scale)

| Metric | HRNS Cluster | Green Cluster |
|--------|-------------|---------------|
| Controller version | 0.38.0 | 0.38.0 |
| Stuck Jobs | 11,502 | 1,158 |
| 404 errors in last 2000 log lines | 423 | 404 |
| Unique stuck job UUIDs | 497 | - |

Controller log (repeating every few seconds):

```
ERR scheduler/pod_watcher.go:336 Could not fail Buildkite job component=podWatcher
    namespace=bk-rippling-main-ci-k8s-rplm-ci-pr-hrns-blue
    name=buildkite-019c7594-8066-4ed8-9a5f-62f4b98b0305-qjtw9
    jobUUID=019c7594-8066-4ed8-9a5f-62f4b98b0305
    error="finish job: POST https://agent.buildkite.com/v3/stacks/.../jobs/.../finish: 404 API::Error::NotFound"
```

Jobs have `ttlSecondsAfterFinished=2` and condition `FailureTarget`, but remain for hours because they were created and never properly completed/deleted by the controller.

## Expected behavior

When `failForK8sObject` returns a 404 error (permanent/non-retryable), `failOnInitContainerFailure` should:

1. Detect that the 404 means the job no longer exists on Buildkite
2. Call `w.ignoreJob(jobUUID)` to stop re-processing
3. Call `forcefullyDeletePod()` to clean up the K8s pod
4. Let the Job's `ttlSecondsAfterFinished` handle Job deletion (or delete it directly)

## Suggested fix

```go
if err := failForK8sObject(ctx, log, pod, failureInfo, w.agentClient); err != nil {
    log.Error("Could not fail Buildkite job", "error", err)
    podWatcherBuildkiteJobFailErrorsCounter.Inc()

    // If the error is permanent (e.g. 404 - job no longer exists on Buildkite),
    // ignore the job and clean up the pod to prevent infinite retry loops.
    if api.IsPermanentError(err) {
        jobUUID, _ := jobUUIDForObject(pod)
        w.ignoreJob(jobUUID)
        forcefullyDeletePod(ctx, log, w.k8s, &pod.ObjectMeta, "init_container_failure_permanent_error")
    }
    return
}
```

## Environment

- agent-stack-k8s version: v0.38.0 (also reproducible on v0.32.2 and likely all versions)
- Kubernetes version: EKS 1.29+
- Scale: ~11,500 max-in-flight jobs, ARM spot instances with Karpenter

zhming0:
### Problem

When a pod completes, native sidecar init containers (`restartPolicy: Always`) are terminated by the kubelet. If any sidecar exits with a non-zero code — either from SIGKILL (137) after the grace period or SIGTERM (143) if the process doesn't handle it gracefully — `failOnInitContainerFailure` misinterprets this as an init container startup failure and attempts to fail the already-finished Buildkite job, which returns 404.

At high scale (200+ concurrent jobs), this creates a burst of blocking 404 API calls that stalls the informer event loop, delaying new job pickup.

Fixes #827.

### Fix

Skip init containers with `RestartPolicy: Always` in `failOnInitContainerFailure`. These are native K8s sidecar containers that run alongside main containers — their non-zero exit codes during pod shutdown are expected, not failures.

### DIFF
--- a/internal/controller/scheduler/pod_watcher.go
+++ b/internal/controller/scheduler/pod_watcher.go
@@ -363,9 +363,15 @@ func (w *podWatcher) failOnInitContainerFailure(ctx context.Context, log *slog.L
 
 	// If any init container fails, whether it's one we added specifically to
 	// check for pull failure or not, the pod won't run.
+	// However, native sidecar containers (init containers with RestartPolicy:
+	// Always) run alongside main containers and are expected to terminate with
+	// non-zero exit codes (e.g. SIGKILL/SIGTERM) during normal pod shutdown.
 	for _, containerStatus := range pod.Status.InitContainerStatuses {
 		term := containerStatus.State.Terminated
 		if term == nil || term.ExitCode == 0 { // not terminated, or succeeded
+			continue
+		}
+		if isSidecarInitContainer(pod, containerStatus.Name) {
 			continue
 		}
 		containerFails[containerStatus.Name] = term
@@ -789,6 +795,18 @@ func (w *podWatcher) stopWatchingForPendingTimeout(jobUUID uuid.UUID) {
 	w.watchingForPendingTimeoutMu.Lock()
 	defer w.watchingForPendingTimeoutMu.Unlock()
 	delete(w.watchingForPendingTimeout, jobUUID)
+}
+
+// isSidecarInitContainer reports whether the named init container is a native
+// sidecar (has RestartPolicy set to Always). These containers run alongside
+// main containers and should not be treated as gating init containers.
+func isSidecarInitContainer(pod *corev1.Pod, name string) bool {
+	for _, c := range pod.Spec.InitContainers {
+		if c.Name == name {
+			return c.RestartPolicy != nil && *c.RestartPolicy == corev1.ContainerRestartPolicyAlways
+		}
+	}
+	return false
 }
 
 // All container-\d containers will have the agent installed as their PID 1.
--- a/internal/controller/scheduler/pod_watcher_test.go
+++ b/internal/controller/scheduler/pod_watcher_test.go
@@ -6,9 +6,11 @@ import (
 	"testing"
 	"time"
 
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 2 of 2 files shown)


CerealBoy (approved):
LGTM 🚀 🚀

PR #866, opened 2026-04-17
chore: update values with agentStackSecret

tylerauerbeck:
This adds the `agentStackSecret` value to the `values.yaml` file to make it more explicit that a token can be defined within a secret instead of passed in directly via `agentToken`


zhming0 (approved):
Thank you.

PR #871, opened 2026-04-29
feat: add workspace-mount-sub-path-expr config option

csweeney-plaid:
### What it is

When set, applies a `subPathExpr` to every /workspace VolumeMount the controller creates (workspace setup init container, image check init containers, agent + checkout system containers, and command containers). 

This allows a shared workspace volume — e.g. a hostPath rooted at the node's /var/lib/buildkite-workspaces — to be sliced into per-pod subdirectories so concurrent jobs on the same node don't collide on /workspace/build, /workspace/sockets, etc.

### Why 

this enables a per-node shared-cache pattern for `docker run` workloads that Plaid is using — a single dockerd DaemonSet on each node bind-mounts the shared workspace hostPath, and per-pod subdirectories prevent concurrent jobs from clobbering each other's workspace state. 
Without this option, per-pod isolation is impossible because the workspace mount is hardcoded with no SubPath.

### Usage

The canonical setup Plaid is using this in is `workspace-mount-sub-path-expr: $(POD_NAME)` paired with a downward-API POD_NAME env var on the relevant containers (via pod-spec-patch or the chart's checkout-params / command-params).

Default ("") preserves the previous behavior.

Tests cover both the configured-value-propagates-everywhere case and the default-no-subPath case.

### DIFF
--- a/charts/agent-stack-k8s/values.schema.json
+++ b/charts/agent-stack-k8s/values.schema.json
@@ -364,6 +364,12 @@
         "workspaceVolume": {
           "$ref": "https://raw.githubusercontent.com/buildkite/kubernetes-json-schema/master/v1.35.0/_definitions.json#/definitions/io.k8s.api.core.v1.Volume"
         },
+        "workspace-mount-sub-path-expr": {
+          "type": "string",
+          "default": "",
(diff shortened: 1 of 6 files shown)


DrJosh9000 (approved):
Thanks @csweeney-plaid! The idea is sound, and the code LGTM. ✅

PR #872, opened 2026-04-29
SUP-6977: Expose reservation-expiry-seconds in Helm chart values

stephanieatte:
## Description
Exposes `reservation-expiry-seconds` in the Helm chart, so users can configure how long a job reservation is held. The controller already supports this config option (added in #847) but it wasn't visible in the Helm chart.

## Changes

- Added `reservation-expiry-seconds: 900` to charts/agent-stack-k8s/values.yaml

- Added `reservation-expiry-seconds` schema entry to charts/agent-stack-k8s/values.schema.json with `default: 900`, `minimum: 0`, `maximum: 3600`


DrJosh9000 (approved):
LGTM, thanks ✅

PR #875, opened 2026-05-04
fix: excess memory use in list-throttle condition

### LINKED ISSUE #874: [BUG] runaway controller memory usage when k8s control plane starts throttling pod LIST calls
## Describe the bug
When the k8s control plane starts throttling and rate-limiting pod LIST requests, we're less able to keep up with culling "Succeeded" pods from k8s. When this happens memory on the buildkite `controller` container grows significantly, in our case quickly blowing up to over 60gb when given unbounded memory. In practice it hits its memory request limit, OOMs, restarts, then within 10 mins OOMs again for the same reason.

Some o11y from a recent iteration of this failure mode on our installation:

Image

## Expected behavior
No controller OOMs

## Environment
- agent-stack-k8s version: 0.41.0
- Kubernetes version: 1.32
- Deployment method: helm

## Additional context

Here is an agent's take at diagnosing the problem through looking at the controller code. I have read through it and believe it looks like a reasonable issue path, but obvious grain of salt.

<details>

<summary>Codex 5.5's take</summary>
<br/>
This looks like an informer/cache amplification problem, with `completionsWatcher` as the hottest bad actor.

The controller creates Kubernetes Jobs with `TTLSecondsAfterFinished`, so ordinary cleanup of completed jobs/pods is delegated to Kubernetes TTL-after-finished cleanup, not to the Buildkite controller itself:

https://github.com/buildkite/agent-stack-k8s/blob/bc5d86d5e96f7aa3f69ebe55062f88f7121f5c16/internal/controller/scheduler/scheduler.go#L378-L379

Buildkite’s own cleanup utility also says the stack relies on `ttlSecondsAfterFinished` for automatic Job cleanup:

https://github.com/buildkite/agent-stack-k8s/blob/bc5d86d5e96f7aa3f69ebe55062f88f7121f5c16/utils/clean-up-job.yaml#L1-L7

When that cleanup falls behind, the controller’s informer watches every matching Buildkite Pod/Job in the namespace using only `buildkite.com/job-uuid` plus controller-id label filtering. There is no phase/status filter, so retained `Succeeded` pods are still part of the pod informer’s object set:

https://github.com/buildkite/agent-stack-k8s/blob/bc5d86d5e96f7aa3f69ebe55062f88f7121f5c16/internal/controller/controller.go#L301-L341

The really pathological part is `completionsWatcher`: it registers as a Pod informer handler, and `OnAdd` calls `cleanupSidecars` for every pod from the initial informer list:

https://github.com/buildkite/agent-stack-k8s/blob/bc5d86d5e96f7aa3f69ebe55062f88f7121f5c16/internal/controller/scheduler/completions.go#L68-L88

`cleanupSidecars` then synchronously does a Kubernetes Job `Get` followed by `Update` whenever the agent container has terminated:

https://github.com/buildkite/agent-stack-k8s/blob/bc5d86d5e96f7aa3f69ebe55062f88f7121f5c16/internal/controller/scheduler/completions.go#L118-L136

That includes old `Succeeded` pods from the initial LIST, even though they do not need sidecar cleanup anymore. Under Kubernetes API throttling, each handler call can block or slow down on those GET/UPDATE calls.

This matters because `client-go` shared informers keep a local cache of relevant objects:

https://github.com/kubernetes/client-go/blob/v0.35.3/tools/cache/shared_informer.go#L35-L55

And `client-go`’s per-listener `pendingNotifications` buffer is unbounded; the source comment explicitly says a stalled listener can accumulate pending notifications until OOM:

https://github.com/kubernetes/client-go/blob/v0.35.3/tools/cache/shared_informer.go#L945-L967

So the failure mode is:

1. TTL cleanup or delete visibility falls behind.
2. Completed Buildkite pods accumulate.
3. The controller informer caches all matching pods.
4. On startup/relist, `completionsWatcher.OnAdd` tries per-pod Kubernetes GET/UPDATE work for terminated-agent pods.
5. API throttling slows that handler.
6. `client-go` buffers notifications without a hard bound.
7. RAM scales with retained pod count, pod object size, and queued notifications.

LIST throttling makes this worse because reflectors start by listing all items and then watching from that resource version:

https://github.com/kubernetes/client-go/blob/v0.35.3/tools/cache/reflector.go#L406-L420

The reflector list path also uses chunked LISTs when supported, so a large retained population makes every relist expensive:

https://github.com/kubernetes/client-go/blob/v0.35.3/tools/cache/reflector.go#L607-L633

Most direct fixes:

- In `completionsWatcher.OnAdd`, skip terminal pods immediately: if `pod.Status.Phase` is `Succeeded` or `Failed`, return.
- Move Kubernetes GET/UPDATE work out of informer callbacks into a bounded, deduping, rate-limited workqueue.
- Only enqueue pods where the agent terminated but the pod is still non-terminal, which is the actual unmanaged-sidecar case.
- Consider a pod informer transform or narrower watch/cache shape so the controller does not retain full pod specs/statuses longer than needed.

Shortening `job-ttl` or adding an external cleanup job may reduce steady-state object count, but it does not fix the controller memory hazard. The code path that turns “many retained completed pods” into “controller OOMs” is the pod informer plus synchronous `completionsWatcher` cleanup.
</details>

DrJosh9000:
### What

- Strip pod and job objects with an informer transform, which should cut down on some memory usage in the informer cache and notification queue
- Make the completion watcher opt-out, so a source of memory consumption can be completely disabled if the completion watcher isn't needed (e.g. because no unmanaged containers are added via a podSpecPatch)
- Ignore succeeded, failed, and unknown phase pods in `cleanupSidecars`, since these are already completed.

### What not

- Filtering objects in the informer: this can be done with a label or field selector, but adding a filter on `status.phase` seems to break the controller entirely (that field doesn't exist for jobs)
- A bounded work queue for events in the completion watcher: that would bound the memory consumption, but when the limit is reached, cleanups will stop being set, which would then cause a pod leak instead. 

### Why

Fixes #874

### DIFF
--- a/internal/controller/scheduler/completions.go
+++ b/internal/controller/scheduler/completions.go
@@ -112,6 +112,16 @@ func (w *completionsWatcher) OnUpdate(old any, new any) {
 // Always), this cleanup is redundant but harmless as Kubernetes already handles
 // their termination automatically.
 func (w *completionsWatcher) cleanupSidecars(ctx context.Context, pod *corev1.Pod) {
+	// Ignore completed (succeeded or failed) and unknown status pods.
+	// Most of the time, the remaining logic applies to "Running" pods.
+	// ("Pending" is dubious... one or more of the containers haven't started,
+	// but the agent has run and terminated? still, it seems theoretically
+	// possible.)
+	switch pod.Status.Phase {
+	case corev1.PodSucceeded, corev1.PodFailed, corev1.PodUnknown:
+		return
+	}
+	// Ignore pods where the agent hasn't terminated.
 	terminated := getAgentTermination(pod)
 	if terminated == nil {
 		return
--- a/charts/agent-stack-k8s/values.schema.json
+++ b/charts/agent-stack-k8s/values.schema.json
@@ -262,7 +262,7 @@
           "type": "string",
           "default": "buildkite-",
           "title": "The prefix to use when creating Kubernetes job names",
-          "examples": ["kubekite-","buildkube-"]
+          "examples": ["kubekite-", "buildkube-"]
         },
         "job-active-deadline-seconds": {
           "type": "integer",
@@ -347,6 +347,12 @@
           "default": false,
           "title": "Allow controller to pause processing the jobs when queue is paused on Buildkite",
           "examples": [false]
+        },
+        "enable-completion-watcher": {
+          "type": "boolean",
+          "default": true,
+          "title": "Runs an extra loop that checks for and terminates pods that are still running but with a terminated agent container",
+          "examples": [false, true]
         },
         "prometheus-port": {
           "type": "integer",
--- a/charts/agent-stack-k8s/values.yaml
+++ b/charts/agent-stack-k8s/values.yaml
@@ -8,6 +8,7 @@ tolerations: []
 config:
   prometheus-port: 9216
   reservation-expiry-seconds: 900
+  enable-completion-watcher: true
 
 resources:
   requests:
--- a/cmd/controller/controller_test.go
+++ b/cmd/controller/controller_test.go
@@ -49,6 +49,7 @@ func TestReadAndParseConfig(t *testing.T) {
 		DefaultImagePullPolicy:               "Never",
 		DefaultImageCheckPullPolicy:          "IfNotPresent",
 		EnableQueuePause:                     true,
+		EnableCompletionWatcher:              true,
 		WorkQueueLimit:                       2_000_000,
 		ImageCheckContainerCPULimit:          "201m",
 		ImageCheckContainerMemoryLimit:       "129Mi",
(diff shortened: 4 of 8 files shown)


internal/controller/scheduler/completions.go
```diff
@@ -112,6 +112,16 @@ func (w *completionsWatcher) OnUpdate(old any, new any) {
...
 func (w *completionsWatcher) cleanupSidecars(ctx context.Context, pod *corev1.Pod) {
+	// Ignore completed (succeeded or failed) and unknown status pods.
+	// Most of the time, the remaining logic applies to "Running" pods.
+	// ("Pending" is dubious... one or more of the containers haven't started,
+	// but the agent has run and terminated? still, it seems theoretically
+	// possible.)
+	switch pod.Status.Phase {
+	case corev1.PodSucceeded, corev1.PodFailed, corev1.PodUnknown:
+		return
+	}
```
zhming0, line 123: The PR gives an impression that this bit is fixing the memory use but it might not move the needle much? 

Reducing informer cache usage might be the real fix? Is it possible to strip some detail from podSpec to reduce informer RAM usage?

zhming0 (approved):
I don't see a blocker so ✅ , but I'd advise that we reframe the PR description to point out that this isn't a fix,  instead it's a opt-out thingy.

internal/controller/scheduler/completions.go
DrJosh9000: On the theory that the informer's notification queue is filling with pods in a completed phase, and the subsequent work is slowing down processing the event, then this should be helping to reduce memory consumption (by allowing the queue to be processed faster).

I'll look into using a transform func to strip the objects.

zhming0: Ah make sense, thanks for clarifying. 👍🏿

PR #878, opened 2026-05-05
Replace and prevent assert/require

DrJosh9000:
### What

Like buildkite/agent#3858: remove uses of github.com/stretchr/testify, and prevent new uses from being added.

### Why

https://www.notion.so/buildkite/Assert-is-bad-33db8dbc2c8980ca8421e9ad47e0dddc?v=217b8dbc2c898063af47000cce90e7c4&source=copy_link


DrJosh9000:
@buildsworth-bk review, I'm OK with you approving L2 risk here

Issue #873, opened 2026-04-30
Sidecar containers still misclassified as init-container failures despite v0.42.0 fix


joeyqliu:
## Summary

We're running `agent-stack-k8s` v0.42.0. The v0.42.0 release notes call out *"Stop detecting init container failure for sidecar containers"* as a fix, but we're still seeing the same symptom: pods are torn down ~10 seconds after a sidecar container starts, with the controller logging `Cancelling deletion of Pod …` immediately after — suggesting the controller noticed the deletion-in-progress and tried to abort it, but the deletion had already been initiated.

The downstream cost is significant: when this fires, the K8s Job dies in seconds, but the Buildkite-server reservation TTL holds the slot for ~15 minutes before re-dispatch. Each affected build burns ~15 minutes of wall-clock waiting for the retry to start.

## Environment

- `agent-stack-k8s` controller: **v0.42.0** (helm chart `oci://ghcr.io/buildkite/helm/agent-stack-k8s:0.42.0`)
- `buildkite-agent` in pods: **v3.120.1**
- Kubernetes: GKE 1.34, regular release channel
- Node pool: `c4-standard-16` spot

## Pod shape

Standard `pod-spec-patch` shape:

- **Init containers**: `chmod`, `copy-agent`
- **Workload containers**: `agent`, `checkout`, `container-0`
- **Sidecars** (defined in step pipeline YAML): `postgres`, `elasticsearch8`, `firestore`, `pubsub`, `redis`

The failure consistently presents around the `elasticsearch8` sidecar coming up. The sidecar's startup probe budget is `initialDelaySeconds: 10` + `failureThreshold: 30 × periodSeconds: 5` = **160 s** before kubelet would mark the startup failed — so the kill at +8 seconds is *not* coming from the startup probe.

## Timeline of one affected build

Job UUID: `019ddf37-69d6-4865-bffa-a35233c28508`. Times in UTC, 2026-04-30:

| time | event | source |
|---|---|---|
| 16:29:51 | Controller created the K8s Job (`K8s Job created successfully, waiting for Agent to start...`) | Buildkite stack notification |
| 16:29:58 | First pod scheduled to a spot c4-standard-16 node, init container `chmod` started | GCP Cloud Logging |
| 16:29:59 | init container `copy-agent` started | GCP Cloud Logging |
| 16:30:00 | sidecar `elasticsearch8` `Started container` | GCP Cloud Logging |
[...]

PR #881, opened 2026-05-07
A-1202 part 1: documenting pod-pending-timeout via helm chart

zhming0:
Explosing pod-pending-timeout config via helm chart. This is one of the asks from #873, we have a few follow up PRs incoming.


DrJosh9000 (approved):
LGTM

PR #882, opened 2026-05-07
fix: Keep some podspec fields

DrJosh9000:
### What

Fix bug introduced with #875 

### Why

The podwatcher component depends on InitContainers in the pod spec to determine which are sidecars. So we shouldn't zero out `Spec` entirely in the informer transform.


zhming0 (approved):
LGTM ✅

PR #883, opened 2026-05-07
A-1202 part 2: handle pod failed before agent start

### LINKED ISSUE #873: Sidecar containers still misclassified as init-container failures despite v0.42.0 fix

zhming0:
## Problem

When a Kubernetes pod fails before the `buildkite-agent` container has a chance to acquire the Buildkite job (e.g. OOMKill, eviction, node failure, external `kubectl delete pod`), the BK job stays in `reserved` state until the reservation TTL (~15 min) expires. Re-dispatch is delayed by the full TTL.

Fixes #873 (ask #3).

Alternative, a theoretically more correct solution would be moving the `agent` container to `initContainer` as a sidecar too, but this will have an impact on how exit signal is handled, so it need a bit of more thinking. 

Even if we were going to make the `agent` as `initContainer`, this change won't harm. 

## Change

Extend `jobWatcher`'s finished-job handling: when the K8s Job ends with `Status.Failed > 0`, query BK for the job state. If still `reserved`, call `FailJob` to release the reservation immediately.

Only acts on `reserved` state — `scheduled` (no reservation made) and any post-acquire state are left to BK.

### DIFF
--- a/internal/integration/integration_test.go
+++ b/internal/integration/integration_test.go
@@ -913,6 +913,67 @@ func TestCancelCheckerDeletePod(t *testing.T) {
 	}
 }
 
+// TestPodKilledBeforeAgentReleasesBKJob exercises the jobWatcher path that
+// fails the corresponding Buildkite job when a K8s pod fails before the
+// buildkite-agent gets a chance to acquire it (e.g. OOM, eviction, external
+// `kubectl delete pod`). Without this, the BK job would sit in `reserved`
+// state until the reservation TTL (~15 min) expires.
+func TestPodKilledBeforeAgentReleasesBKJob(t *testing.T) {
+	tc := testcase{
+		T:       t,
+		Fixture: "pod-killed-before-agent.yaml",
+		Repo:    repoHTTP,
+		GraphQL: api.NewGraphQLClient(cfg.BuildkiteToken, cfg.GraphQLEndpoint),
+	}.Init()
+	ctx := context.Background()
+	pipelineID := tc.PrepareQueueAndPipelineWithCleanup(ctx)
+	tc.StartController(ctx, cfg)
+	build := tc.TriggerBuild(ctx, pipelineID)
+
+	job := build.Jobs.Edges[0].Node.(*api.JobJobTypeCommand)
+	jobName := cfg.JobPrefix + job.Uuid
+	opts := metav1.ListOptions{
+		LabelSelector: fields.OneTermEqualSelector("batch.kubernetes.io/job-name", jobName).String(),
+	}
+
+	// Wait for the pod to exist (init container will be sleeping).
+	var pods *corev1.PodList
+	if err := roko.NewRetrier(
+		roko.WithMaxAttempts(10),
+		roko.WithStrategy(roko.Exponential(2*time.Second, 0)),
+	).DoWithContext(ctx, func(r *roko.Retrier) error {
+		got, err := tc.Kubernetes.CoreV1().Pods(cfg.Namespace).List(ctx, opts)
+		if err != nil {
+			return err
+		}
+		if len(got.Items) == 0 {
+			return fmt.Errorf("found no pod, waiting for pod start")
+		}
+		pods = got
+		return nil
+	}); err != nil {
+		t.Fatalf("waiting for pod to exist: %v", err)
+	}
+
+	// Force-delete the pod with grace=0 to simulate an external kill before
+	// the agent container has a chance to start.
+	gracePeriod := int64(0)
+	for _, p := range pods.Items {
+		if err := tc.Kubernetes.CoreV1().Pods(cfg.Namespace).Delete(ctx, p.Name, metav1.DeleteOptions{
+			GracePeriodSeconds: &gracePeriod,
+		}); err != nil {
+			t.Fatalf("deleting pod %q: %v", p.Name, err)
+		}
+	}
+
+	tc.AssertFail(ctx, build)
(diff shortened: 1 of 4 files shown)


internal/integration/integration_test.go
```diff
@@ -913,6 +913,69 @@ func TestCancelCheckerDeletePod(t *testing.T) {
...
+	opts := metav1.ListOptions{
+		LabelSelector: fields.OneTermEqualSelector("batch.kubernetes.io/job-name", jobName).String(),
+	}
+
+	// Wait for the pod to exist (init container will be sleeping).
+	if err := roko.NewRetrier(
+		roko.WithMaxAttempts(10),
+		roko.WithStrategy(roko.Exponential(2*time.Second, 0)),
+	).DoWithContext(ctx, func(r *roko.Retrier) error {
+		exist, err := hasJobPod(tc, ctx, opts)
```
DrJosh9000: `hasJobPod` calls the same `List` method as used below to list the pods, so perhaps this should call `List` and save the returned list of pods for deletion?

DrJosh9000 (approved):
Logic seems good to me, I have 1 suggestion for the test

PR #889, opened 2026-05-20
limiter: sort the work queue, not the input batch

vladi99:
## Change

One-character fix in `Limiter.HandleMany`: switch the `slices.SortStableFunc` target from the `jobs` parameter to `l.queue`. Plus five new tests in `internal/controller/limiter/limiter_priority_test.go`.

## Context

In `HandleMany`, the post-append `SortStableFunc` call sorted the `jobs` parameter instead of `l.queue`. After `l.queue = append(l.queue, jobs...)` copies the pointers into `l.queue`'s backing array, sorting `jobs` no longer affects `l.queue` — they're independent slice headers backed by independent arrays. Net effect:

- Workers pop `l.queue[0]`, so dequeue was insertion-order (FIFO) regardless of priority.
- The `WorkQueueLimit` truncation dropped by arrival recency, not by lowest priority — so the *newest* arrivals got evicted, often including high-priority jobs that just landed.
- `Priority` only influenced ordering at the BK API → controller boundary (because `getScheduledJobs` returns priority-sorted pages); once a wave was buffered inside the controller, priority did nothing.

Both callers — `Reserver.HandleMany` and `Monitor` — only use `len()` of the slice after `HandleMany` returns, so the in-place side effect on `jobs` was invisible upstream and the wrong behavior had no failing test.

Targeting `l.queue` instead makes the two adjacent comments ("Sort by priority…", "highest priority jobs are more likely to remain") describe what the code actually does.

**Tests added:**
- `TestSortingSrcDoesNotReorderDst` pins the Go-semantics premise (deterministic at both sufficient and forced-realloc capacities).
- `TestHandleMany_SortsBatchByPriority` covers single-batch ordering.
- `TestHandleMany_SortsAcrossBatches` covers a later high-priority job jumping ahead of earlier low-priority jobs.
- `TestHandleMany_WorkQueueLimitDropsByPriority` covers the truncation side of the fix.
- `TestHandleMany_ConcurrentProducersRaceFree` exercises the now-shared sort target under `-race`.

**Test plan:**
- `go test ./internal/controller/limiter/...` — all pass.
- `go test -race -run TestHandleMany_ConcurrentProducersRaceFree ./internal/controller/limiter/` — passes.
- Note: running multiple limiter tests in parallel under `-race` trips a pre-existing race on the package-level `tokensAvailableFunc` / `workQueueLengthFunc` closures in `metrics.go`, unrelated to this change.

**Performance note:** the fix now sorts the entire `l.queue` on each `HandleMany` call. For a queue of several thousand entries this stays sub-millisecond inside the mutex critical section, so I don't think it warrants a data-structure change for this PR — but a heap-based priority queue would be more elegant if you wanted to revisit it separately.

Meta — found this while debugging a high-priority job that took unexpectedly long to be picked up by an agent during a large CI burst.

## Disclosures / Credits

Investigation and patch developed with assistance from Claude (Anthropic).

### DIFF
--- a/internal/controller/limiter/limiter.go
+++ b/internal/controller/limiter/limiter.go
@@ -171,7 +171,7 @@ func (l *Limiter) HandleMany(ctx context.Context, jobs []*api.AgentScheduledJob)
 
 	// Sort by priority, preserving existing ordering between jobs with equal
 	// priority.
-	slices.SortStableFunc(jobs, func(a, b *api.AgentScheduledJob) int {
+	slices.SortStableFunc(l.queue, func(a, b *api.AgentScheduledJob) int {
 		// Higher number = higher priority.
 		// See https://buildkite.com/docs/pipelines/configure/workflows/managing-priorities
 		return cmp.Compare(b.Priority, a.Priority)
--- a/internal/controller/limiter/limiter.go
+++ b/internal/controller/limiter/limiter.go
@@ -14,10 +14,11 @@ import (
 	"github.com/buildkite/agent-stack-k8s/v2/internal/controller/config"
 	"github.com/buildkite/agent-stack-k8s/v2/internal/controller/model"
 
+	"log/slog"
+
 	batchv1 "k8s.io/api/batch/v1"
 	"k8s.io/client-go/informers"
 	"k8s.io/client-go/tools/cache"
-	"log/slog"
 )
 
 // Limiter manages a queue of jobs. While there are jobs available and less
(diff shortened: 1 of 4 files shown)


DrJosh9000 (approved):
Thanks @vladi99 , the fix looks sound. ✅ 

I've gone ahead and pushed some Go modernisations to your tests (t.Context added in Go 1.24, WaitGroup.Go added in Go 1.25, math/rand/v2 added in Go 1.22) as well as a fix to the data race on the metrics callback funcs. With that, LGTM.

PR #896, opened 2026-06-12
Add additional hooks configuration

jamiemonserrate:
## Change

- Adds first-class `agent-config.additional-hooks-paths` and `agent-config.additional-hooks` config/schema support for image-baked and volume-backed hook directories.
- Mounts additional hook volumes into agent, checkout, and command containers, and sets `BUILDKITE_ADDITIONAL_HOOKS_PATHS` for the agent start container.
- Validates `additional-hooks` entries and updates the example config.
- Adds controller/scheduler unit coverage plus an integration fixture and test for additional lifecycle hooks.

## Context

Linear: [A-1349: [canva] Unable to set BUILDKITE_ADDITIONAL_HOOKS_PATHS on agent](https://linear.app/buildkite/issue/A-1349/canva-unable-to-set-buildkite-additional-hooks-paths-on-agent)

[#3975](https://github.com/buildkite/agent/issues/3975)  reported a regression from agent `v3.127.1` where `BUILDKITE_ADDITIONAL_HOOKS_PATHS` set through a container env var was overwritten with an empty string. This change gives agent-stack-k8s a supported configuration path for additional hooks instead of relying on container-level env overrides.

Buildkite Agent supports `BUILDKITE_ADDITIONAL_HOOKS_PATHS`; this exposes that through agent-stack-k8s config for hook directories already present in images, plus hook directories mounted from Kubernetes volumes.

### Verification

- Tested a locally running controller against local bk/bk with additional hooks from:
  - an image-baked path: `/opt/buildkite/additional-hooks-image`
  - a ConfigMap-mounted path: `/buildkite/additional-hooks/configmap`
  - a Secret-mounted path: `/buildkite/additional-hooks/secret`
- Confirmed the job output ran the image, ConfigMap, and Secret `pre-command` hooks before the repository hook.
- Tested a Helm-deployed controller with `additional-hooks-paths` plus ConfigMap and Secret backed `additional-hooks`; triggered another build and confirmed the hooks ran correctly.
- Applied the new `.buildkite/fixtures/additional-hooks.yaml` fixture to the shared EKS-backed integration cluster.

### Deployment

The new config is opt-in. Existing installs are unchanged unless they set `agent-config.additional-hooks-paths` or `agent-config.additional-hooks`.

### Rollback

Revert this PR. It only adds config fields, volume mounts, env wiring, schema/example updates, and tests; there are no migrations or feature flags. If the CI fixture is no longer needed, remove the `integration-tests-fixture-additional-hooks` ConfigMap from `buildkite-k8s-integration-test`.

N/A

## Disclosures / Credits

Drafted with Codex.

### DIFF
--- a/cmd/controller/flags.go
+++ b/cmd/controller/flags.go
@@ -280,6 +280,10 @@ func validateConfig(cfg *config.Config) error {
 		return err
 	}
 
+	if err := validateAgentConfig(cfg.AgentConfig); err != nil {
+		return err
+	}
+
 	if cfg.PodSpecPatch != nil {
 		if err := validatePodSpecPatch(cfg.PodSpecPatch, cfg.AllowPodSpecPatchUnsafeCmdMod); err != nil {
 			return fmt.Errorf("invalid pod spec patch: %w", err)
@@ -300,6 +304,25 @@ func validateConfig(cfg *config.Config) error {
 		}
 		if _, exists := cfg.ResourceClasses[cfg.DefaultResourceClassName]; !exists {
 			return fmt.Errorf("default-resource-class-name %q not found in resource-classes", cfg.DefaultResourceClassName)
+		}
+	}
+
+	return nil
+}
+
+func validateAgentConfig(agentConfig *config.AgentConfig) error {
+	if agentConfig == nil {
+		return nil
+	}
+
+	for i, h := range agentConfig.AdditionalHooks {
+		switch {
+		case h.Path == "":
+			return fmt.Errorf("agent-config.additional-hooks[%d].path is required", i)
+		case h.Volume == nil:
+			return fmt.Errorf("agent-config.additional-hooks[%d].volume is required", i)
+		case h.Volume.Name == "":
+			return fmt.Errorf("agent-config.additional-hooks[%d].volume.name is required", i)
 		}
 	}
 
--- a/cmd/controller/flags.go
+++ b/cmd/controller/flags.go
@@ -3,6 +3,7 @@ package controller
 import (
 	"bytes"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"os"
 	"reflect"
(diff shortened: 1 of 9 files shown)


jamiemonserrate:
@buildsworth-bk - Can you review please?

internal/integration/integration_test.go
```diff
@@ -1044,6 +1044,52 @@ func TestHooksAndPlugins(t *testing.T) {
...
+	cfg.AgentConfig = &config.AgentConfig{
+		AdditionalHooks: []config.AdditionalHook{
+			{
+				Path: "/buildkite/additional-hooks",
+				Volume: &corev1.Volume{
+					Name: "additional-hooks",
+					VolumeSource: corev1.VolumeSource{
+						ConfigMap: &corev1.ConfigMapVolumeSource{
+							LocalObjectReference: corev1.LocalObjectReference{
+								Name: "integration-tests-fixture-additional-hooks",
```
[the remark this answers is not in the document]
jamiemonserrate: Applying this config map is out of the scope of this repo. I've applied it and the checks are now passing.

jamiemonserrate:
@buildsworth-bk - Can you review again please?

cmd/controller/flags.go
```diff
@@ -306,6 +310,25 @@ func validateConfig(cfg *config.Config) error {
...
+		case h.Path == "":
+			return fmt.Errorf("agent-config.additional-hooks[%d].path is required", i)
+		case h.Volume == nil:
+			return fmt.Errorf("agent-config.additional-hooks[%d].volume is required", i)
+		case h.Volume.Name == "":
+			return fmt.Errorf("agent-config.additional-hooks[%d].volume.name is required", i)
+		}
+	}
+
+	return nil
```
DrJosh9000: Since this returns on the first error, for the first additional hook config with an error, the user would only get one validation error but their config could have several. Perhaps we should return an error representing all validation failures?

```suggestion
	var errs []error
	for i, h := range agentConfig.AdditionalHooks {
		if h.Path == "" {
			errs = append(errs, fmt.Errorf("agent-config.additional-hooks[%d].path is required", i))
		}
		switch {
		case h.Volume == nil:
			errs = append(errs, fmt.Errorf("agent-config.additional-hooks[%d].volume is required", i))
		
		case h.Volume.Name == "":
			errs = append(errs, fmt.Errorf("agent-config.additional-hooks[%d].volume.name is required", i))
		}
	}

	return errors.Join(errs...)
```

DrJosh9000 (approved):

cmd/controller/flags.go
jamiemonserrate: That makes sense. Thanks @DrJosh9000. Will address.

jamiemonserrate: I made another commit addressing these changes.

PR #899, opened 2026-06-19
build: use buildkite cache

buildkate:
## Change

- Defines Buildkite caches in  `cache.yml`
- Uses them in `pipeline.yml` instead of the `go-with-cache` pod template
- Removes `go-with-cache` template (the file)
- Consolidates installation of `aws-cli` and `jq` into `assume-role.sh`

TODO: remove pod template from the cluster (it was manually applied)

## Context

Resolves A-1119

## Disclosures / Credits

Claude for researching CI failures related to the target paths.


.buildkite/pipeline.yaml
```diff
@@ -12,34 +12,53 @@ x-anchors:
...
   - label: ":golang::robot_face: check code generation"
     key: check-code-generation
+    env: *go-cache-env
     image: public.ecr.aws/docker/library/golang:1.26.4-alpine@sha256:f1ddd9fe14fffc091dd98cb4bfa999f32c5fc77d2f2305ea9f0e2595c5437c14
-    command: .buildkite/steps/check-code-generation.sh
-    plugins:
-      - kubernetes:
-          podTemplate: go-with-cache
+    command:
+      - . .buildkite/steps/assume-role.sh
```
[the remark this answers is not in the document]
buildkate: Discussed with @zhming0 and we will accept this risk. We usually run with the assumption that all the commands within a command step have access to the same thing. TODO: create a separate role just for cache access at a later time.

.buildkite/pipeline.yaml
```diff
@@ -138,9 +168,10 @@ steps:
       REGISTRY_USERNAME: ghcr_username
       GITHUB_TOKEN: github_release_token
     image: public.ecr.aws/docker/library/golang:1.26.4-alpine@sha256:f1ddd9fe14fffc091dd98cb4bfa999f32c5fc77d2f2305ea9f0e2595c5437c14 # Note goreleaser shells out to go!
-    command: .buildkite/steps/release.sh
-    plugins:
-      - kubernetes:
-          podTemplate: go-with-cache
-          checkout:
-            fetchFlags: -v --tags
+    command:
```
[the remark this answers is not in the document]
buildkate: Updated

.buildkite/cache.yml
```diff
@@ -0,0 +1,22 @@
+caches:
+  - name: go_build
+    cache_key:
+      - agent: os
+      - {agent: arch, fallbackLimit: true}
+      - checksum: go.sum
```
zhming0, line 6: Is go.sum an appropriate cache key for the go build cache? 

But our option seems to be limited. 🤔

zhming0 (approved):
LGTM, finally k8s plugin free pipeline! The question around go build cache key is worth a slack chat I think

.buildkite/cache.yml
buildkate: @zhming0 I could not think of a perfect key for go build.. as the inputs are pretty much all the go files. 
As I understand it,  Go's build cache is internally content-addressed, so even if go.sum gives a stale cache, it can still reuse the unchanged parts.

DrJosh9000: Hey, just saw this. I would use `go.mod` and not `go.sum` for the cache key. Basically, `go.sum` isn't a lock file, all the dependency information is in `go.mod`.

DrJosh9000: https://words.filippo.io/gosum/

buildkate: Thanks @DrJosh9000 will update in another PR.

PR #906, opened 2026-07-03
job_watcher: serve diagnostic event LIST from watch cache

Megh03:
## Change
`fetchEvents()` in `job_watcher.go` now lists Kubernetes Events with `ResourceVersion: "0"`, so the read is served from the API server's in-memory watch cache instead of a quorum read from etcd. 

## Context

One of our customers reported that fetchEvents() in job_watcher.go is causing API server latency spikes on busy clusters. The function LISTs Kubernetes events using the default quorum read (hitting etcd), but these events are only used for diagnostic failure messages. Adding ResourceVersion: "0" to the ListOptions would serve these reads from the API server's in-memory watch cache instead, eliminating the etcd round-trip and reducing cluster-wide contention.


DrJosh9000 (approved):
LGTM, thanks @Megh03 !

PR #913, opened 2026-07-13
A-1673: Fix stale CI caches by keying them per branch and commit

zhming0:
## Problem

The `go_build` and `golangci_lint` caches were keyed on checksums of `go.mod` and `.golangci.yaml`. Since `cache save` never overwrites an existing entry, each cache was saved once and then never updated. As the code changed the caches went stale, hits dropped to near zero, and every build still paid the restore and save overhead.

## Fix

Key the evolving caches on `BUILDKITE_BRANCH` + `BUILDKITE_COMMIT`. Every commit saves a fresh entry, and restores fall back in order:

1. same commit
2. latest entry from the same branch
3. latest entry from any branch
4. anything with the same os/arch

`go_build` also gets `BUILDKITE_STEP_KEY` in its key. Each step builds a different Go cache (lint compiles under golangci-lint's toolchain, tests add test binaries and results, the controller build cross compiles). With a shared key, whichever step saved first claimed the commit's entry and the rest skipped saving. With the step key, each step keeps its own cache warm.

Both caches validate their own contents, so a stale restore only means some recompilation, never wrong results.

`go_mod` stays checksum keyed: the module cache is fully determined by the dependency graph, so it only needs to change when `go.mod` does.

## Also

`cache restore` opens a log group without closing it, which swallows the output that follows. Add an explicit log group heading after each restore.

### DIFF
--- a/.buildkite/cache.yml
+++ b/.buildkite/cache.yml
@@ -3,7 +3,9 @@ caches:
     cache_key:
       - agent: os
       - {agent: arch, fallback_limit: true}
-      - checksum: go.mod
+      - env: BUILDKITE_STEP_KEY
+      - env: BUILDKITE_BRANCH
+      - env: BUILDKITE_COMMIT
     target_paths:
       - /root/.cache/go-build
   - name: go_mod
@@ -17,6 +19,7 @@ caches:
     cache_key:
       - agent: os
       - {agent: arch, fallback_limit: true}
-      - checksum: .golangci.yaml
+      - env: BUILDKITE_BRANCH
+      - env: BUILDKITE_COMMIT
     target_paths:
       - /root/.cache/golangci-lint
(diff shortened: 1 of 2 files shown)


.buildkite/cache.yml
```diff
@@ -3,7 +3,9 @@ caches:
     cache_key:
       - agent: os
       - {agent: arch, fallbackLimit: true}
-      - checksum: go.mod
+      - env: BUILDKITE_STEP_KEY
```
buildkate, line 6: If we remove step key, the cache can be reused across steps in the same branch/commit combos.

zhming0: The steps' `GOCACHE`s don't overlap much: lint compiles everything under golangci-lint's toolchain, tests add test binaries and results, and the controller build cross compiles. Sharing one entry means most steps restore a cache built by a different workload and it basically nullified the cache benefit.

PR #916, opened 2026-07-13
Report signal_reason when failing jobs via the Stacks API

walkerdb:
## Problem

When the controller fails a Buildkite job for infrastructure reasons — pod pending timeout, init container failure, podSpec rejection, image-pull failure — the job lands in Buildkite as a bare `failed` / `exit_status: -1` with **no `signal_reason`**.

Every failure site already constructs `FailureInfo{Reason: agent.SignalReasonStackError}`, but `failForK8sObject` only forwards `failureInfo.Message`, and `stacksapi.FinishJobRequest` has no field to carry a reason — so `FailureInfo.Reason` is currently dead code.

This appears to be a regression from the temporary-agent era (a leftover comment in `fail_job.go` still referenced connecting "the temporary agent"): an agent finishing a job could report `signal_reason: stack_error`, which is why the [retry docs](https://buildkite.com/docs/pipelines/configure/retry) document matching on it and the v0.29.0 release notes advertise it for stack-level visibility. The Stacks API finish path lost it.

## Impact

- `retry: automatic: [{signal_reason: stack_error}]` — the documented way to auto-retry infra failures — can't match jobs failed by this controller.
- Observability: stack-failed jobs are indistinguishable from ordinary agent-lost jobs (`exit -1`) in the REST API, webhooks, EventBridge, and OpenTelemetry notification payloads, except by parsing free-text detail. (We found this while trying to build an infra-failure metric for our CI: jobs killed by pod-scheduling failures had no machine-readable marker at all.)

## Change

- Plumb `FailureInfo.Reason` through `failForK8sObject` → `AgentClient.FailJob` → `FinishJobRequest.SignalReason`.
- Set `SignalReasonStackError` on the image-pull-failure path, which previously sent no reason (it carried a `// Do we have a better status code to report here?` comment).
- Remove the stale temporary-agent comment.

## Dependencies

Depends on buildkite/stacksapi#7 adding `SignalReason` to `FinishJobRequest` — and on the `/stacks/.../finish` endpoint accepting the parameter server-side. If the server already supports it, only these client changes are needed; if not, please treat the pair of PRs as the feature request. The go.mod bump is intentionally omitted until stacksapi tags a release containing #7.

## Testing

`go build ./...` and `go test ./internal/controller/scheduler/...` pass with a local `replace` pointing at the stacksapi branch.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/api/agent_client.go
+++ b/api/agent_client.go
@@ -333,12 +333,16 @@ func (c *AgentClient) DeregisterStack(ctx context.Context) error {
 	return err
 }
 
-func (c *AgentClient) FailJob(ctx context.Context, jobUUID string, errorDetail string) error {
+// FailJob marks the job as finished with exit status -1, recording
+// errorDetail in the job log and signalReason (e.g. "stack_error") as the
+// job's machine-readable signal reason.
+func (c *AgentClient) FailJob(ctx context.Context, jobUUID string, errorDetail string, signalReason string) error {
 	req := stacksapi.FinishJobRequest{
-		StackKey:   c.stack.Key,
-		JobUUID:    jobUUID,
-		ExitStatus: -1,
-		Detail:     errorDetail,
+		StackKey:     c.stack.Key,
+		JobUUID:      jobUUID,
+		ExitStatus:   -1,
+		Detail:       errorDetail,
+		SignalReason: signalReason,
 	}
 
 	_, err := c.stacksAPIClient.FinishJob(ctx, req)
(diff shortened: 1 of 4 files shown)


isaacsu:
Thanks for submitting this @walkerdb. I'll be reviewing and hope to respond in a couple of days.

DrJosh9000:
I believe this has been resolved with a backend fix. As such I am closing. If this would still be useful, please let us know!

PR #917, opened 2026-07-15
Dogfood agent v4

DrJosh9000:
## Change

When the pipeline redeploys the stack to itself (the stack running the pipeline that builds the stack 🥴 ), use `buildkite/agent:beta` for the agent image, so that subsequent builds will use agent v4.

Note that this does **not** change the default image for the next release - that will still be v3.

## Context

To gain more confidence that agent v4 won't break stuff.

https://linear.app/buildkite/issue/A-1543/use-v4-in-agent-stack-k8s-pipeline

### DIFF
--- a/.buildkite/steps/deploy.sh
+++ b/.buildkite/steps/deploy.sh
@@ -20,4 +20,8 @@ helm upgrade agent-stack-k8s "${helm_repo_pecr}/agent-stack-k8s" \
   --wait \
   -f .buildkite/production-helm-values.yaml \
   --set agentToken="${DEFAULT_CLUSTER_TOKEN}" \
-  --set config.image="${agent_image}"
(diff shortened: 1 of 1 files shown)


zhming0 (approved):
LGTM but we might want to advertise broadly now fix a bug for k8s stack need to target the v4 branch 🤔 .

DrJosh9000:
> LGTM but we might want to advertise broadly now fix a bug for k8s stack need to target the v4 branch 🤔 .

I don't follow - this change isn't upgrading the default agent image for releases, just the agent image for our own pipeline.

zhming0:
ah yes, I was confused 🤦

DrJosh9000:
> ah yes, I was confused 🤦

All good, I could have written a clearer PR message. I've added more words.

PR #918, opened 2026-07-20
Add a guard to prevent false positive in stalled job detection

omehegan:
## Change

Adds a guard in `cleanupStalledJob` which checks the actual status of the job on the Buildkite platform. If the job is found to be in a running state, then it should not be cleaned up. This addresses an issue where the k8s cluster informer cache can lag under heavy load, causing us to mistakenly think that a job is stalled waiting for a pod when it is actually running normally.

## Context

Reference: [A-1565](https://linear.app/buildkite/issue/A-1565/notion-agent-unexpectedly-terminated)

## Disclosures / Credits

Claude was used to assist in analyzing the original issue based on cluster logs and metrics, and then in designing and implementing the proposed fix.

### DIFF
--- a/internal/controller/scheduler/job_watcher.go
+++ b/internal/controller/scheduler/job_watcher.go
@@ -193,7 +193,7 @@ func (w *jobWatcher) checkFinished(ctx context.Context, log *slog.Logger, jobUUI
 	log.Info("The Kubernetes job ended without starting a pod. Failing the corresponding Buildkite job")
 	message := "The Kubernetes job ended without starting a pod.\n"
 	message += w.fetchEvents(ctx, log, kjob)
-	w.failJob(ctx, log, kjob, message)
+	_ = w.failJob(ctx, log, kjob, message)
 }
 
 // failBeforeAgentAcquire is called when a k8s Job has finished with a failed
@@ -219,7 +219,7 @@ func (w *jobWatcher) failBeforeAgentAcquire(ctx context.Context, log *slog.Logge
 	log.Info("Kubernetes pod failed before the buildkite-agent acquired the Buildkite job. Failing the BK job to release the reservation.", "bk_job_state", string(state.State))
 	message := "The Kubernetes pod failed before the buildkite-agent acquired this Buildkite job. The pod was likely killed by Kubernetes (eviction, OOM, node failure) or terminated externally before the agent could start.\n"
 	message += w.fetchEvents(ctx, log, kjob)
-	w.failJob(ctx, log, kjob, message)
+	_ = w.failJob(ctx, log, kjob, message)
 	w.ignoreJob(jobUUID)
 }
 
@@ -392,7 +392,7 @@ func (w *jobWatcher) cleanupStalledJob(ctx context.Context, kjob *batchv1.Job) {
 
 	state, _, err := w.agentClient.GetJobState(ctx, jobUUID.String())
 	if err != nil {
-		log.Warn("Failed to fetch Buildkite job state; skipping stalled job cleanup to avoid killing a potentially running job", "error", err)
+		log.Warn("Failed to fetch BK job state; skipping stalled job cleanup to avoid killing a potentially running job", "error", err)
 		return
 	}
 
--- a/internal/controller/scheduler/job_watcher.go
+++ b/internal/controller/scheduler/job_watcher.go
@@ -273,7 +273,7 @@ func (w *jobWatcher) fetchEvents(ctx context.Context, log *slog.Logger, kjob *ba
 	return w.formatEvents(evlist)
 }
 
-func (w *jobWatcher) failJob(ctx context.Context, log *slog.Logger, kjob *batchv1.Job, message string) {
+func (w *jobWatcher) failJob(ctx context.Context, log *slog.Logger, kjob *batchv1.Job, message string) error {
 	failureInfo := FailureInfo{
 		Message: message,
 		// We can know almost all failures triggered by job watcher are stack related error.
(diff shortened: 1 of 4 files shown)


internal/controller/scheduler/job_watcher.go
```diff
@@ -381,11 +382,49 @@ func (w *jobWatcher) stalledJobChecker(ctx context.Context) {
...
+	state, _, err := w.agentClient.GetJobState(ctx, jobUUID.String())
+	if err != nil {
+		log.Warn("Failed to fetch BK job state; skipping stalled job cleanup to avoid killing a potentially running job", "error", err)
+		return
+	}
+
+	// Only proceed if the job is still in a pre-agent state.
+	// If the agent has already acquired the job (Running, Accepted, etc.),
+	// the "stalled without pod" signal was a false positive from a stale cache.
+	if state.State != api.JobStateReserved && state.State != api.JobStateScheduled {
```
petetomasik: `GetJobState` returns `State: state[id]` (a plain map index), so when Buildkite doesn't recognize the UUID (job deleted/expired/GC'd) `state.State` is the zero value `""`, which fails both checks and takes this early return path, same as a healthy running job. An orphaned k8s Job whose BK job is already gone is then left in place (and, per the ignore ordering flagged above, ignored permanently), even though those orphans are arguably exactly what this cleanup should reap. Can we distinguish "UUID absent from the response" from a real terminal/running state?

internal/controller/scheduler/job_watcher_test.go
```diff
@@ -0,0 +1,298 @@
...
+				config.UUIDLabel: jobUUID,
+			},
+		},
+		Status: batchv1.JobStatus{
+			StartTime: &now,
+		},
+	}
+}
+
+func TestCleanupStalledJob(t *testing.T) {
```
petetomasik, line 62: A few reachable branches aren't exercised yet: the recheck-also-errors abort (`recheckErr != nil`), the empty/unknown-state skip (leave `server.JobStates` empty so `GetJobState` returns `""`), and the new `jobWatcherStalledCleanupSkippedCounter`, which is never asserted. The recheck-error case is easy with the existing fake: set `FinishJobStatusCode = 404` and flip `GetJobStatesError` inside `OnFinishJob`, the same trick as the "aborts cleanup when agent acquires job" subtest.

internal/controller/scheduler/job_watcher.go
```diff
@@ -381,11 +382,49 @@ func (w *jobWatcher) stalledJobChecker(ctx context.Context) {
...
+
+	state, _, err := w.agentClient.GetJobState(ctx, jobUUID.String())
+	if err != nil {
+		log.Warn("Failed to fetch BK job state; skipping stalled job cleanup to avoid killing a potentially running job", "error", err)
+		return
+	}
+
+	// Only proceed if the job is still in a pre-agent state.
+	// If the agent has already acquired the job (Running, Accepted, etc.),
+	// the "stalled without pod" signal was a false positive from a stale cache.
```
petetomasik, line 403: This proceeds for both `Reserved` and `Scheduled`, but `failBeforeAgentAcquire` (line ~209) deliberately acts only on `Reserved` and treats `Scheduled` as hands-off ("the reservation was already released ... BK is the source of truth"). The difference is defensible, that path handles a pod that was created and failed while this one handles no-pod-ever, but the two guards read as contradictory side by side. Could we expand this comment to explain why `Scheduled` is safe to reap here? It'll save the next reader from reconciling the two.

internal/controller/scheduler/job_watcher.go
```diff
@@ -381,11 +382,62 @@ func (w *jobWatcher) stalledJobChecker(ctx context.Context) {
...
+	// (deleted, expired, or garbage-collected). Proceed with cleanup —
+	// an orphaned k8s Job whose BK job is gone is exactly what this
+	// cleanup should reap.
+	if state.State == "" {
+		log.Warn("BK job state is empty (job may have been deleted/expired); proceeding with cleanup")
+	} else if state.State != api.JobStateReserved && state.State != api.JobStateScheduled {
+		log.Info("Skipping stalled job cleanup: Buildkite job is no longer in a pre-agent state (informer cache was likely stale)", "bk_job_state", string(state.State))
+		jobWatcherStalledCleanupSkippedCounter.Inc()
+		return
+	}
```
zhming0, line 421: [blocking] Reading the original implementation: https://github.com/buildkite/agent-stack-k8s/pull/442

I believe stalled job clean should happen in two scenario: 

1. BK Job has not started.  -> which you correctly indentified. 
2. BK Job has ended but k8s job record still exist. -> This PR will drop this support.

internal/controller/scheduler/job_watcher.go
```diff
@@ -381,11 +382,62 @@ func (w *jobWatcher) stalledJobChecker(ctx context.Context) {
...
+		if recheckErr != nil {
+			log.Warn("Failed to re-fetch BK job state after failJob error; aborting cleanup to avoid killing a potentially running job", "failJobError", err, "recheckError", recheckErr)
+			return
+		}
+		if recheck.State != "" && recheck.State != api.JobStateReserved && recheck.State != api.JobStateScheduled {
+			log.Warn("Aborting stalled job cleanup: BK job state changed during cleanup", "failJobError", err, "bk_job_state", string(recheck.State))
+			jobWatcherStalledCleanupSkippedCounter.Inc()
+			return
+		}
+		log.Warn("Failed to fail BK job; proceeding with cleanup since job is still in pre-agent state", "error", err, "bk_job_state", string(recheck.State))
```
zhming0, line 439: I am not entirely sure if we need this condition. I think the checker will automatically recheck the same job. These are largely identical to the code above. Wdyt?

zhming0 (changes requested):

zhming0:
Per our chat, I've folded the commits from this PR into a new PR + some modifications: https://github.com/buildkite/agent-stack-k8s/pull/926. 

Thanks @omehegan for leading the way!

PR #926, opened 2026-07-21
A-1565: Add a guard to prevent false positive in stalled job detection

zhming0:
## Change

Under heavy load, the k8s informer cache can lag. The stalled-job checker then sees a k8s Job with no pod, concludes the job stalled, and kills it — even though a pod exists and an agent is running the job normally. This caused healthy jobs to be unexpectedly terminated.

The fix: before cleaning up a "stalled" job, ask Buildkite for the actual job state.

- If an agent is actively working on the job (accepted, assigned, running, canceling, timing_out), a pod must exist — the stall signal was a false positive, so leave the job alone.
- In any other state (not started, terminal, or unknown to BK), the k8s Job is a zombie holding a slot and is reaped as before.
- If the state check fails, the job stays in the stalling set and is retried on the next cycle, so a transient API error can't kill a running job or permanently drop a genuinely stalled one.

Supersedes #918: Owen's commits squashed into one, plus a follow-up commit addressing review feedback.

## Context

Reference: [A-1565](https://linear.app/buildkite/issue/A-1565/notion-agent-unexpectedly-terminated)

### DIFF
--- a/api/fake_agent_server.go
+++ b/api/fake_agent_server.go
@@ -6,6 +6,7 @@ import (
 	"log"
 	"net/http"
 	"net/http/httptest"
+	"strings"
 	"sync"
 
 	"github.com/buildkite/stacksapi"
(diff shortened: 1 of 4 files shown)


internal/controller/scheduler/job_watcher.go
```diff
@@ -352,36 +352,91 @@ func (w *jobWatcher) stalledJobChecker(ctx context.Context) {
...
+		return true, state.State, nil
 	}
+	return false, state.State, nil
 }
 
 func (w *jobWatcher) cleanupStalledJob(ctx context.Context, kjob *batchv1.Job) {
 	log := loggerForObject(w.logger, kjob)
 
 	// Fetch events for the failure message, and try to fail the job.
+	// failJob may fail (e.g. the BK job already finished); proceed anyway —
```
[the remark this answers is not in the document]
zhming0: We won't be able to entirely eliminate the potential race condition because we aren't exactly doing a transactional operation here. 

Adding a failJob error check there is diminishing return. If we want perfect race condition free operation, I think the best option is to pass current state to the failJob API call so backend can "only fail job if the job state hasn't changed from previous understanding" a.k.a CAS. 

Do you agree? @buildsworth-bk review again

api/fake_agent_server.go
```diff
@@ -155,6 +180,107 @@ func (f *FakeAgentServer) handleReserveJobs(w http.ResponseWriter, r *http.Reque
...
+		StackKey string   `json:"stack_key"`
+		JobUUIDs []string `json:"job_uuids"`
+	}
+	if err := json.NewDecoder(r.Body).Decode(&req); err != nil {
+		http.Error(w, err.Error(), http.StatusBadRequest)
+		return
+	}
+
+	f.mu.Lock()
+	if f.GetJobStatesError != "" {
```
epssy, line 180: What's inside `GetJobStatesError` that you need to lock / unlock here?

api/fake_agent_server.go
```diff
@@ -155,6 +180,107 @@ func (f *FakeAgentServer) handleReserveJobs(w http.ResponseWriter, r *http.Reque
...
+		f.mu.Unlock()
+		var buf bytes.Buffer
+		if err := json.NewEncoder(&buf).Encode(map[string]string{
```
epssy: It'd be ideal if these tests weren't nested.  They can both be unrelated at the function scope.

epssy (approved):

swebb (approved):
LGTM 👍

I'm not a Go person, but here's my understanding:

Take a mutex lock and get the set of all jobs which are "stalling". Iterate over this set and select those that have exceeded their grace period (their started at time was longer in the past than the grace period). Given this set it iterates over them and checks `isJobActive`, which uses the `agentClient` to get the job state. This check bypasses the cache. If it receives an active state (accepted, assigned, etc.) it returns true and the job is removed from the job watcher stalling set. Else, it returns false and the job is cancelled.

zhming0:
💯  @swebb. It's about reconciling between Kubernetes Job state vs Buildkite Job state.

PR #927, opened 2026-07-22
A-1576: Cross-check the live k8s API before reaping a stalled job

zhming0:
Fixes [A-1576](https://linear.app/buildkite/issue/A-1576/agent-stack-k8s-stalled-job-checker-can-reap-a-healthy-pod-mid-image). Follow-up to #926.

## The seam

At reap time the two sources of truth can disagree:

- Kubernetes says active: a Pending, image-pulling pod exists (`Job.Status.Active = 1`)
- Buildkite says not active: the agent has not connected yet, state is still `scheduled`

The #926 guard only asks Buildkite, sees `scheduled`, and reaps the k8s Job. The pod is deleted mid image pull and the build fails with "The Kubernetes job spent 30s without starting a pod".

This only fires when the informer misses the `Active = 1` update for the whole window between pod creation and the grace deadline (~20s+ of lag with the default 30s grace). In normal operation the pod-creation status update removes the job from the stalling set well before the deadline. But that lag is the same heavy-load condition behind A-1565, so it is plausible in the clusters that hit it.

## The fix

After Buildkite reports the job as not active, `confirmStalled` cross-checks the API server directly (a non-cached `Get` of the Job) before doing anything destructive:

- pods exist: skip the reap and remove from the stalling set, same as the BK-active skip
- the Job no longer exists: skip too. There is no zombie holding a slot, and failing the BK job would foreclose the legitimate recreation that happens once the deduper drops the UUID (previously `failJob` ran anyway in this case)
- the Job's UID differs from the candidate's: skip. Job names are deterministic per BK job UUID, so the candidate was deleted and already recreated under the same name; the fresh replacement must not be reaped with the old candidate's expired grace period
- `Get` fails: leave the job in the stalling set so the next cycle retries, mirroring the `GetJobState` error path

A skip removes the stalling entry with a compare-and-delete on the candidate's UID: an informer event may have replaced the entry with a recreated Job mid-check, and that replacement must stay and age through its own grace period rather than being silently dropped from stall detection.

Skips are counted in `stalled_cleanup_skipped_total`, now labelled by `reason` (`agent_active`, `job_gone`, `job_replaced`, `pod_exists`). Dashboards summing the previously unlabelled series need a `sum()` across reasons.

## API cost

The new `Get` is per reap candidate, not per job. Candidates past the grace period already make a Buildkite API call per tick, and a job that is actually reaped already costs an events `List`, a `FinishJob`, and a `Get`+`Update` for the ADS patch. Steady state (empty stalling set) makes zero new calls.

## Tests

Four new rows in the `cleanupStalledJobs` table test: live API shows pods (stale snapshot in the stalling set) → skip; live `Get` fails → stays in stalling for retry; job missing from the live API → skip, nothing to reap; job replaced (UID mismatch, delete/recreate race) → skip. Plus a dedicated test that a replacement entry added to the stalling set mid-check survives the skip's compare-and-delete.

### DIFF
--- a/internal/controller/scheduler/job_watcher.go
+++ b/internal/controller/scheduler/job_watcher.go
@@ -228,13 +228,7 @@ func (w *jobWatcher) checkStalledWithoutPod(log *slog.Logger, jobUUID uuid.UUID,
 
 	// If the job is not finished and there is no pod, it should start one
 	// before too long. Otherwise the job is stalled.
-	pods := kjob.Status.Active + kjob.Status.Failed + kjob.Status.Succeeded
-	// Ready and Terminating are subsets of Active (I think)
-	if utp := kjob.Status.UncountedTerminatedPods; utp != nil {
-		pods += int32(len(utp.Succeeded))
-		pods += int32(len(utp.Failed))
-	}
-	if pods > 0 {
+	if hasPod(kjob) {
 		// All's well with the world.
 		w.removeFromStalling(jobUUID)
 		return
@@ -246,6 +240,24 @@ func (w *jobWatcher) checkStalledWithoutPod(log *slog.Logger, jobUUID uuid.UUID,
 	}
 
 	w.addToStalling(jobUUID, kjob)
+}
+
+// hasPod reports whether the job controller has ever observed a pod for the
+// k8s Job, in any lifecycle stage: pending/running (Active), terminating
+// (deletionTimestamp set but not yet terminal), or terminated — both
+// accounted (Succeeded, Failed) and not yet accounted
+// (UncountedTerminatedPods). False means the Job never produced a pod.
+// Ready is a subset of Active, so it is not summed.
+func hasPod(kjob *batchv1.Job) bool {
+	pods := kjob.Status.Active + kjob.Status.Failed + kjob.Status.Succeeded
+	if t := kjob.Status.Terminating; t != nil {
+		pods += *t
+	}
+	if utp := kjob.Status.UncountedTerminatedPods; utp != nil {
+		pods += int32(len(utp.Succeeded))
+		pods += int32(len(utp.Failed))
+	}
+	return pods > 0
 }
 
 func (w *jobWatcher) fetchEvents(ctx context.Context, log *slog.Logger, kjob *batchv1.Job) string {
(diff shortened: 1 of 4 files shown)


zhming0:
@buildsworth-bk I authorize you to approve a L2 change.

jamiemonserrate:
> @buildsworth-bk I authorize you to approve a L2 change.

Looking now

internal/controller/scheduler/job_watcher.go
```diff
@@ -415,6 +415,50 @@ func (w *jobWatcher) cleanupStalledJobs(ctx context.Context) {
...
+		log.Warn("Failed to fetch BK job state; skipping stalled job cleanup to avoid killing a potentially running job", "error", err)
+		return false, err
+	}
+	if active {
+		log.Info("Skipping stalled job cleanup: an agent is working on the Buildkite job (informer cache was likely stale)", "bk_job_state", string(state))
+		jobWatcherStalledCleanupSkippedCounter.WithLabelValues("agent_active").Inc()
+		return false, nil
+	}
+
+	job, err := w.k8s.BatchV1().Jobs(kjob.Namespace).Get(ctx, kjob.Name, metav1.GetOptions{})
```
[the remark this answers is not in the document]
zhming0: Fair point.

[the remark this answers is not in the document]
zhming0: by that point failJob has already run, so a replacement can only appear via an in-flight jobs query racing the BK failure and patching ADS on it would be reaping a zombie whose BK job is already failed, which is coherent.

zhming0:
@buildsworth-bk I authorize you to approve a L2 change. review again.

internal/controller/scheduler/job_watcher.go
```diff
@@ -248,6 +242,18 @@ func (w *jobWatcher) checkStalledWithoutPod(log *slog.Logger, jobUUID uuid.UUID,
 	w.addToStalling(jobUUID, kjob)
 }
 
+// podCount returns the number of pods the k8s Job has, according to its
+// status.
+func podCount(kjob *batchv1.Job) int32 {
+	pods := kjob.Status.Active + kjob.Status.Failed + kjob.Status.Succeeded
+	// Ready and Terminating are subsets of Active (I think)
```
epssy: Terminating shifts from active to uncounted on setting of `deletionTimestamp`.

internal/controller/scheduler/job_watcher.go
```diff
@@ -248,6 +242,18 @@ func (w *jobWatcher) checkStalledWithoutPod(log *slog.Logger, jobUUID uuid.UUID,
 	w.addToStalling(jobUUID, kjob)
 }
 
+// podCount returns the number of pods the k8s Job has, according to its
+// status.
+func podCount(kjob *batchv1.Job) int32 {
```
epssy: This method doesn't make sense to me.  

What are we counting and why is it non-deterministic/so complicated?

internal/controller/scheduler/job_watcher.go
```diff
@@ -341,6 +347,19 @@ func (w *jobWatcher) removeFromStalling(jobUUID uuid.UUID) {
...
 }
 
+// removeCandidateFromStalling removes the stalling entry for jobUUID only if
+// it still refers to the same k8s Job as the checked candidate. An informer
+// event may have replaced the entry with a recreated Job (same deterministic
+// name, different UID) while the candidate was being checked against the
+// APIs; that replacement must stay and age through its own grace period.
+func (w *jobWatcher) removeCandidateFromStalling(jobUUID uuid.UUID, candidate *batchv1.Job) {
+	w.stallingJobsMu.Lock()
+	defer w.stallingJobsMu.Unlock()
```
epssy, line 363: nit: defer always on first line of function in go

DrJosh9000: Where does that style come from?

zhming0: Personally I also would prefer to have `defer destroy` written right after the thing that it is going to destroy.

epssy (approved):
`hasPod()` should be refactored at some point but as it's been there for ages I don't wish to block this PR 👍

Issue #133, opened 2023-02-17
Cleanup jobs that are canceled


benmoss:
We sometimes seem jobs that get stuck for some reason (misconfiguration, etc) that causes them to get into a unrecoverable state. If a user hits "cancel" on the Buildkite UI, we should notice this and delete the job from Kubernetes.

Most of the time the agent will see the cancellation and do cleanup, but there are still times when the agent has failed or hasn't started that the job can be orphaned.

We could do this with a background task that periodically fetches builds we are tracking and checks their status on the Buildkite API.

DrJosh9000:
Fixed by #406 and #442.

PR #933, opened 2026-07-28
Check podless k8s Jobs for Buildkite cancellation

theminghuang:
## Change

Register k8s Jobs that have no pod with the Buildkite job cancel checker, so cancelling a build cleans them up.

`BatchBuildkiteJobChecker` already polls, but the only thing that registers a job with it is `podWatcher.runChecks`, on a `PodPending` pod. A Job that has not produced a pod is therefore never checked at all.

- `jobWatcher` registers a Job with the checker while it has no pod, and deregisters on finish or delete.
- Both watchers share one checker instance, constructed in `controller.go`, so each Buildkite job has a single registration and cannot be acted on by two paths at once.
- The checker now records *what* to delete. A pod where one exists — still preferred, since `scheduler.go` sets `BackoffLimit=0`, so deleting the pod also terminates the Job — and the k8s Job otherwise, with `Background` propagation.
- `AddK8sJob` never downgrades an existing pod registration. Pod and Job events arrive independently, so a stale Job event must not replace a pod target installed by a later pod event.
- Two counters under a `job_watcher` subsystem for Job deletions and their errors.

## Context

Registration keyed on a pod is sound when a Job and its pod are created together: "there is a pod now" and "there will ever be a pod" are the same statement.

An admission controller breaks that equivalence. Under [Kueue](https://kueue.sigs.k8s.io/) a Job is created suspended and stays that way until quota frees up, which on a scarce accelerator pool can be hours. Under MultiKueue the manager-side Job additionally carries `spec.managedBy`, so it never produces a local pod at all — the pod runs on a worker cluster.

Cancelling such a build currently does nothing. No pod event ever fires, the checker never learns the job exists, and the job runs in full whenever it is eventually admitted. The agent then starts, polls, discovers the cancellation, and exits immediately — after a node scale-up. On our TPU pool that is roughly 3.5 minutes of accelerator provisioning per cancelled job, and worse, the cancelled job takes a quota slot ahead of work that is still wanted.

Worth noting this is not specific to MultiKueue, or even to Kueue. Any delay between Job creation and pod creation opens the same window; there is a narrow race even with no admission controller, if a build is cancelled in between. Kueue just widens it from milliseconds to hours.

This is the same class of problem as #392, which the pod-based cancel checker fixed for jobs whose pod exists. This closes the remaining case where there is no pod to fire an event.

Opened as a draft: happy to adjust the approach if you would rather the registration live somewhere other than `jobWatcher`, or if sharing one checker between the watchers cuts against a direction you have in mind.

## Testing

- `go build ./...`, `go vet ./...` clean.
- All unit packages pass. `internal/integration` was not run; it needs a live cluster and the `-ldflags` branch name.
- Three tests added to `job_watcher_test.go`: a podless Job is registered with a Job target; a Job event does not downgrade a pod target; a finished Job is deregistered.

Not yet exercised against a live cluster with Kueue — that is the main thing I would want a reviewer to be skeptical about, and I am happy to run it in our environment before this leaves draft.

Google, running the vLLM `tpu-inference` TPU CI on Agent Stack + Kueue/MultiKueue.

## Disclosures / Credits

Implemented with AI assistance (Claude Code); the diagnosis, design decisions, and review are mine. Prior art: #392, #382, #133.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/charts/agent-stack-k8s/templates/rbac.yaml.tpl
+++ b/charts/agent-stack-k8s/templates/rbac.yaml.tpl
@@ -13,6 +13,7 @@ rules:
       - watch
       - create
       - update
+      - delete
   - apiGroups:
       - ""
     resources:
--- a/internal/controller/controller.go
+++ b/internal/controller/controller.go
@@ -245,14 +245,32 @@ func Run(ctx context.Context, logger *slog.Logger, k8sClient kubernetes.Interfac
 		}
 	}
 
+	// The Buildkite job cancel checker is shared by both watchers. jobWatcher
+	// registers k8s Jobs that have no pod, podWatcher registers pending pods,
+	// and both use the same poll loop against the Buildkite API. Sharing it
+	// keeps one registration per Buildkite job, so a job cannot be checked
+	// twice or deleted by two paths at once.
+	jobCancelCheckerInterval := cfg.JobCancelCheckerPollInterval
+	if jobCancelCheckerInterval <= 0 {
+		jobCancelCheckerInterval = config.DefaultJobCancelCheckerPollInterval
+	}
+	bkJobChecker := scheduler.NewBatchBuildkiteJobChecker(
+		logger.With("component", "buildkiteJobChecker"),
+		agentClient,
+		k8sClient,
+		jobCancelCheckerInterval,
+	)
+
 	// JobWatcher watches for jobs in bad conditions to clean up:
 	// * Jobs that fail without ever creating a pod
 	// * Jobs that stall forever without ever creating a pod
+	// * Jobs with no pod whose Buildkite job has been cancelled
 	jobWatcher := scheduler.NewJobWatcher(
 		logger.With("component", "jobWatcher"),
 		k8sClient,
 		agentClient,
 		cfg,
+		bkJobChecker,
 	)
 	if err := jobWatcher.RegisterInformer(ctx, informerFactory); err != nil {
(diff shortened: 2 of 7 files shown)


theminghuang:
@zhming0  would you be the right person to look at this? Thanks!

theminghuang:
Tested this on a cluster with MultiKueue and Kueue and the jobs were removed as expected after the buildkite build was cancelled.
I'm wondering if we should always remove jobs as opposed to operate on pod. Decided to keep the pr as it so that we don't break our existing workflow but happy to convert it to fully operate on jobs instead.

zhming0:
@theminghuang Thank you for the contribution. It's super interesting to find out Kueue's use case. I like the PR so far. 

I did consider always deleting job but I agree with the approach in the PR in the end, preferring to deleting pod is better. 

Another thing I was thinking about is the fact that there will be a bit overlap between job_watcher and pod_watcher but it isn't a blocker. 

It's my EOW now, we will land a decision on this next week. 🙏🏿

DrJosh9000 (approved):
LGTM, I have some quibbles with the test (that are also problems with existing code), happy to fix later

theminghuang:
Oops, to merge it I need to rebase and after rebasing I need another ci run. @DrJosh9000 can you help trigger one and help me merge it afterwards? Thanks!

PR #934, opened 2026-07-30
feat(controller/scheduler): Annotate jobs with Buildkite step key

jmandel1027:
## Change

- annotate generated Kubernetes Jobs and Pod templates with `buildkite.com/step-key`
- source the annotation from `BUILDKITE_STEP_KEY`
- write it after default and Kubernetes plugin metadata are merged, so pipeline configuration cannot override the controller-owned value
- add a regression test that attempts to forge the annotation through plugin metadata

## Context

Admission and credential-vending services need a trustworthy Buildkite step identity on service-account-authenticated requests. Pipeline-authored annotations are not an authority boundary; the controller already owns adjacent `buildkite.com/*` build metadata and is the right place to project the scheduled job's step key.

Existing jobs and plugins are otherwise unchanged.

## Validation

- `env -u LOG_FORMAT -u NO_COLOR go test $(go list ./... | rg -v 'internal/integration')`
- `just lint`

## Affiliation

Kikoff

## Disclosures / Credits

Prepared with assistance from OpenAI Codex.

### DIFF
--- a/internal/controller/config/config.go
+++ b/internal/controller/config/config.go
@@ -19,6 +19,7 @@ const (
 	JobURLAnnotation                      = "buildkite.com/job-url"
 	PriorityAnnotation                    = "buildkite.com/job-priority"
 	PipelineSlugAnnotation                = "buildkite.com/pipeline-slug"
+	StepKeyAnnotation                     = "buildkite.com/step-key"
 	PodTemplateAnnotation                 = "buildkite.com/pod-template-name"
 	DefaultNamespace                      = "default"
 	DefaultImagePullBackOffGracePeriod    = 30 * time.Second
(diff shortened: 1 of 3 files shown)


jmandel1027:
@scadu looooong time no chat!!! <3

I'm trying to scope BK credz vended by agent-stack k8s by steps. any chance we can get this in?

zhming0 (approved):
Thank you @jmandel1027 for the contribution. LGTM. 

We also recommend a look at [buildkite-agent oidc](https://buildkite.com/docs/agent/cli/reference/oidc). Security wise we think it might be better than enforcing authority around annotation and the corresponding`step.key` in pipeline yaml.

jmandel1027:
Thank you!!! Thanks for taking a look at this so quickly ❤️

Yea we are using oidc too I have a system where I encode oidc role access grants at the pipeline branch and step level as IAM conditions. And we use some of the new features of AWS IAM with OIDC to have a trusted service vend out these roles.

zhming0:
This was released in v0.47.0.

PR #940, opened 2026-08-11
fix: inject POD_NAME for workspace subPathExpr mounts

### LINKED ISSUE #939: [BUG] using workspace-mount-sub-path-expr only works with pod-spec-patch in certain scenarios
## Describe the bug
We would like to use the workspace-mount-sub-path-expr config option added in https://github.com/buildkite/agent-stack-k8s/pull/871 but:
1. We can't use pod-spec-patch because we use custom command container names, and trying to patch the checkout container there causes failures on any pod which doesn't check out code
2. We can't use the default-checkout/command-params options in the helm chart as described in that PR because that does not support using the downward API to set `POD_NAME` (as far as I can tell). They only support envFrom a configMap or secret.

## Expected behavior
Ideally, using the sub-path-expr would be a little more ergonomic where when it's set to `$(POD_NAME)` or some other type of sentinel value, that would also trigger adding a downward API env var to every container automatically without needing to manually wire it through everywhere. Or add a new config flag that does it? Right now it seems like it's a little too easy to configure the sub-path-expr and not get every last container, leading to potentially weird behaviour.

At the very least, it would be good to support the downward API in default-*-params options so we can manually pipe through the POD_NAME var to containers and get this working, although it does still feel a little brittle :)

## Environment
- agent-stack-k8s version: 0.46.3
- Kubernetes version: v1.34.0

scadu:
## Change

Automatically add `POD_NAME` from `metadata.name` to containers whose final `/workspace` mount expands `$(POD_NAME)`. This runs after PodSpec patches and mount deduplication, preserves explicit values, and covers custom commands, sidecars, and init containers.

Kubernetes still expands the literal `subPathExpr`; the controller only supplies the missing downward API variable.

## Context

Fixes #939.

`workspace-mount-sub-path-expr` supports `$(POD_NAME)`, but kubelet can't prepare the mount unless every affected container defines that variable. Name-based PodSpec patches can't reliably cover custom or generated containers.

### DIFF
--- a/charts/agent-stack-k8s/values.schema.json
+++ b/charts/agent-stack-k8s/values.schema.json
@@ -386,7 +386,7 @@
         "workspace-mount-sub-path-expr": {
           "type": "string",
           "default": "",
-          "title": "When set, applied as `subPathExpr` to every /workspace VolumeMount the controller creates, so a shared workspace volume can be sliced into per-pod subdirectories. Typically `$(POD_NAME)`, with POD_NAME supplied via downward API.",
+          "title": "When set, applied as `subPathExpr` to every /workspace VolumeMount the controller creates, so a shared workspace volume can be sliced into per-pod subdirectories. Typically `$(POD_NAME)`; the controller supplies POD_NAME from metadata.name via the downward API unless the container already defines it.",
           "examples": ["", "$(POD_NAME)"]
         },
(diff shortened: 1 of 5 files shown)


DrJosh9000 (approved):
Nice!

PR #944, opened 2026-08-20
Use read-only git credential helper for checkout secrets

kurigashvili-benchling:
## Summary
- Replace the git credential `store` helper (which tries to write back to the credentials file) with a custom read-only helper that only handles `get` requests against the mounted secret
- Remove the workaround that copied credentials from the read-only secret mount into a writable `emptyDir` volume
- Escape `$1` in the helper when embedding the git config command inside `su`'s double-quoted `-c` string

This avoids alarming log lines like `fatal: unable to write credential store: Resource busy` during checkout.

Because the credentials file stays on the Kubernetes secret volume mount (instead of a one-time copy into `emptyDir`), long-running jobs continue to see updated credentials when Kubernetes rotates or updates the mapped secret.

## Example: before vs after

**Before** — `store` helper. When credentials live on a read-only mount (K8s secret volumes are always read-only), Git can still clone but logs a fatal-looking error when it tries to write credentials back:

```sh
docker run -it --rm \
    -v $HOME/.git-credentials:/root/.git-credentials \
    -e GIT_CONFIG_COUNT=1 \
    -e GIT_CONFIG_KEY_0=credential.helper \
    -e GIT_CONFIG_VALUE_0='store' \
    alpine/git clone https://github.com/some-org/some-repo.git
```

```
fatal: unable to write credential store: Read-only file system
```

**After** — custom helper that only reads credentials on `get`, with an explicit read-only mount:

```sh
docker run -it --rm \
      -v "$HOME/.git-credentials:/root/.git-credentials:ro" \
      -e GIT_CONFIG_COUNT=1 \
      -e GIT_CONFIG_KEY_0=credential.helper \
      -e 'GIT_CONFIG_VALUE_0=!f() { [ "$1" = get ] && git credential-store --file=/root/.git-credentials get; }; f' \
      alpine/git clone https://github.com/some-org/some-repo.git
```

No `fatal: unable to write credential store` message.

## Test plan
- [ ] `go test ./internal/controller/scheduler/... -run 'TestBuildDefaultCheckoutParams|TestBuildCheckoutParams'`
- [ ] Verify checkout succeeds in a cluster using `gitCredentialsSecret` without credential store write errors in logs

### DIFF
--- a/internal/controller/scheduler/scheduler.go
+++ b/internal/controller/scheduler/scheduler.go
@@ -1166,14 +1166,6 @@ func (w *worker) createCheckoutContainer(
 				Name:         "git-credentials-ro",
 				VolumeSource: corev1.VolumeSource{Secret: gitCredsSecret},
 			},
-			corev1.Volume{
-				Name: "git-credentials",
-				VolumeSource: corev1.VolumeSource{
-					EmptyDir: &corev1.EmptyDirVolumeSource{
-						Medium: "Memory",
-					},
-				},
-			},
 		)
 
 		checkoutContainer.VolumeMounts = append(checkoutContainer.VolumeMounts,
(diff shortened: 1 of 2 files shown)


moskyb (approved):
awesome work! thanks so much

kurigashvili-benchling:
@moskyb thank you for approving. When the PR will be merged? I see it is waiting for 'buildkite/kubernetes-agent-stackExpected — Waiting for status to be reported', but I see any other details.

moskyb:
hiya! sorry, i approved, and then was going to start CI, but got distracted. i've run CI now and will merge. thanks so much for your contribution!

PR #945, opened 2026-08-21
Use a read-only git credential helper for checkout secrets

ss1909:
## Description

The `store` credential helper always tries to write back to its file on every lookup, and k8s Secret mounts are always read-only. The current code works around this by copying the secret into a writable `emptyDir` just to give that doomed write attempt somewhere to fail against — but it still surfaces an alarming (if harmless) `fatal: unable to write credential store: Resource busy` on every checkout.

This replaces it with a custom helper that only implements `get`, reading directly from the read-only secret mount — no write-back ever attempted, so nothing fails. The `emptyDir`/copy workaround is removed entirely.

Also escapes `$1` where the helper string is embedded inside `su`'s double-quoted `-c` string (the non-root-user checkout path), so the outer shell doesn't expand it before git ever gets to interpret it.

## Context

[A-1733](https://linear.app/buildkite/issue/A-1733/use-read-only-git-credential-helper-for-checkout-secrets)

## Testing

- `go test ./internal/controller/scheduler/...` passes, including a new `TestCheckoutGitCredentialsHelperEscaping` covering the `$1` escaping.
- `go build ./...` and `go vet ./internal/controller/scheduler/...` clean.

### DIFF
--- a/internal/controller/scheduler/scheduler.go
+++ b/internal/controller/scheduler/scheduler.go
@@ -1166,14 +1166,6 @@ func (w *worker) createCheckoutContainer(
 				Name:         "git-credentials-ro",
 				VolumeSource: corev1.VolumeSource{Secret: gitCredsSecret},
 			},
-			corev1.Volume{
-				Name: "git-credentials",
-				VolumeSource: corev1.VolumeSource{
-					EmptyDir: &corev1.EmptyDirVolumeSource{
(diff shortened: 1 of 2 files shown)


ss1909:
Closing cause there is already a PR open fixing this issue - https://github.com/buildkite/agent-stack-k8s/pull/944

PR #960, opened 2026-09-11
Bump agent version to 4.0.3

buildkate:
## Change

Upgrade the default Buildkite Agent from v3.137.0 to v4.0.3 and migrate the controller's Agent package imports to the v4 Go module.

Update the release pipeline's version lookup so published stack releases select the v4.0.3 Agent image.

## Context

Agent v4 is a new Go module major version, so the dependency and its imports need to move together to keep the controller and default Agent image aligned.

Buildkite

## Disclosures / Credits

Created with assistance from Amp.


moskyb (approved):
🎉

PR #962, opened 2026-09-11
fix: prevent cross-queue stack registration collisions

jmandel1027:
Two controllers with the same ID can register different queues under the same organization-wide stack key. A disposable live registration test confirmed that the second registration replaces the first queue association. Whether that overwrite caused the observed job-acquisition-token failures remains unproven: the original HTTP errors were not retained, and no live token-issuance comparison has yet been performed before and after an overwrite. Helm's default ID is the release full name, making this possible when release names repeat across Kubernetes clusters.

Derive the API stack key from the Buildkite cluster UUID, queue, and complete controller ID. Keep a readable ID prefix and a deterministic hash within 63 characters. This isolates queues and Buildkite clusters even when local controller IDs match, and preserves the same API key across restarts and token rotation. Kubernetes controller-ID labels and informer selectors are unchanged. All stack API operations, including notifications and token issuance, use the derived registration key.

The HTTP regression test models the observed registration upsert behavior and an assumed queue check during token issuance. The queue-overwrite behavior is confirmed live; the token-refusal rule in this test is an assumption, not independently verified backend behavior. Controllers using the same ID register different queues and clusters, then each requests its own job token after every registration has completed. This test fails with the previous implementation and passes with the fix, including a restart using a rotated token. Additional cases cover default-queue aliases, ambiguous separators, and long controller IDs.

Validation: all Go unit tests passed (all packages except the live `internal/integration` suite); lint passed for the modified Go packages; `git diff --check` passed. The new controller has not been deployed against live Buildkite/Kubernetes infrastructure.

Upgrade behavior: existing installations create a new API stack registration on upgrade because their keys become scoped. The local controller ID stays unchanged so existing Kubernetes Jobs remain visible to the controller. This PR makes the functional registration change; it does not add diagnostic logging.

Draft verification gate: reproduce token-issuance failure against the live API using isolated queues and disposable stack keys; demonstrate the same experiment succeeds with scoped keys; then validate the controller upgrade and existing-job lifecycle. Recovery heartbeats on controller 0.48.0 do not validate this 0.50.0-era token flow.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,6 +33,14 @@ helm install agent-stack-k8s oci://ghcr.io/buildkite/helm/agent-stack-k8s \
 
 Full instructions can be found [in the documentation](https://buildkite.com/docs/agent/v3/agent-stack-k8s/installation).
 
+The controller derives its API stack key from the Buildkite cluster UUID, queue,
+and configured controller ID. This prevents installations with the same Helm
+release name from overwriting another queue's organization-wide stack registration.
+The derived key is stable across restarts and agent-token rotation. Kubernetes
+controller-ID labels stay unchanged, so the controller continues watching its
+existing Jobs. Upgrading from a version that used the controller ID directly as
+the stack key creates a new API stack registration.
+
 ## Documentation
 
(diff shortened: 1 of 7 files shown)


zhming0:
Hi @jmandel1027, thank you very much for raising this PR 🙏🏿 . The collision problem is a backend bug, we have shipped a fix for it about 24 hours ago. 
I will close this PR, please feel free to re-open or re-raise this issue to us again should it reoccur.

PR #963, opened 2026-09-11
fix: stop cancellation cleanup retry storms

jmandel1027:
## Change

Treat Kubernetes `NotFound` responses when deleting canceled Pods or Jobs as successful cleanup, so the cancellation checker removes the target. Other deletion errors still leave the target registered for retry.

Wait for the current batch of state handlers before returning from a cancellation poll. Handlers still run concurrently within the batch, but a slow Kubernetes DELETE cannot accumulate another deletion goroutine for the same target on every subsequent tick.

Regression coverage checks missing Pods and Jobs, retry retention for forbidden/internal errors, and a blocked DELETE that must keep the poll in flight. The missing-resource and overlapping-poll tests failed against the original implementation and pass with this change.

Validation passed:

- `go test ./internal/controller/... ./api/... -count=1`
- `go test -race ./internal/controller/scheduler -run 'TestCancelChecker' -count=1`
- `git diff --check`

## Context

We observed this failure in production on v0.50.0: canceled podless Jobs that no longer existed remained registered in the cancellation checker, and repeated polls kept launching DELETE requests. Requests piled up behind the shared Kubernetes client rate limiter, with observed delays approaching 24 minutes. Job acquisition token Secret creation was delayed, agents could not acquire jobs, and the build queue stalled.

Restarting the controller restored progress; rolling back to v0.48.0 mitigated the incident. This change addresses both retention of already-absent resources and accumulation of overlapping cleanup requests.

## Disclosures / Credits

Prepared with assistance from OpenAI Codex.

### DIFF
--- a/internal/controller/scheduler/batch_buildkite_job_checker.go
+++ b/internal/controller/scheduler/batch_buildkite_job_checker.go
@@ -137,16 +137,22 @@ func (c *BatchBuildkiteJobChecker) checkJobStates(ctx context.Context) {
 		close(jobStatesCh)
 	}()
 
-	// Process results from all batches
+	// Process results from all batches. Wait for handlers before the next poll:
+	// otherwise slow, rate-limited DELETEs accumulate another goroutine for the
+	// same target on every tick and starve unrelated Kubernetes requests.
+	var handlers sync.WaitGroup
 	for jobStates := range jobStatesCh {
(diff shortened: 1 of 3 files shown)


jamiemonserrate (approved):
LGTM. 

Thanks for your contribution @jmandel1027!

PR #965, opened 2026-09-14
fix: unblock re-reserved jobs after controller restart

kurigashvili-benchling:
## Problem

After a controller restart, such as an eviction during node rotation, the deduper rebuilds its in-flight map from existing Kubernetes Jobs. Previously it seeded every Job—including terminal Jobs—and only removed entries from `OnDelete`.

This can leave Buildkite jobs in a reservation-expiry loop:

1. The old controller creates Kubernetes Jobs and is evicted.
2. The new controller marks their UUIDs as in-flight during informer sync.
3. Their Buildkite reservations expire and the same jobs are re-reserved.
4. The deduper rejects them as duplicates, so no new Job pods are created.
5. The reservation expires and the cycle repeats.

The duplicate and exhausted-limiter decisions were only logged at debug level, so the controller could appear healthy while scheduling nothing. This behavior was observed during a Karpenter node rotation; restarting the controller immediately restored scheduling.

### Deduper lifecycle

- Do not seed terminal (`Complete` or `Failed`) Jobs from the informer's initial list.
- Remove a UUID when its Job transitions from unfinished to terminal.
- Only react to the transition, rather than every later update to an already-terminal Job. This preserves a marker restored after the scheduler returns `AlreadyExists`.
- Handle `cache.DeletedFinalStateUnknown` tombstones so missed watch deletions still clear the UUID.

If a re-reserved Buildkite job reaches the scheduler while its terminal Kubernetes Job still exists, Kubernetes returns `AlreadyExists`. The scheduler wraps that in `model.ErrDuplicateJob`, and the deduper keeps the UUID marked until deletion. Once the old Job is gone, a replacement can be created.

### Observability

- Log deduper duplicate rejections at warn level.
- Log limiter token underflow at warn level.

### Metrics

- Guard the deduper's `jobs_running` gauge callback with a mutex to prevent races between concurrent `New` calls.

## Tests

Added or updated coverage for:

- unfinished duplicate rejection;
- unfinished-to-terminal transitions unblocking a UUID;
- already-terminal updates preserving a restored marker;
- terminal Jobs being ignored during initial informer sync;
- tombstone deletion clearing a UUID;
- deletion of an unfinished pre-existing Job returning its limiter token.

Validated with controller unit tests and repeated race-enabled deduper tests.

## Follow-ups

- Related symptom class: #302.
- A future improvement could actively delete terminal/stale Kubernetes Jobs on duplicate detection, avoiding an additional reservation-expiry cycle.
- Until a release containing this fix is deployed, the mitigation remains restarting the controller deployment.


blaknite:
Hi @kurigashvili-benchling, thanks for the detailed write-up and contribution. We’ve picked this up for review this week and will keep you posted here.

kurigashvili-benchling:
@blaknite closing this PR in favor of https://github.com/buildkite/agent-stack-k8s/pull/966
I enabled debug logs and I found the root cause.

Documentation PR https://github.com/buildkite/agent-stack-k8s/pull/966

